Issue #17, opened 2019-09-26
Support for standalone AppInsights


Dzoukr:
This one should be quite easy, but I am not so skilled with ARM templates yet.

It would be really hand to have possibility to create standalone instance of AppInsights using comp. expression and have other parts depend on it and use its key. For example in our scenario we have a webApp and functionApp and want them both to be logged by same AppInsights where we need to setup InstrumentationKey.

isaacabraham:
Yep, can definitely add that and that's a good use case 😊

isaacabraham:
Note that currently if by default the template will try to create an AI instance automatically for a webapp or functionsapp and will link them together. It'll also set an app setting with the instrumentation key automatically.

Would it be easier to simply have this:

```fsharp

let myWebApp = webApp {
    name "mysuperwebapp"
    app_insights_name "myappinsights" // same name as above
}

let myFunctionsApp = functions {
    name "myFunctionsApp"
    app_insights_name "myappinsights" // same name as above
}

arm {
    location NorthEurope
    add_resource myFunctionsApp
    add_resource myWebApp
}
```

And then Farmer automatically just creates a single resource and ties them all together?

Dzoukr:
Yes, I like it. Maybe this could be the easiest way. 
Just thinking out loud... shouldn't there be an third option? Because you can create new AI (`app_insights_name "my-ai"`), you can skip it (`no_app_insights`), but cannot reuse (or link-to) existing one?

isaacabraham:
That's the last bit that isn't properly support yet, yes :-)

isaacabraham:
There's sort of support for that - storage account has the same thing - but yes that needs to be tested out and thought through. I'm creating a dedicated AI builder object (it's very simple as there are no config options on AI ARM elements anyway).

isaacabraham:
@Dzoukr I've just pushed a change to master that does the following:

1. Implements a basic AppInsights builder.
2. Adds a "distinct" check towards the end of the Arm Builder pipeline so if two resources of the same type have the same name, it only takes the first one. This means you can do the code sample above and it should just work.

Dzoukr:
Thank you!!!!

PR #18, opened 2019-09-27
Allow turning AI off, linking to external or implict creation.

isaacabraham:
Related to #17. @Dzoukr what do you think of this - one can now do one of four things with AI in functions (and I would add this to webapp as well):

1. Default is to automatically create, configure and link an AI instance (as before).
2. You can also change the name of the "auto" created AI using `app_insights_auto_name`. This lets you link multiple resources together.
3. You can turn off AI completely: `app_insights_off`
4. You can link to an externally created AI instance: `app_insights_link` which takes in the resource name of the "external" AI instance. In this case, you need to still manually add the AI resource that you made into the arm builder.

I have mixed feelings about this PR and whether the flexibility is really needed or not. Do you think all four options are really needed?


Dzoukr:
I think this is great way how to do it! 👍

Issue #27, opened 2019-10-01
Template output contains serialized DU value "Case"


Dzoukr:
Found issue with serialization output for WebApp

```
"Case": "ResourceName",
"Fields": [
    "my-test-web-farmer"
]
```

Will send PR in few minutes.

PR #28, opened 2019-10-01
Fixing issue with serialization of DU #27 + Support for always_on for WebApp

Dzoukr:
With:
```
webApp {
    // other things
    always_on
}
```
image

Without:

image

Also fixing issue #27


isaacabraham:
Thanks!!

isaacabraham:
l

isaacabraham:
Great stuff. I wonder if as a future change it's worth trying to model this somehow in the Sku type on the builder or on the CE, since you can't make always on for the Free SKU?

PR #29, opened 2019-10-01
Added support for setting RuntimeStack for WebApp

Dzoukr:
Added initial support for `runtime_stack` for `WebApp`.

.NET Core 2.2
```
webApp {
  // other stuff
  runtime_stack (WebApp.RuntimeStack.DotNetCore(DotNetCoreRuntime.DotNetCore22, OS.Windows))
}
```

image

.NET FW

```
webApp {
  // other stuff
  runtime_stack (WebApp.RuntimeStack.AspNet AspNetRuntime.AspNet47)
}
```

image

The relations between OS and possible runtimes is designed by DUs. Most of Node values skipped for sanity reasons 😄  (could be nice up-for-grabs).

### DIFF
--- a/src/Farmer/Builders.fs
+++ b/src/Farmer/Builders.fs
@@ -79,7 +79,7 @@ module Storage =
 module WebApp =
     type WorkerSize = Small | Medium | Large
     type WebAppSku = Shared | Free | Basic of string | Standard of string | Premium of string | PremiumV2 of string | Isolated of string
-    type WorkerRuntime = DotNet | Node | Java | Python
+    type FunctionsRuntime = DotNet | Node | Java | Python
     type OS = Windows | Linux
     
     type DotNetCoreRuntime =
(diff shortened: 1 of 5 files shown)


isaacabraham:
Cool! So I've had a quick look through, the design looks sound at a high level but I would like to remove the props anonymous types in the builders. I've started a branch to do this (origin/refactor-29) - can you have a review through what I've started?

isaacabraham:
I've also added overloads on the CE so you don't have to specify the "top level" of the runtime stack DU - does that work for you?

Dzoukr:
Great, this looks good! I am 100% ok with that.

Dzoukr:
Regarding this functionality... I still find issue with creating WebApp for Linux with e.g. Tomcat... It does not setup configuration properly, until I set `"reserved": true` under `properties` of `Microsoft.Web/serverfarms` resource, BUT this one is currently connected with `IsDynamic` flag on WebFarm. To fix this, I would need help from your side probably (you created such flag and most probably know, what it affects).

isaacabraham:
IsDynamic was for functions apps in consumption mode with Y1 sku. Don't know much more than that!

I'll complete those changes later today and we can then push all the way through.

Dzoukr:
> IsDynamic was for functions apps in consumption mode with Y1 sku. Don't know much more than that!

Ok, then once we stabilize this feature, I'll add internal setting for `IsReserved`, keep the current behavior and overwrite it in case of Linux machine to be created. Thanks Isaac!

isaacabraham:
Yes, looks like IsReserved must always be true, see https://github.com/Azure/azure-quickstart-templates/blob/master/101-webapp-basic-linux/azuredeploy.json (and the commit history) as well as [this guide](https://docs.microsoft.com/en-us/azure/azure-functions/functions-infrastructure-as-code) that I used to drive out the Functions code in the first place.

PR #30, opened 2019-10-03
Added support for Azure Functions Default and Master keys

Dzoukr:
Function app host keys `Default` and `_Master` can now be used as params (or appSettings) for other resources 🎉 

This one cost me nearly 50% of my hair (great ARM documentation FTW! :))


isaacabraham:
Thanks again. These sorts of things are always hard to find...

PR #32, opened 2019-10-05
Split out builders into separate modules and files.

isaacabraham:
@Dzoukr As the other person actively working on this, would you have five minutes to review this PR?

It splits out the `Builder.fs` file into separate parts for each ARM resource. Benefit: it's probably easier to make changes to a single ARM template as most of the code is now all in one file. Cost: the Builder logic is no longer one huge `match` expression but split over multiple files.

We could do the same for the Outputters in the Writer as well.


Dzoukr:
👍 Actually I was thinking to propose something like that - having 3 files is great for quick alpha draft, but as it grows, your PR makes it way better to maintain. 100% ok!

PR #36, opened 2019-10-13
Upgrade to F#4.7 and netcore 3.0.

isaacabraham:
Takes advantage of implicit yields in two ways:

1. Removal of existing `yield` statements (`yield!` is expanded to `for x in y do x` - not sure about this).
2. Removed use of `collect` with a `for x in y do` in `Arm.fs`. This was an unexpected "benefit" in that you remove `collect` completely as well as the use of `[ ]` syntax for each case.

Thoughts?


isaacabraham:
@Dzoukr I can't add you as a reviewer for some reason but have a look at this anyway :-)

Dzoukr:
As I would consider .net core 3 & F# 4.7 as the "new default" so totally ok with me. :+1:

isaacabraham:
Same here!

PR #42, opened 2019-11-22
Support for container groups.

ninjarobot:
This adds support for container groups, which can hold one or more Azure Container Instances. Besides the type safety, this also handles a few implementation details, such as container instance names must be lowercase.


isaacabraham:
Thanks a lot :-)

PR #43, opened 2019-11-22
Validation tests for Container Groups.

ninjarobot:


isaacabraham:
Hey @ninjarobot first thanks for the PR, and apologies for taking so long to look at this. I'm wondering: would it be "easier" to simply compare the generated JSON with an "expected" output? Benefit is that the tests become trivial i.e. compare generated JSON with a string - be that might that be too brittle.

isaacabraham:
Sounds good - are there equivalent ways to test other parts of ARM e.g. VM, Storage, App Service etc.?

isaacabraham:
@ninjarobot I've merged this in manually. It's not showing here but it's in the master branch now - thanks :-)

PR #49, opened 2020-01-15
Update code example so it compiles, update json example output.

viktorvan:
Working through the examples in the readme I found some typos, and a code example that no longer compiled. There seems to have been a change in the api for the type returned from the `arm` CE. Which is now a Deployment record instead of a tuple.

There also seem to have been some slight changes in the generate json arm-template.

In the **Deploying by hand** section the `az cli` did not work for me unless I changed `--group` to `--resource-group`.

These were just some minor changes, so I hope it is ok I made a PR directly. Feel free to abandon it if you already have plans to rewrite the readme.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -28,6 +28,7 @@ This is an example bit of Farmer F#:
 
 ```fsharp
 open Farmer
+open Farmer.Resources
 
 // Create a storage resource with Premium LRS
 let myStorage = storageAccount {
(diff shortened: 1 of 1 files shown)


isaacabraham:
Thank you!!

isaacabraham:
@viktorvan FYI - I'm probably going to start creating a proper docs site for Farmer in the near future, so this might be a bit more managed from then.

PR #51, opened 2020-01-16
Add quickdeploy support for linux and os-x

### LINKED ISSUE #50: Make Writer.quickDeploy cross-platform
Is anyone working on making Writer.quickDeploy cross-platform?

I have written something that at least works on my mac, so that would at least be one more platform 🙂 but I think it _should_ work on linux as well. I can make a pull-request from that to get a discussion going. Maybe you have some more requirements for this function that my solution doesn't support.

viktorvan:
This fixes #50 

This is a first draft for making `Writer.quickDeploy` cross-platform. 
I have run the changes in a script on my mac and it works fine. Still left to do is to build a new version of the Farmer package and load that and verify that it works.

And of course we need to verify that I didn't break anything for windows.

### DIFF
--- a/src/Farmer/Writer.fs
+++ b/src/Farmer/Writer.fs
@@ -495,18 +495,46 @@ let toFile armTemplateName json =
     File.WriteAllText(templateFilename, json)
     templateFilename
 
-let toBatchFile armTemplateName resourceGroupName (Location location) templateFilename =
-    let batchFilename = sprintf "%s.bat" armTemplateName
+open System.Runtime.InteropServices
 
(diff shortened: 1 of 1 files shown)


viktorvan:
It may be possible to get around having to set execute permissions for linux/os-x by running the bash script in-line, without first saving it to a file. 

But as a first version I wanted to keep the approache the same as for windows. And maybe it is nice for debugging to actually generate the file?

isaacabraham:
Thanks! I'll test this out on Windows to ensure it's still working and then merge in.

PR #54, opened 2020-01-18
Auto generate password

### LINKED ISSUE #52: Using SecureParameters when running quickDeploy
A question, when trying to deploy a sql-server by running the quickDeploy-command I get an error that the password parameter is missing:

> Please provide securestring value for 'password-for-v-test-sql' 

When deploying the generated arm-template manually I can provide the parameter, of course, so in a production scenario it works fine.

But what is a good way to do this when using quickDeploy?

I actually interpreted the following section in the readme to mean that a (development) password would be automatically generated, but maybe I misunderstood:

> Farmer does support securestring parameters for e.g. SQL and Virtual Machine passwords - these are automatically generated based on the contents of the template rather than explicitly by yourself. However, we don't currently plan on providing rich support for either parameters or variables for several reasons:

So is there a workaround for this that I am not seeing, or can I not use quickDeploy for resources that require SecureParameters?

isaacabraham:
Fixes #52. Automatically generate passwords into a parameters file and supply to the deployment when using `quickDeploy`.

### DIFF
--- a/src/Farmer/Builders.Arm.fs
+++ b/src/Farmer/Builders.Arm.fs
@@ -23,63 +23,70 @@ type ArmBuilder() =
           Location = WestEurope }
 
     member __.Run (state:ArmConfig) =
+        let resources =
+          [ for resource in state.Resources do
+              match resource with
+              | :? StorageAccountConfig as config ->
+                  StorageAccount (Converters.storage state.Location config)
+              | :? WebAppConfig as config ->
+                  let outputs = Converters.webApp state.Location config
+                  WebApp outputs.WebApp
+                  ServerFarm outputs.ServerFarm
+                  match outputs.Ai with (Some ai) -> AppInsights ai | None -> ()
+              | :? FunctionsConfig as config ->
+                  let outputs = config |> Converters.functions state.Location
+                  WebApp outputs.WebApp
+                  ServerFarm outputs.ServerFarm
+                  match outputs.Ai with (Some ai) -> AppInsights ai | None -> ()
+                  match outputs.Storage with (Some storage) -> StorageAccount storage | None -> ()
+              | :? ContainerGroupConfig as config ->
+                  ContainerGroup (Converters.containerGroup state.Location config)
+              | :? CosmosDbConfig as config ->
+                  let outputs = config |> Converters.cosmosDb state.Location
+                  CosmosAccount outputs.Account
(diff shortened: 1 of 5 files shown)


isaacabraham:
@viktorvan can you have a look at this and see if it works as you'd expect?

viktorvan:
The generated password does not fulfill the password complexity requirements for a sql server admin account: https://docs.microsoft.com/en-us/sql/relational-databases/security/password-policy?view=sql-server-ver15

From the link above one must use: 
- Latin uppercase letters (A through Z)
- Latin lowercase letters (a through z)
- Base 10 digits (0 through 9)
- Non-alphanumeric characters such as: exclamation point (!), dollar sign ($), number sign (#), or percent (%).

So instead you could do something like:

```fsharp
    let [<Literal>] private ValidRule1 = "ABCDEFGHIJKLMNOPQRSTUVWXYZ"
    let [<Literal>] private ValidRule2 = "abcdefghijklmnopqrstuvwxyz"
    let [<Literal>] private ValidRule3 = "1234567890"
    let [<Literal>] private ValidRule4 = "!$#%"

    let generatePassword length (template:ArmTemplate) =
        let rndFromArray = 
            let rnd = Random (template.GetHashCode())
            fun (arr: string) ->
                arr.[rnd.Next arr.Length]

        [ Seq.init (length/4) (fun _ -> rndFromArray ValidRule1) 
          Seq.init (length/4) (fun _ -> rndFromArray ValidRule2)
          Seq.init (length/4) (fun _ -> rndFromArray ValidRule3)
          Seq.init (length/4) (fun _ -> rndFromArray ValidRule4) ]
        |> Seq.concat
        |> Seq.toArray
        |> String
```

But that snippet kind of requires that the length parameter is dividable by 4 😉, but you get the idea.

isaacabraham:
Hmmm, weird, I can swear I tested this with SQL. I'll double check.

isaacabraham:
Ok, it works because you only have to pick 3 of those 4, which the existing password text provides for (assuming we randomly get at least one of each).

viktorvan:
Oh, I see. It failed for me at least, so maybe I didn’t get so lucky. It could be good to make it a little more robust maybe, a validation of the generated password or something perhaps?

Other than that I think it works great! It’s really convenient to get a dev-environment setup with a single call to `Writer.quickDeploy`.

isaacabraham:
Fixed.

PR #55, opened 2020-01-18
Make functions_extension_version configurable.

### LINKED ISSUE #53: Make FUNCTIONS_EXTENSION_VERSION configurable
To deploy a function app with runtime v3 using the azure function core tools version 3 you need to set FUNCTIONS_EXTENSION_VERSION = ~3. See https://docs.microsoft.com/sv-se/azure/azure-functions/set-runtime-version

Right now this value is hardcoded to ~2 in the generated arm-template. 

This should be configurable, I think, and it shouldn’t be too hard to make that change. Probably you could use a DU for the supported values 1, 2, and 3 and then add a CustomOperation `use_extension_version` in the function builder. I can take a crack at it in a PR if that’s fine? Unless there are any other specific design decisions to consider?

viktorvan:
This closes #53.

I have added an implementation with a CustomOperation "use_extension_version" in the FunctionsBuilder.

There are no other unit tests as far as I can see, so I haven't added any for this feature.


isaacabraham:
Thanks!!

Issue #69, opened 2020-02-15
Move to System.Text.Json


isaacabraham:
Remove Newtonsoft and migrate to Text.Json. This benefits netcore3 consumers who get it for free. The serialization we're doing is extremely limited so I'm hopeful this would be possible.

PR #70, opened 2020-02-15
Move to System.Text.Json

slang25:
Addresses #69 

I've had a play with some of the samples and the output looks alright, but I haven't actually ran them through Azure.

### DIFF
--- a/src/Farmer/Deploy.fs
+++ b/src/Farmer/Deploy.fs
@@ -1,8 +1,8 @@
 ﻿module Farmer.Deploy
 
 open Farmer.ArmBuilder
-open Newtonsoft.Json
 open System
+open System.Text.Json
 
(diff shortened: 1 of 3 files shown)


isaacabraham:
I've got the branch merged in master (you can pull from there) - doesn't look like it's going to work just yet. Although serialization seems to work ok (I tested simple values and maps), deserialization doesn't work with anonymous records (needs a default constructor i.e. `CLIMutable`) which would mean adding quite a few extra types just to beat the serializer. I'm not sure if that's really worthwhile at the moment.

isaacabraham:
@slang25 appreciate the work in looking into this - but seems that System.Text.Json isn't *quite* ready to replace Newtonsoft. Closing.

PR #78, opened 2020-02-25
Decentralise converters from ARM builder to each builder.

isaacabraham:
This PR has two effects:

1. We decentralise the mapping from the higher level `Config` values into the lower-level Azure resources e.g. from WebAppConfig -> ServerFarmer and WebApp etc. We achieve this through extension members on each builder. I had to add "dummy" `add_resource` and `add_resources` methods on the top ArmBuilder level so that the custom keywords are always available.

2. This provides us with more type safety. Instead of passing arbitrary objects into `add_resource` and needing type checks, we now get compiler safety if no valid overload for `add_resource` exists.

@Dzoukr Thoughts?

### DIFF
--- a/src/Farmer/Builders.Arm.fs
+++ b/src/Farmer/Builders.Arm.fs
@@ -9,7 +9,7 @@ type ArmConfig =
     { Parameters : string Set
       Outputs : (string * string) list
       Location : Location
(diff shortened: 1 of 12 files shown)


isaacabraham:
@Dzoukr Merged :-)

PR #79, opened 2020-03-21
Merge Container and ContainerGroup builders.

isaacabraham:
What do you think of this @ninjarobot ? I've started to apply this pattern elsewhere where there's a "parent" resource and possibly multiple child elements:

* Server Farm -> Web Apps
* Server Farm -> Functions
* Namespace -> Event Hubs

This PR does the same for containers, which have the same relationship i.e. Container Group -> Containers:

Instead of a having a dedicated parent ContainerGroup and child Container builders, we just have a single container builder. If you want to "join" two children to the same container group, you "link" them as follows:

```fsharp
let c1 = container {
    name "foo"
    restart_policy ContainerGroupRestartPolicy.OnFailure
    cpu 2
}

let c2 = container {
    name "bar"
    memory 4.<Gb>
    link_to_container_group c1.GroupName
}

let deployment = arm {
    location NorthEurope
    add_resources [ c1; c2 ]
}
```

The generated ARM template would contain a single container group with both containers nested inside it.

### Benefits of this approach:
* Easier to get started - for single containers, you just need a single builder.
* No need to create a container object and then "connect" to a container group.

### Costs:
* Potential for confusion - you could theoretically set "group" level fields on both containers and link them.
* Doesn't exactly match the ARM style which has containers nested inside a container group.
* Need to add both resource configs to the ARM object, rather than a single container group.

What do you think?


isaacabraham:
@Dzoukr what do you think? This is something that really should be made consistent across all the builders e.g. think about web apps or functions: currently we completely hide the server element - this makes it super easy to get started because you don't need to worry about it, just create a web app and the server is magically made. If you want to "co-locate" two apps on one server, you need to create two resources and then "link" them (I've yesterday created some helper keywords like `link_to_server` for this). The alternative is that we fully expose the server which I think increases the complexity for getting started but it (a) more closely matches the ARM template (not an especially important aim for me - the whole point of Farmer was to simplify things), and (b) is maybe a bit easier for "multi resource" instances e.g. a web server with multiple web apps / functions.

Dzoukr:
Sorry @isaacabraham I've been jumpin' around thousand things.

I like this `link_to_server` approach. If well documented, it will still hide away max of ARM complexity with minimum effort (from Farmer consumer's side). As you wrote, the main benefit of Farmer is easy DSL and some more complex things are hidden (on purpose). I would take exposing server as the last (and probably breaking) option.

PR #82, opened 2020-04-01
Update Functions documentation - wrong setting name in functions.md

Youenn-Bouglouan:
Looking at the source code, it seems like the proper name is 'link_to_storage_account' and not 'storage_account_link'.


isaacabraham:
Yes, this has been updated in v0.7 but the docs are lagging behind :-/ Thanks for fixing.

PR #85, opened 2020-04-01
Basic support for web app deploy.

isaacabraham:
@Dzoukr what do you think?


Dzoukr:
I love it! This is how "right click to deploy to Azure" should look like from the beginning 😄

PR #86, opened 2020-04-01
Fix value of FUNCTIONS_WORKER_RUNTIME in the generated ARM

Youenn-Bouglouan:
Right now, Values of the FUNCTIONS_WORKER_RUNTIME setting for Azure Function Apps are generated using the same casing as defined in the associated DU:

```fsharp
type FunctionsRuntime = DotNet | Node | Java | Python
```
However according to the docs, the values should be lowercase.

image

This causes an issue when trying to deploy a function to the Function App, using VS Code for instance (I presume the same issue might happen using a different deployment means).

image

On a side note, changing my local value from `dotnet` to `DotNet` didn't seem to solve the issue either. The suggested change  should fix this issue.

### DIFF
--- a/src/Farmer/Builders.WebApp.fs
+++ b/src/Farmer/Builders.WebApp.fs
@@ -334,7 +334,7 @@ module Converters =
               Location = location
               AppSettings = [
                 yield! fns.Settings |> Map.toList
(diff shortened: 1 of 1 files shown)


isaacabraham:
@Youenn-Bouglouan thanks for this - merged :-)

Issue #84, opened 2020-04-01
Automatic documentation deployment


isaacabraham:
I have no clue how to do this but I imagine that it's possible to automatically push Hugo sites to github io?

khalidabuhakmeh:
Here are the steps necessary:

1. GitHub Action to build on PR / Tag / Or Whatever
    1. Push results into a "docs" branch or a directory on "master"
2. Set up repository settings to look for branch or docs directory

khalidabuhakmeh:
https://github.com/peaceiris/actions-hugo

SeanKilleen:
Picking this one up.

SeanKilleen:
Note: I've merged the Github Actions workflow to master but it's not yet appearing in the UI. 

Before merging my doc change (just a simple local README that won't change contents), I'm going to make sure the action either shows up or that I confirm that it shouldn't yet.

SeanKilleen:
Confirmed [here](https://github.community/t5/GitHub-Actions/Github-Action-Not-Showing/td-p/52635) that the action should show up after the first run.

That means I'm going to go ahead and get #88 merged which should kick off the action without changing any docs content.

SeanKilleen:
OK, I made another tweak to the docs branch after updating this workflow.

It has not run yet. My guess is it's on some sort of a timeline so I'm going to wait for a little bit before assuming something's wrong. 

I'll check in on it periodically. Feel free to ping me in a little bit if something's gone wrong.

Even if something is wrong in the process, I can force push my local output to the `gh-pages` branch and fix it up.

SeanKilleen:
FYI, I have reached out to the GitHub community about this because I am unsure why the Action workflow isn't being kicked off. My understanding is that as long as the `gh-pages` branch exists and is set for github pages (and it is), the `GITHUB_SECRET` token is what we should be using.

Link in case anyone wants to follow along: https://github.community/t5/GitHub-Actions/GitHub-Action-not-starting/td-p/52767

managedkaos:
Hi @SeanKilleen !  I saw your request and wanted to help out.

I looks like your workflow is not getting triggered because the `docs` branch didn't have a workflow.  In fact, the worflow file is only  on the `master` branch.

I submitted a PR: https://github.com/CompositionalIT/farmer/pull/94

Hope that it helps!

-M

SeanKilleen:
[...]

PR #89, opened 2020-04-04
Add GitHub actions for publishing docs

SeanKilleen:
Supports #84.


SeanKilleen:
@isaacabraham FYI, I am going to merge this, as I don't believe it will be a destructive action. It does not run unless a change is committed to the `docs` branch; my `README` merge will be the first test of that.

PR #103, opened 2020-04-11
Fix #100

### LINKED ISSUE #100: Deploy.quick is not blocking
It should be so that we can run post deploy scripts when everything ist done

isaacabraham:
Adds in `WaitForExit()` on the process. No error checking though.


forki:
Give it to me now!

;-)

00:08:

PR #104, opened 2020-04-11
Create custom deploy folder

isaacabraham:
Creates a `.farmer` folder which contains any transient files created as part of the deployment process. The folder is cleaned on every deployment.


forki:
Nice

00:38:

PR #105, opened 2020-04-11
Create Server Farm builder.

isaacabraham:
Doing this has actually drawn out some nice refactoring regarding the creation of the server farm between webapp and functions (and possibly actually improved correctness around `kind`). However, this still needs some testing to confirm it all works.

cc: @forki can you give this a bash? Will look something like this.

image

Note that with this approach, because you are taking direct control of the server farm, you will need to explicit add it as a resource in addition to the web app.

### DIFF
--- a/src/Farmer/Builders.WebApp.fs
+++ b/src/Farmer/Builders.WebApp.fs
@@ -5,8 +5,8 @@ open Farmer.Helpers
 open Farmer.Resources.Storage
 open Farmer
 
-type WorkerSize = Small | Medium | Large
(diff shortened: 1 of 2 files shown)


forki:
Can we have farm.Name in the web apps?

01:46:

isaacabraham:
It's already there are "server_plan_name". I should rename `serverFarm` to `servicePlan`.

isaacabraham:
@forki I've renamed some bits: the builder is now a `servicePlan` - which reflects the documentation and naming parlance within Azure at the user level. Internally, this gets translated into a `ServerFarm` which is the internal name at the ARM level. I don't know why they're named differently, but this seems the right way to go.

PR #106, opened 2020-04-12
Only prompt for AZ login if required.

isaacabraham:


isaacabraham:
@forki this should remove logins unless necessary.

forki:
Hugs

forki:
Can I pass user and password?

isaacabraham:
@forki show me the command line args you would send - we can add it if needed

forki:
let args = sprintf "login -u \"%s\" -p \"%s\"" user pw
        runAZ args |> ignore

isaacabraham:
Yes, you could. I'm not sure that this is best practice though and therefore I'm not convinced it's something we should bake in (might be totally wrong here - if Azure CLI offers it, can it really be that bad?) - I think you should look at using a [service principal](https://docs.microsoft.com/en-us/cli/azure/reference-index?view=azure-cli-latest#az-login) instead. And we could add that in, too.

forki:
The service principal also uses -u and -p

Isaac Abraham <notifications@github.com> schrieb am Di., 14. Apr. 2020,
09:20:

isaacabraham:
And tenant. :-)

But yes, true. We could make tenant optional.

PR #108, opened 2020-04-14
Azure cli

### LINKED ISSUE #107: Return DeploymentResult in quick deploy instead of unit
Hi there,

for the fullDeploy you receive the DeploymentResult response back after the deployment.
Would be nice to get the DeploymentResult back for the quick deploy as well instead of returning just an unit.

Would be very cool if this can be added.

Thanks in advance

isaacabraham:
This PR removes the HTTP REST API implementation and focuses exclusively on wrapping around the Azure CLI. Reasons:

* Quicker and less F# code to get things going
* We can still interrogate the results as the CLI gives back everything as JSON by default
* If you're using Azure today, you will probably have the CLI already installed, and if not, it's a quick install.
* It supports interactive and non-interactive deployments.

This PR removes the somewhat hacky batch file and replaces it with a more composable and flexible F# wrapper around the Azure CLI, with automatic lifting of calls into Results. Whilst this temporarily removes some nice features that existed in the REST model (notable automatic validation and what-if support), we can always add them in the future again relatively easily.

This closes #107.

### DIFF
--- a/src/Farmer/Builders.Arm.fs
+++ b/src/Farmer/Builders.Arm.fs
@@ -37,7 +37,7 @@ type ZipDeployKind =
             None
     /// Processes a ZipDeployKind and returns the filename of the zip file.
     /// If the ZipDeployKind is a DeployFolder, the folder will be zipped first and the generated zip file returned.
-    member this.GetZipPath(targetFolder) =
+    member this.GetZipPath targetFolder =
         match this with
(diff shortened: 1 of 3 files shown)


forki:
Az ftw

Isaac Abraham <notifications@github.com> schrieb am Mi., 15. Apr. 2020,
00:11:

PR #112, opened 2020-04-16
add settings keyword for webapp builder

vilinski:
This adds, additionally to already existing keyword `setting`, a new CE keyword `settings`, which can take a list of key,value settings. It allows to provide them from extern instead of defining each literally within CE, which by bigger count of settings could be not such comfortable

Example:

```fsharp
let webAppSettings =
    [ "foo", "foo"
      "bar", "bar"
      "baz", "baz"
      "port", "8080" ]

let myFunctions = webApp {
    name "isaacsuperfun"
    settings functionSettings
}
```

Workaround:

```fsharp
type WebAppBuilderExtended() =
    inherit Farmer.Resources.WebApp.WebAppBuilder() with
        /// Sets a list of app setting of the web app in the form "key" "value".
        [<CustomOperation "settings">]
        member __.AddSettings(state:WebAppConfig, settings: (string*string) seq) =
            settings
            |> Seq.fold (fun state (key, value: string) -> __.AddSetting(state, key, value)) state
let webApp = WebAppBuilderExtended()
```


isaacabraham:
Thanks, looks good!

Issue #48, opened 2019-12-31
Support for ACR


dburriss:
I wanted to check if there were plans for ACR support? Either way, unless someone is already working on it, I would like to take a stab at contributing.

isaacabraham:
Not as far as I know - @ninjarobot do you have any plans for this?

isaacabraham:
@dburriss I think @forki also has some interest in this and we might be adding it to web apps shortly.

isaacabraham:
Here's how it's looking at the moment...

```fsharp
let myDockerApp = webApp {
    name "isaacssuperapp"
    app_insights_off
    docker_image "myImage" "startup.dll"
    docker_ci
    docker_registry_credentials "www.myuri.com" "userisaac"
}
```

Generates the following ARM template:

```json
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "outputs": {},
  "parameters": {
    "docker-password-for-isaacssuperapp": {
      "type": "securestring"
    }
  },
  "resources": [
    {
      "apiVersion": "2016-08-01",
      "dependsOn": [
        "isaacssuperapp-plan"
      ],
      "kind": "app,linux,container",
      "location": "northeurope",
      "name": "isaacssuperapp",
      "properties": {
        "httpsOnly": false,
        "serverFarmId": "isaacssuperapp-plan",
        "siteConfig": {
          "alwaysOn": false,
          "appCommandLine": "startup.dll",
          "appSettings": [
            {
              "name": "DOCKER_ENABLE_CI",
              "value": "true"
            },
            {
              "name": "DOCKER_REGISTRY_SERVER_PASSWORD",
              "value": "[parameters('docker-password-for-isaacssuperapp')]"
            },
            {
              "name": "DOCKER_REGISTRY_SERVER_URL",
              "value": "www.myuri.com"
            },
            {
              "name": "DOCKER_REGISTRY_SERVER_USERNAME",
              "value": "userisaac"
            }
          ],
          "linuxFxVersion": "DOCKER|myImage",
          "metadata": []
        }
      },
      "type": "Microsoft.Web/sites"
    },
    {
      "apiVersion": "2018-02-01",
      "kind": "linux",
      "location": "northeurope",
      "name": "isaacssuperapp-plan",
      "properties": {
        "name": "isaacssuperapp-plan",
        "perSiteScaling": false,
        "reserved": true
      },
[...]

PR #114, opened 2020-04-17
Docker registry support

isaacabraham:
Linked to #48. This doesn't *create* container registries (which actually shouldn't be too difficult) but allows you to more easily link a web app to an existing container registry.


isaacabraham:
I think that this is now good to go.

PR #116, opened 2020-04-17
Rework using CliWrap

Tyrrrz:
Not as clean as one would hope but at least it's net negative code lines :)

Main issues:

- Azure CLI on Windows is a "cmd" so `Process.Start` doesn't find it automatically. So for Windows I search for the full path first before running.
- This has the unfortunate downside that the implementation is tied to the assumption that `az` is a cmd file, so if Microsoft ever decides to ship it as an exe, it will not work (or the code will have to be adapted in both cases).
- Process execution is asynchronous in CliWrap (because executing a process is asynchronous by nature) so it breaks compatibility a bit. I put `Async.RunSynchronously` as a temporary solution.

In any case, I tried. Not sure if the end result is better than the original though.

### DIFF
--- a/src/Farmer/Deploy.fs
+++ b/src/Farmer/Deploy.fs
@@ -21,7 +21,6 @@ module Az =
 
     [<AutoOpen>]
     module AzHelpers =
-        let outputFile = Path.Combine(deployFolder, "output.txt")
         let (|OperatingSystem|_|) platform () =
             if RuntimeInformation.IsOSPlatform platform then Some() else None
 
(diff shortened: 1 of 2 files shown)


isaacabraham:
Thank you very much for taking the time to try this out - I'll think through this. The good thing is that at least this approach doesn't hard-code the path, but scans through the PATH environment variable. Maybe we could adopt that approach but stick with the use of Process.

Tyrrrz:
Yeah that's potentially more flexible. It's not the best use of CliWrap because the hardest part was not in running the process, but in figuring out what to run :D

Issue #120, opened 2020-04-24
Deploy.execute returns Result type - let's have another option that fails directly


forki:
Deploy.execute returns Result type. This is useful if we want to reason about a particular deployment as part of a larger script.
But it would be nice to have another option that just fails directly.

isaacabraham:
Fixed by #122

PR #122, opened 2020-04-25
Add tryExecute method

dpraimeyuu:
Naive (hopefully pragmatic enough) approach for solving #120

### DIFF
--- a/src/Farmer/Deploy.fs
+++ b/src/Farmer/Deploy.fs
@@ -141,7 +141,7 @@ let NoParameters : (string * string) list = []
 
 /// Executes the supplied Deployment against a resource group using the Azure CLI.
 /// If successful, returns a Map of the output keys and values.
-let execute resourceGroupName parameters deployment = result {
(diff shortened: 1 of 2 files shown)


isaacabraham:
First - thanks :-) Second - would it be better to flip this on it's head and have the `execute` version as throwing the exception, and the `try` variant returning a `Result`? @forki - this was your idea, what do you think?

dpraimeyuu:
@isaacabraham I thought about approach you mentioned, but wouldn't it be a breaking change?

isaacabraham:
Whilst we're not at v1, whilst we should obviously look to avoid breaking changes, I'm not too fussed - some of the iterations in the past weeks have been sweeping changes to the deployment API and this would be a relatively minor change.

dpraimeyuu:
Ok, renamed them 👍

isaacabraham:
Just need to fix up the unit test now :-)

isaacabraham:
Ah you were too quick! :-)

isaacabraham:
Thank you :-)

PR #130, opened 2020-04-28
refactored some types

johnazariah:


isaacabraham:
Firstly... thanks for doing this :-)

I tried to merge in your changes but also make some modifications (I didn't think the use of attributes were required and in some ways limited us because you couldn't so easily add an escape hatch for "custom" items). At any rate, it doesn't show the PR as closed even though the code was merged in.

Thus I'm closing this PR, although the code has made it into master (with some tweaks). I'd suggest you fetch the main remote and merge in with your branch before you go much further.

PR #137, opened 2020-04-30
Service bus queue support

isaacabraham:
I think that this is good to go for a first version. @JamesRandall perhaps you want to have a look at it? The sample script and the unit tests show the sort of things it exposes.


isaacabraham:
That's awesome - thank you very much :-)

Issue #140, opened 2020-05-01
Support the new WhatIf feature


dburriss:
Azure CLI recently announced the new `what-if` functionality for ARM templates. It is currently in preview so I understand if we should wait to support this.

But it only took a few minutes to add it in so I thought I would float the idea.

```fsharp
let myRegistry = containerRegistry {
    name "devonRegistry"
    sku Basic
    enable_admin_user
}

let deployment = arm {
    location NorthEurope
    add_resource myRegistry
    output "registry" myRegistry.Name
    output "loginServer" myRegistry.LoginServer
}

deployment
|> Deploy.whatIf "FarmerTest" Deploy.NoParameters // <- NOTE THE CALL to whatif
|> printfn "%A"
```

```console
Compatible version of Azure CLI 2.5.1 detected
Checking Azure CLI logged in status... you are already logged in, nothing to do.
Creating resource group FarmerTest...
"Note: As What-If is currently in preview, the result may contain false positive predictions (noise).
You can help us improve the accuracy of the result by opening an issue here: https://aka.ms/WhatIfIssues.

Resource and property changes are indicated with this symbol:
  + Create

The deployment will update the following scope:

Scope: /subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/FarmerTest

  + Microsoft.ContainerRegistry/registries/devonRegistry [2019-05-01]

      apiVersion:                  "2019-05-01"
      id:                          "/subscriptions/00000000-0000-0000-0000-000000000000/resourceGroups/FarmerTest/providers/Microsoft.ContainerRegistry/registries/devonRegistry"
      location:                    "northeurope"
      name:                        "devonRegistry"
      properties.adminUserEnabled: true
      sku.name:                    "Basic"
      type:                        "Microsoft.ContainerRegistry/registries"

Resource changes: 1 to create.
This command is in preview. It may be changed/removed in a future release.
"
```

PR #141, opened 2020-05-01
Adds whatif capability to Deploy

dburriss:
This feature resolves CompositionalIT/farmer#140

### DIFF
--- a/samples/scripts/container-registry.fsx
+++ b/samples/scripts/container-registry.fsx
@@ -18,4 +18,5 @@ let deployment = arm {
 }
 
 deployment
-|> Deploy.execute "FarmerTest" Deploy.NoParameters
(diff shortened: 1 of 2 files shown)


isaacabraham:
@dburriss is this draft or ready to go?

dburriss:
Well it works. Was more just a question of whether you wanted the feature :)

isaacabraham:
Yes, for sure. We should iterate on it though - in the older (REST) API implementation, I had the code parse through the JSON and give nice F# DUs for the results. Happy to do that later though, unless you want to dig that code out yourself.

dburriss:
Could do. For me this is a good first step. It is just informational. One thing I did not check was error codes. First iteration would be making sure it can be used as a gate. I think it is useful as is so happy to merge in if you happy.

PR #143, opened 2020-05-03
Adds ExpressRoute support

ninjarobot:


isaacabraham:
Great stuff. This will cause a couple of conflicts with the Interfaces branch but I can deal with that. Thanks a lot!

Issue #147, opened 2020-05-08
Support for PostgreSQL service


hravnx:
Currently Farmer has support for SQL Service, but not pg - if I wanted to add support myself (and do a PR, obviously) where would I start? Any guidance?

isaacabraham:
@hravnx Hey! @dburriss is putting together a contributions guide - and we're kind of in the middle of a big refactor of the internals - but for now I would say the best way to start is:

1. Get a hold of the ARM documentation for PG e.g. [here](https://docs.microsoft.com/en-us/azure/templates/microsoft.dbforpostgresql/allversions) and maybe an example e.g. [here](https://github.com/Azure/azure-postgresql/tree/master/arm-templates).
1. Pick a sample and start by seeing if you can write a function that can create an object that maps 1:1 with the JSON. [Here](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Farmer.Resources.fs#L658) is the equivalent for the SQL Server resource.
1. See if you can then plug that into this [helper function](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Builders.Arm.fs#L6) which will allow you to then test out your new resource directly from the Farmer top level.
1. Parameterise and create formal IResources for each ARM resource in the official spec e.g. servers, servers/databases, servers/firewallrules etc.
1. Create a "config" record that represents a logical abstraction across all the ARM resources and implement the IResourceBuilder interface on it.
1. Create a top-level computation expression that makes it even easier to construct.

Totally happy to support you as you go.

dburriss:
@hravnx you can find the current version here.
https://github.com/dburriss/farmer/blob/contributing/contributing.md
I will update it this weekend with the latest refactorings.

isaacabraham:
@dburriss Feel free to merge that into master. I can then keep it upto date in any branches where I make (sometimes sweeping) refactorings without undoing your hard work.

dburriss:
Will do. Updating with the new direction and names from big reactors. Explaining it helps solidify it in my head.

PR #148, opened 2020-05-09
Wip PostgreSQL support.

hravnx:
This is a very very WiP first stab at adding PostgreSQL support to Farmer, as per issue #147 

This is merely intended to get some feedback on approach/direction etc.

One big question I have, is what would the correct testing strategy be here?

I've modelled it closely on the existing MSSql support, but there are no tests for that, or at least none that I could find. Also, there is apparently no client library from Microsoft for PostgreSQl (there is one for SqlServer). So I guess the best we can do is to compare the resulting json template with some expected content? And is there an example of how best to do that?

Or what would you suggest?

### DIFF
--- a/Farmer.sln
+++ b/Farmer.sln
@@ -23,6 +23,7 @@ ProjectSection(SolutionItems) = preProject
 	samples\scripts\vm.fsx = samples\scripts\vm.fsx
 	samples\scripts\webapp-appinsights.fsx = samples\scripts\webapp-appinsights.fsx
 	samples\scripts\webapp-storage.fsx = samples\scripts\webapp-storage.fsx
+	samples\scripts\postgresql.fsx = samples\scripts\postgresql.fsx
 EndProjectSection
 EndProject
 Project("{F2A71F9B-5D33-465A-A702-920D77279786}") = "SampleApp", "samples\SampleApp\SampleApp.fsproj", "{2F230746-189D-4AFD-9B60-D49BB23C98DA}"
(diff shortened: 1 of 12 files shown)


isaacabraham:
Hey. I think, if there's no official MS SDK (and I have doubts as to how much they represent the "golden source of truth" anyway), then yes, I suggest your best bit is to simply compare fields on the generated JSON with expected values. I think that there are a few examples of this in the tests I did, but if not, this is a good opportunity to come up with a good pattern for doing so :-)

dburriss:
@hravnx if I recall correctly I didn't end up using any client reference for the ContainerRegistery tests. I just used straight JObject to inspect. The asserts were overkill but I was experimenting with some ideas to protect the tests from change.

isaacabraham:
The extra benefit we get from using the client libraries is that this proves that the template we emit conforms to Microsoft's own types. Without that, we confirm that we generate the JSON that *we* expect, but not more than that.

isaacabraham:
@hravnx if this is a WIP, can you convert the PR to draft?

hravnx:
I can, and I have

dburriss:
@isaacabraham good point.

hravnx:
So it works now, in that all tests pass, and I have verified that I can create an actual riunning PostgreSQL service through the Azure CLI.

There's a bunch of tests and a sample script in the samples folder.

Support for creating an actual database on the server is still missing, but I wanted to get some feedback before adding it in, and you can create a database by running SQL commands against the instance.

isaacabraham:
Great stuff! Let's get this in :-)

I'll merge this up into the cleanup branch immediately afterwards (which will necessitate moving your code around, but that's fine, I'll do that) and then push that in. We can probably do a release shortly afterwards.

PR #151, opened 2020-05-09
Cleanup

isaacabraham:
Another über PR. Sorry. This one does the following:

1. Splits out all the ARM resources into their own namespace files. This adds a *lot* of files but it will make it easier to navigate and maintain (and reduce risk of merge conflicts).
2. I've rationalised the naming here of the core abstractions based on feedback:
   * The "top level" abstraction is now simply `IBuilder`.
   * The "low leve" one is now called `IArmResource`.
3. All Builders now live in `Farmer.Builders`, not `Farmer.Resources` (this is a breaking change to clients but a simple one to fix).
4. All IArmResources live in `Farmer.Arm`.
5. Added an override to `add_resource` so that you can add an `IArmResource` directly to the arm { } builder.
6. Made a tiny change to the `IArmResource` interface - instead of having a method `ToArmObject` that returns the raw json-ready object, it's now just a property called `JsonValue`.


src/Farmer/Arm/Sql.fs
```diff
@@ -0,0 +1,77 @@
+[<AutoOpen>]
+module Farmer.Arm.Sql
+
+open Farmer
+
+type Server =
```
dburriss, line 6: I noticed this with the Postgres implementation but with AutoOpen and using a generic name like `Server` we are going to run into ambiguity issues. At some stage soon we may want to change this to `SqlServer`. Doesn't have to be this PR though.

src/Farmer/Types.fs
```diff
@@ -13,10 +13,10 @@ type ResourceName =
         | r -> r
     member this.Map mapper = match this with ResourceName r -> ResourceName (mapper r)
 
-/// An Azure ARM resource
-type IResource =
+/// An Azure ARM resource value, which can be serialized to JSON.
+type IArmResource =
     abstract member ResourceName : ResourceName
-    abstract member ToArmObject : unit -> obj
+    abstract member JsonValue : obj
```
dburriss, line 19: I am being pedantic here, so really I am fine with this name :)
`JsonValue` makes me expect a Json string. Sure then why is the type not string?

What about `JsonObject` or `JsonModel`.

Just some suggestions and I think as it is also fine.

src/Farmer/Arm/ContainerRegistry.fs
```diff
@@ -0,0 +1,24 @@
+[<AutoOpen>]
+module Farmer.Arm.ContainerRegistry
+
+open Farmer
+
+type Registries =
```
dburriss, line 6: `Registry` maybe?

src/Farmer/Types.fs
isaacabraham: Either of those work for me - I'll update.

src/Farmer/Arm/Sql.fs
isaacabraham: Mmmm. Perhaps we should turn AutoOpen off for the Arm ones. Not sure. I wanted to keep the names identical to the official names so it aligned with the ARM terminology (including namespaces). Perhaps we will need to rethink this.

src/Farmer/Arm/ContainerRegistry.fs
```diff
@@ -0,0 +1,24 @@
...
+    member this.LoginServer =
+        (sprintf "reference(resourceId('Microsoft.ContainerRegistry/registries', '%s'),'2019-05-01').loginServer" this.Name.Value)
+        |> ArmExpression
+    interface IArmResource with
+        member this.ResourceName = this.Name
+        member this.JsonValue =
+            {| name = this.Name.Value
+               ``type`` = "Microsoft.ContainerRegistry/registries"
+               apiVersion = "2019-05-01"
+               sku = {| name = this.Sku.ToString().Replace("_", ".") |}
```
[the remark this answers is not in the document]
isaacabraham: Funny you mention that - I just noticed this myself. Yes, it is. There another branch I'm working on which accidentally snuck in here (caused by the gitignore file ignoring the ARM folder!). Happy to remove this although it's harmless, and will be overwritten by the other branch if it goes in.

src/Farmer/Arm/ContainerRegistry.fs
isaacabraham: Again, this is based on the official ARM namespace.

src/Farmer/Arm/ContainerRegistry.fs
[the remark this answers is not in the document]
dburriss: Yeah, it's harmless. Just noticed it so I thought I would point it out. No rush.

PR #163, opened 2020-05-21
Implemented "add_unique_key" feature for Cosmos DB cosmosContainer.

JordanMarr:


src/Farmer/Arm/DocumentDb.fs
```diff
@@ -36,6 +38,10 @@ module DatabaseAccounts =
                                   partitionKey =
                                    {| paths = this.PartitionKey.Paths
                                       kind = string this.PartitionKey.Kind |}
+                                  uniqueKeyPolicy = 
+                                   {| uniqueKeys = 
+                                      this.UniqueKeyPolicy.UniqueKeys 
+                                      |> List.map (fun k -> {| paths = k.Paths |}) |}
```
isaacabraham, line 44: Can this be simplified to just `uniqueKeyPolicy = this.UniqueKeyPolicy`?

src/Farmer/Builders/Builders.Cosmos.fs
```diff
@@ -73,6 +81,7 @@ type CosmosDbContainerBuilder() =
         { Name = ResourceName ""
           PartitionKey = [], Hash
           Indexes = []
+          UniqueKeys = []
```
isaacabraham, line 84: Should this be a set or dictionary (do we need to worry about uniqueness of paths or something)?

src/Farmer/Arm/DocumentDb.fs
JordanMarr: I was just manually mapping through the `this.UniqueKeyPolicy.UniqueKeys` to manually create camel cased versions of the children in the object graph (since there is an extra layer beneath that).  I agree that it's a little cumbersome compared to just using a camel case serialization in Newtonsoft.Json, but I was trying to keep it simple and just copy the methodology that was already in-place above.

Btw, here is the desired output:

image

src/Farmer/Builders/Builders.Cosmos.fs
JordanMarr: That's a great point -- duplicate paths would likely be illegal.

src/Farmer/Arm/DocumentDb.fs
JordanMarr: And for good measure, here is what was generated if two unique keys are entered via Azure Portal (and what this currently generates as well):

image

JordanMarr:
I changed UniqueKeys from a List to a Set.  Also added a Cosmos unit test to verify that errant duplicate keys are ignored.

JordanMarr:
I was just doing some more testing on Azure which prompted to me to add an important hint to the "add_unique_key" property: the uniqueness of the key only applies within the logical partition.  If someone doesn't know this, they might think that there was an issue with the ARM template.  This comment will hopefully help with that.

PR #171, opened 2020-05-23
Firewall and DB creation support for PostgreSQL

hravnx:
This PR adds two features to the PostgreSQL support in Farmer

1) Database creation
2) Firewall rules

Both features are modelled closely on the corresponding features in the SQL server support

This is related to issue #147 and would probably enable enough use cases to close it for now


isaacabraham:
Awesome. Thanks so much.

PR #173, opened 2020-05-23
Support for topics and subscriptions

### LINKED ISSUE #165: Support for ServiceBus Topics
Would be nice to be able to create topics
https://docs.microsoft.com/en-us/azure/templates/microsoft.servicebus/2017-04-01/namespaces/topics

Also would be nice to be able to build the service bus namespace separately so it can be passed into both queues & topics.

I'm not sure what the builders should be called, `serviceBus` is currently used for the queue + namespace combination.  Could maybe have `serviceBusTopic`, `serviceBusQueue`, `serviceBusNamespace`?

isaacabraham:
Closes #165. @evilpilaf can you have a look at this from a consumption point of view if it makes sense?

```fsharp
let myServiceBus = serviceBus {
    name "allMyQueues"
    sku Standard
    add_queues [
        queue { name "queuenumberone" }
        queue { name "queuenumbertwo" }
    ]
    add_topics [
        topic {
            name "thetopic"
            add_subscriptions [
                subscription {
                    name "thesub"
                }
            ]
        }
    ]
}
```

The reference docs are a bit hazy around the exact properties permitted for subs and topics but I've done a few tests and they seem to work ok in that topics and subs are now created as well.

### DIFF
--- a/samples/scripts/service-bus.fsx
+++ b/samples/scripts/service-bus.fsx
@@ -4,27 +4,30 @@ open Farmer
 open Farmer.Builders
 open Farmer.ServiceBus
 
-let myQueue = serviceBus {
-    namespace_name "allMyQueues"
+let myServiceBus = serviceBus {
+    name "allMyQueues"
     sku Standard
-
-    name "isaacssuperqueue"
+    add_queues [
+        queue { name "queuenumberone" }
+        queue { name "queuenumbertwo" }
+    ]
+    add_topics [
+        topic {
+            name "thetopic"
+            add_subscriptions [
+                subscription {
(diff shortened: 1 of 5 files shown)


evilpilaf:
This looks pretty good! Can you subscribe a queue to a topic this way?

isaacabraham:
@evilpilaf not yet but I don't see why we couldn't add forwarding support as part of this. The docs for service bus are really shaky - they don't see to match between the reference docs and the walkthroughs - so it was a bit of trial and error to get it working.

I'll merge this one in and we can iterate on from here.

Issue #186, opened 2020-05-28
Support for Azure SignalR Service


sWW26:
Azure SignalR Service
https://docs.microsoft.com/en-us/azure/templates/microsoft.signalrservice/2018-10-01/signalr

PR #188, opened 2020-05-29
Add support for SignalR

sWW26:
Implements #186


isaacabraham:
Wow, that's awesome! :-)

Issue #175, opened 2020-05-23
Smoother workflow when writing docs


hravnx:
At the moment the documentation source lives in a separate branch, imaginatively called `docs`. 

So adding features with docs consists of these steps:

1) Fork/pull master
2) Code feature and tests
3) Commit, push and create PR for feature
4) Pull docs
5) Write docs
6) Commit, push and create PR for feature

While not especially difficult or bad, it does seem suboptimal. More importantly, perhaps, it signals a separation between docs and features, where (IMHO) the two should go together - it makes it much easier to discuss a feature if the docs are part of the PR for that feature. 

I'm not an expert in Hugo, but from looking at the GH actions defines for deploying docs, it seems like the following should be feasible:

1) Move documentation source away from the `docs` branch and into master, in a folder called .... *drumroll* .... `docs`. 
2) Move all Hugo configuration into that folder as well, in effect making the docs folder be a fully self-contained Hugo site.
3) Change the documentation deploy job to be something like:
  a. Copy files from the `docs` folder in master in to the working directory of a CI agent.
  b. Run Hugo build step against that folder
  c. Push the build output into the gh-pages branch (like it is done today)
  
Are there any obvious reasons that I'm missing for why the docs and code are kept separate, and/or something like the above is impossible?

isaacabraham:
The historical reason is: when I started, I had no clue about hugo and started on a separate branch as an experiment. It's still there. I know that when @SeanKilleen did this work we discussed merging them - I'm not sure if there was anything holding us back on that?

Definitely happy to do this as it'll make it easier to keep them in sync.

SeanKilleen:
@isaacabraham I don't think anything is holding us back from doing this. It was one of the options I presented way back when and I think we'd just happened to go a different way (note to self: document decisions like this within the PR in the future :smile: )

@hravnx do you want to take a crack at a PR for it? Otherwise I'm happy to. Docs folder on master is my preference as well.

SeanKilleen:
Found a historical mention of these options: https://github.com/CompositionalIT/farmer/issues/84#issuecomment-609498633

I think the docs were in a docs branch prior to my involvement, and we kept it that way for inertia's sake.

hravnx:
I can certainly take a look at it, and come up with an attempt, but I will likely need help when it comes to testing it.

SeanKilleen:
@hravnx ping me whenever you've got your PR ready and I'll check it out. 👍

isaacabraham:
[...]

PR #189, opened 2020-05-30
Move docs to master branch

hravnx:
As per issue #175 this PR moves the docs off the dedicated branch into master.

Pay special attention to the git submodule workflow - while I have tested it, I'm generally deeply suspicious of submodules, so extra attention is warranted here.


isaacabraham:
@hravnx thank you so much! Does this mean we should now kill off docs completely?

hravnx:
I would suggest closing it to avoid possible confusion.  But it’s probably not a big deal either way.

PR #199, opened 2020-06-05
Add more webapp settings

sWW26:
- Add http20Enabled
- Add clientAffinityEnabled

Is there a new way to update the docs after #189?


isaacabraham:
@sWW26 just make the changes directly in master branch in docs now :-)

isaacabraham:
do it as part of this PR, we can then merge it into as one

sWW26:
noticed another setting I use that wasn't available. `webSocketsEnabled`

PR #201, opened 2020-06-06
Adding private vnet support for container instances

ninjarobot:
* Adds network profile resource for specifying a container subnet IP configuration
* Adds service delegations to vnet subnets to allow delegation to `Microsoft.ContainerInstance/containerGroups`
* Adds optional container group dependency for network profile that indicates which subnet in a vnet to attach the container group

### DIFF
--- a/samples/SampleApp/Program.fs
+++ b/samples/SampleApp/Program.fs
@@ -3,63 +3,10 @@ open Farmer.Builders
 
 //TODO: Create resources here!
 
-//TODO: Testing out the user experience
-
-let privateNetwork = vnet {
-    name "private-vnet"
-    add_address_spaces [
-        "10.30.0.0/16"
-    ]
-    add_subnets [
-        subnet {
-            name "ContainerSubnet"
-            prefix "10.30.19.0/24"
-            add_delegations [
-                SubnetDelegationService.ContainerGroups
-            ]
-        }
-    ]
-}
-
-let aciProfile = networkProfile {
-    name "vnet-aci-profile"
-    vnet "private-vnet"
-    subnet "ContainerSubnet"
-    // Typically just one subnet is needed
-    // but instead they might want to add many, like this
-    (* add_interface_configs [
-        interface_config {
-            add_ipconfigs [
-                ip_config { subnet "ContainerSubnet" }
-            ]
-        }
-    ]*)
-}
-
-
-
-let myContainer = container {
-    name "container1"
-    image "microsoft/aci-helloworld"
-    network_profile "vnet-aci-profile"
-    ports [ 80us ]
-    // the old member (marked obsolete)
-    //ip_address (ContainerGroup.PrivateAddressWithIp (System.Net.IPAddress.Parse "10.30.19.4")) [TCP, 80us]
-    // prefer one of these
-    public_dns "my-container" [TCP, 80us]
-    private_ip [TCP, 80us]
-    private_static_ip "10.30.19.4" [TCP, 80us]
-}
-
 let deployment = arm {
-    location Location.WestUS
-    
+    location Location.NorthEurope
+
     //TODO: Assign resources here using the add_resource keyword
-    add_resources [
-        myContainer
-        aciProfile
-        privateNetwork
-    ]
 }
 
 // Generate the ARM template here...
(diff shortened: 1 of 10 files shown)


samples/SampleApp/Program.fs
```diff
@@ -3,6 +3,48 @@ open Farmer.Builders
 
 //TODO: Create resources here!
 
+//TODO: Testing out the user experience
+
+(*
+// Question - would it make sense to add a separate builder for vnets?
```
ninjarobot, line 9: @isaacabraham I'm trying to figure out a useful way to represent this.  Usually if running container groups in a private vnet, there will be other services attached in that vnet as well.  Does it make sense to create a separate vnet builder for this?

samples/SampleApp/Program.fs
```diff
@@ -3,6 +3,48 @@ open Farmer.Builders
...
+            prefix "10.30.19.0/24"
+            delegations [
+                delegation { serviceName "Microsoft.ContainerInstance/containerGroups" }
+            ]
+        }
+    ]
+}
+*)
+
+let aciProfile = networkProfile {
```
ninjarobot, line 27: Similar to the vnet - should it be possible to create a network profile as a standalone resource?

samples/SampleApp/Program.fs
```diff
@@ -3,6 +3,48 @@ open Farmer.Builders
...
+                ip_config { subnet "ContainerSubnet" }
+            ]
+        }
+    ]*)
+}
+
+let myContainer = container {
+    name "container1"
+    image "aci-hello-world"
+    network_profile "vnet-aci-profile"
```
ninjarobot, line 45: This specifies the network profile, which specifies the subnet where it can be attached.

isaacabraham:
We had a [similar discussion for VMs](https://github.com/CompositionalIT/farmer/pull/190) recently. We fixed that one by allowing you to link a VM to a VNet by either resource name or by supplying the raw IArmResource for the VNet. As you can see from that PR, the IArmResource for VNet is pretty simple to create so I didn't bother creating a builder for it - of course, if you feel it will be useful, go ahead and create a builder for VNet as well - no objection!

samples/SampleApp/Program.fs
```diff
@@ -3,6 +3,56 @@ open Farmer.Builders
...
+
+
+
+let myContainer = container {
+    name "container1"
+    image "aci-hello-world"
+    network_profile "vnet-aci-profile"
+    // the old member (marked obsolete)
+    ip_address (ContainerGroup.PrivateAddressWithIp (System.Net.IPAddress.Parse "10.100.200.3")) [TCP, 80us]
+    // prefer one of these
```
ninjarobot, line 50: You should specify one of these, although whichever one you use last is the one that will take effect.  I'm not sure if this is confusing in that it might imply you can set both, but the alternative is a more verbose builder for the container group `ipAddress` or having to specify an increasingly complicated DU case as shown above.  Opinion on which is preferable?

isaacabraham:
Whilst you're at it, take a quick look at [this issue](https://github.com/Azure/azure-quickstart-templates/issues/2786) - it might (or might not) inform you on how to define the abstractions

isaacabraham:
@ninjarobot once this is ready for review mark it as such and I'll have a look through it :-)

src/Farmer/Common.fs
```diff
@@ -509,4 +509,20 @@ module Maps =
     type Sku = S0 | S1
 
 module SignalR =
-    type Sku = Free | Standard
\ No newline at end of file
+    type Sku = Free | Standard
+  
+type internal IPAddressCidr =
```
ninjarobot, line 514: @isaacabraham - this `IPAddressCidr` was previously under the ExpressRoute builder, but it's helpful elsewhere. What do you think of moving it here?

I'm even thinking of using it in a later PR to take care of assigning IP's in a vnet, e.g. 
```fsharp
let cidr = { Address = IPAddress.Parse "10.100.0.0"; Prefix = 28us }
// Then a sequence you can use for assigning some addresses, skipping the first 2 (network and gateway)
let (assignments:IPAddress seq) = cidr |> IPAddressCidr.toRange |> Seq.skip 2
```

isaacabraham:
Thank you :-)

PR #212, opened 2020-06-12
Changed the sku-part of the web-app part.

jostyposty:
I didn't get the current code in the example to work,


isaacabraham:
Thank you for this :-)

PR #217, opened 2020-06-15
add_reader_policies

forki:


isaacabraham:
Thank you

isaacabraham:
It might also be worth generalising this with a simple `add_policies` keyword and a helper function to create the reader policy. This way you could use that with "simple" reader policies and more complex ones.

PR #225, opened 2020-06-15
Simplify AccessPolicy on KeyVault

isaacabraham:
This removes the short-lived `add_reader_policy` keyword add goes back to just `add_access_policy`. However, I've added a couple of simple helper functions to enable the sample simple behaviour:

```fsharp
keyVault {
    add_access_policy (AccessPolicy.createReader principal)
    add_access_policy (AccessPolicy.create [ Secret.Get ] principal)
}
```

This also works great with `add_access_policies`. Here you can see several different ways of adding a policy to the vault.

```fsharp
keyVault {
        add_access_policies [
            AccessPolicy.createReader principal
            AccessPolicy.create [ Secret.Get ] principal
            accessPolicy { secret_permissions [ Secret.Get ]; object_id principal }
        ]
}
```

@forki does this work for you?

### DIFF
--- a/docs/content/api-overview/resources/keyvault.md
+++ b/docs/content/api-overview/resources/keyvault.md
@@ -59,8 +59,6 @@ The Key Vault builder contains access policies, secrets, and configuration infor
 | disable_recovery_mode | Sets the Creation Mode to Default. |
 | add_access_policy | Adds an access policy to the vault. |
 | add_access_policies | Adds access policies to the vault. |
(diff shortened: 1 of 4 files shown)


forki:
:shipit:

isaacabraham:
TIL there is something called a shipit squirrel.

forki:
image

what should I use here?

forki:
ok `SystemIdentity` seems to be the correct one

PR #229, opened 2020-06-16
User support in KeyVault

isaacabraham:
Adds the ability to list users and groups based on a filter, and retrieve the object ids. These may then be passed into KeyVault as access policies e.g.

```fsharp
let users = [ "user@company.com"; "user2@company.com" ]

let kv = keyVault {
    name "myVault"
    add_access_policies [
        for user in AccessPolicy.findUsers users do
            AccessPolicy.create user.ObjectId
        for group in AccessPolicy.findGroups [ "Developers" ] do
            AccessPolicy.create group.ObjectId
    ]
}
```

cc: @forki - does this work for you?


forki:
I give it 100 squirrels

PR #233, opened 2020-06-17
Adds virtual network gateway support

ninjarobot:
Adds support for virtual network gateways, with a focus on settings required to create one as an ExpressRoute gateway at this time.  Given the defaults, it can work for a simple VPN gateway as well, but more customization will come in later PR's as needed for common VPN configurations.

### DIFF
--- a/docs/content/api-overview/resources/vnet-gateway.md
+++ b/docs/content/api-overview/resources/vnet-gateway.md
@@ -41,7 +41,6 @@ let gw = gateway {
     name "er-gateway"
     vnet "my-vnet" // Must contain a subnet named 'GatewaySubnet'
     er_gateway_sku ErGatewaySku.Standard
-    gateway_ip_config DynamicPrivateIp "gw-pip"
 }
 
 let privateNet = vnet {
--- a/docs/content/api-overview/resources/vnet-gateway.md
+++ b/docs/content/api-overview/resources/vnet-gateway.md
@@ -15,10 +15,12 @@ The Virtual Network Gateway builder creates virtual network gateways for Express
 | Keyword | Purpose |
 |-|-|
 | name | Specifies the name of the virtual network gateway |
-| gateway_type | Sets the type of gateway to create (ExpressRoute or VPN) and its SKU |
-| gateway_ip_config | Specifies the gateway public and private IP addresses |
 | vnet | The name of the virtual network to which the gateway connects |
-| active_active_ip_config | Specifies the second public and private IP configuration for an redundant gateway |
+| er_gateway_sku | SKU for an ExpressRoute gateway |
+| vpn_gateway_sku | SKU for a VPN gateway |
+| vpn_type | Sets the VPN type to route-based (default) or policy-based. |
+| gateway_ip_config | Specifies the gateway public and private IP addresses |
+| active_active_ip_config | Specifies the second public and private IP configuration for a redundant gateway |
 | disable_bgp | BGP is enabled by default, but this can disable it |
 
 #### Example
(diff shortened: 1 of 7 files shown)


isaacabraham:
Thanks :-) Just need some docs added and then this can be merged in :-)

docs/content/api-overview/resources/vnet-gateway.md
```diff
@@ -0,0 +1,52 @@
...
+
+```fsharp
+open Farmer
+open Farmer.Builders
+open Farmer.VirtualNetworkGateway
+
+let gw = gateway {
+    name "er-gateway"
+    vnet "my-vnet" // Must contain a subnet named 'GatewaySubnet'
+    er_gateway_sku ErGatewaySku.Standard
```
ninjarobot, line 36: @isaacabraham thanks for the suggestion to simplify from `gateway_type` to separate methods for setting either an ER or a VPN SKU, that is implemented now.

samples/SampleApp/Program.fs
```diff
@@ -1,41 +1,84 @@
 ﻿open Farmer
```
isaacabraham, line 1: Can you undo changes to this file and instead add a sample script? I'm at fault for committing to this as well - it's useful as a quick scratchpad but not more.

ninjarobot: Yes, I mistakenly committed this - reverted it to the empty sample.

samples/scripts/vnet-gateway.fsx
```diff
@@ -0,0 +1,45 @@
...
+    Location = Location.NorthEurope
+    DomainNameLabel = None
+}
+
+let myGateway = gateway {
+    name "er-gateway"
+    er_gateway_sku ErGatewaySku.Standard
+    vpn_type VpnType.RouteBased
+    vnet "my-vnet"
+    gateway_ip_config DynamicPrivateIp gatewayIp.Name.Value
```
isaacabraham, line 33: Is it worth collapsing the `gatewayIP` into the gateway itself? I'm wondering (and this is lack of knowledge showing here) what configuration is actually being set here - there's no domain label so what is that actually doing?

If it genuinely does need to be a separate type from the user's point of view, consider adding an overload for `gateway_ip_config` to take in a `GatewayIp` rather than needing to specify `Name.Value` (or at the very least, take in the `ResourceName` rather than the raw string and if needed, unwrap it in the builder).

ninjarobot: There are basically 2 scenarios with respect to the public IP:

* ExpressRoute gateway - there's a public IP on the gateway, but you don't even think about it because connectivity is established "automagically" through the ExpressRoute.
* VPN gateway - you've got a VPN in your office / home and you need to point it to the public IP of the gateway to establish the VPN tunnel.

With the VPN gateway, you would probably care about the DNS label here so you can automate it end to end, but not with an ExR gateway like I made in the sample.

src/Farmer/Builders/Builders.VirtualNetworkGateway.fs
```diff
@@ -27,17 +27,26 @@ type VNetGatewayConfig =
     interface IBuilder with
         member this.DependencyName = this.Name
         member this.BuildResources location _ = [
-            { Name = this.Name
-              Location = location
-              IpConfigs = [
-                    yield {| Name = ResourceName "default"; PrivateIpAllocationMethod = this.GatewayPrivateIpAllocationMethod; PublicIpName = this.GatewayPublicIpName |}
+            if this.GatewayPublicIpName = ResourceName.Empty then
```
ninjarobot, line 30: If they don't specify a GatewayPublicIpName, then generate one here - this simplifies what they need to call in the builder.

samples/scripts/vnet-gateway.fsx
```diff
@@ -19,18 +20,20 @@ let privateNet = vnet {
...
+    DomainNameLabel = Some "mygateway"
 }
 
 let myGateway = gateway {
     name "er-gateway"
     er_gateway_sku ErGatewaySku.Standard
     vpn_type VpnType.RouteBased
     vnet "my-vnet"
-    gateway_ip_config DynamicPrivateIp gatewayIp.Name.Value
+    gateway_ip_config DynamicPrivateIp gatewayIp
```
ninjarobot, line 36: @isaacabraham I adjusted it so this takes a PublicIpAddress instance rather than referencing the name.  Combined with your suggestion to generate a public IP when they don't specify one, this simplifies the builder so `gateway_ip_config` doesn't *need* to be called unless they need to customize the IP.

isaacabraham:
Let's get it in and then iterate on it before we hit v1 - if it's good for you now, I'll accept this PR :-)

isaacabraham:
@ninjarobot not sure where this is - ready for another review?

isaacabraham:
In!

Let's consider further improvements that can be made before the next release

PR #235, opened 2020-06-20
Adds resourceId expression function

ninjarobot:
There are a lot of places the `resourceId` template function needs to be generated, and this prevents formatting errors with it.


src/Farmer/Builders/Builders.ResourceId.fs
```diff
@@ -0,0 +1,28 @@
...
+    [<CustomOperation "group">]
+    member __.Group(state:ResourceId, group) = { state with Group = Some group }
+    /// Sets the subscription of the resource
+    [<CustomOperation "subscription_id">]
+    member __.SubscriptionId(state:ResourceId, sub) = { state with SubscriptionId = Some sub }
+let resourceId = ResourceIdBuilder()
+
+type ResourceTypes =
+    static member Connection = "Microsoft.Network/connections"
+    static member VNetGateway = "Microsoft.Network/virtualNetworkGateways"
```
ninjarobot, line 28: Just added a couple for now to show what they look like, but will add the rest for the ARM resources with builders if this pattern works.

isaacabraham:
Nice, this is definitely going to help. I wonder though if this needs a builder etc. - firstly, would it be exposed to end-users? And secondly, I wonder if it could be replaced with a single static method that had optional parameters (or perhaps overloads)?

isaacabraham:
Mmmm. This is probably the first such time we've needed to link to "external" resources. I think there's some overlap with the `ArmExpression` type (basically represents anything that lives in `[ ]` syntax in ARM. Maybe a function that generates one of them for resource ids?

src/Farmer/Types.fs
```diff
@@ -47,7 +56,17 @@ type ArmExpression =
     /// Evaluates the expression for emitting into an ARM template. That is, wraps it in [].
     static member Eval (expression:ArmExpression) = expression.Eval()
     static member Empty = ArmExpression ""
-
+    /// Builds a resourceId ARM expression from the parts of a resource ID.
+    static member ResourceId (resourceType: ResourceType, name:string, ?group:string, ?subscriptionId:string) =
+        match name, group, subscriptionId with
```
ninjarobot, line 61: Removed the builder and using a static member with optional parameters for the group and subscription.

src/Tests/Template.fs
```diff
@@ -135,4 +135,22 @@ let tests = testList "Template" [
 
         Expect.equal b.Dependencies [ ResourceName "a" ] "Dependency should have been set"
     }
+    
+    test "Generates ARM Id with name only" {
+        let rid = ArmExpression.ResourceId (ResourceType.Connection, "test")
+        let id = rid.Eval()
```
ninjarobot, line 141: @isaacabraham what do you think of this instead?  It is a static member to build an `ArmExpression` for a `resourceId`.  I didn't know if this would be better or an additional DU case for it.

src/Farmer/Types.fs
isaacabraham: `ResourceType` as a DU will quickly grow too large I think. Perhaps a single-case DU with some static values e.g.

```fsharp
namespace Farmer

type ResourceType = ResourceType of string

....

// Or maybe in Farmer.Arm.Network?
module Farmer.Builders.Network.ResourceTypes

let connection = ResourceType "Microsoft.Network/connections"
```

isaacabraham: Alternatively we could remove ResourceType completely and just make it a bunch of strings?

isaacabraham: Yet another option is to put these as static member on the appropriate IArmResource?

src/Tests/Template.fs
isaacabraham: I think this is a more succinct fit. Not sure about the DU though - I think an open-ended set might be easier to work with.

src/Farmer/Types.fs
ninjarobot: I had a bunch of string constants originally, but then the signature was all string parameters so it was really easy to forget which of the strings was the name and which was the type, as in `ArmExpression.ResourceId("name-of-resource", "Microsoft.Whatever/things")`.  Since I personally made that mistake even though I'm writing the method itself, I thought some type safety was in order :)

ninjarobot: `type ResourceType = ResourceType of string` seems like a good middle ground and makes the signature of `resourceId` more apparent.

ninjarobot: I changed it to make the `ResourceType` more flexible like you suggested.  The resource type literals seemed to make the most sense been in with the code that creates the JSON itself, to keep it close to the payload that needs it.  I also used tried out the usage there as well as in the tests.  Thoughts?

isaacabraham:
Thanks as usual. Merged - will apply to existing templates

PR #236, opened 2020-06-21
IPAddressCidr subnet carving utilities

ninjarobot:
This gives us some helpful capabilities for dividing up a virtual network address space into subnets of various sizes using CIDR notation.


isaacabraham:
Nice, thanks!

PR #237, opened 2020-06-21
Break out container group and instance again.

isaacabraham:
Based on our chat on Friday @ninjarobot, this splits container groups and instances back out. I've looked at the naming and the term "container group" actually rarely shows up in the documentation - the service (somewhat misleadingly) is called Container Instance, or Container Instances!

I'm wondering whether it makes sense to rename the "group" one to `containerInstance` and the child to `containerImage` or similar?

### DIFF
--- a/samples/scripts/container-instance.fsx
+++ b/samples/scripts/container-instance.fsx
@@ -1,0 +1,36 @@
+#r @"./libs/Newtonsoft.Json.dll"
+#r @"../../src/Farmer/bin/Debug/netstandard2.0/Farmer.dll"
+
+open Farmer
+open Farmer.Arm.ContainerInstance
+open Farmer.Builders
+
+let template =
+    createTemplate
+        Location.WestEurope [
+            containerGroup {
+                name "isaac-container-group"
+                operating_system Linux
+                restart_policy ContainerGroup.AlwaysRestart
+                add_instances [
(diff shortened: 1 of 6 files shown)


isaacabraham:
Also @ninjarobot in my testing, I noticed that you are not permitted to create a TCP port on the group which doesn't exist on a container instance. Therefore I've made it so that ports are automatically from the container instances cascaded to the group and you can no longer set them at the group level. However, I've now realised that there are also PrivateIp and Private Static IP modes for the group which could impact this, so this last change might be a mistake - what are your thoughts?

With a container instance, would you ever create an IP address on the instance and *not* want to expose it at the group level?

isaacabraham:
Perhaps we can create a simple abstraction for creating ports on containers where you specify if the port should be public or internal. This way we avoid the need to manually map the ports, and it's much more explicit if the ports are intended to be public or not.

isaacabraham:
That's now applied, so you add ports to a container like so:

```fsharp
containerInstance {
    name "foo"

    add_ports PublicPort [ 123us ]
    add_ports InternalPort [ 456us ]

    add_public_ports [ 123us ]
    add_internal_ports [ 456us ]
}
```

Which style do you prefer - having two methods or a just method which is parameterised by the port accessibility?

PR #240, opened 2020-06-23
Fix setter for sku

gmanvel:
I'm an amateur in F#, Visual Studio doesn't like `sku WebApp.B1`, looks like it should be `sku WebApp.Sku.B1`


isaacabraham:
Thank you for fixing this!

PR #246, opened 2020-06-25
Environment variable support for container instances

### LINKED ISSUE #124: Support environment variables for container groups
Hi there, how to add `environment variables` for container groups?

Thanks

ninjarobot:
Fixes #124


isaacabraham:
I would force the secret env vars to be set as parameters - look at the webapp where we allow passing "secret" settings https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Builders/Builders.WebApp.fs#L375

isaacabraham:
Thank you! I also realised I hadn't updated the docs - thank you for doing that as well.

PR #251, opened 2020-06-26
Separate CDN resources.

isaacabraham:


isaacabraham:
OK, I've tested all the modified resources so this is going in. The changes at the public API level are minimal.

Issue #264, opened 2020-07-01
CosmosDbConfig.PrimaryKey returns a string that is incompatible with ArmExpression at runtime


bobbydazzler:
I'm trying to add a Cosmos PrimaryKey as a Secret in KeyVault, however the secret deployed to KeyVault is the arm template string, e.g., "[listKeys(resourceId('Microsoft.DocumentDB/databaseAccounts', 'local-db-server'), providers('Microsoft.DocumentDB','databaseAccounts').apiVersions[0]).primaryMasterKey]", rather than the primary key value itself.

```
    test "Generate Cosmos+KeyVault" {
        let db = cosmosDb {
            name "local-db"
        }

        let primaryKeySecret = secret {
            name "pk"
            value (db.PrimaryKey |> ArmExpression) // is the the correct way to consume the property?
        }

        let kv = keyVault {
            name "kv"
            add_secret primaryKeySecret
        }

        let deployment = arm {
            location (Location.UKSouth)
            add_resource kv
        }

        let str =
            deployment.Template
            |> Writer.toJson

        printfn "%O" (db.PrimaryKey |> ArmExpression)
        printfn "%O" (primaryKeySecret.Value)
        printfn "%O" (kv.Secrets.Head.Value)
        printfn "%s" str
    }
```

Produces the following outputs:

```
ArmExpression
  "[listKeys(resourceId('Microsoft.DocumentDB/databaseAccounts', 'local-db-server'), providers('Microsoft.DocumentDB','databaseAccounts').apiVersions[0]).primaryMasterKey]"
ExpressionSecret
  (ArmExpression
     "[listKeys(resourceId('Microsoft.DocumentDB/databaseAccounts', 'local-db-server'), providers('Microsoft.DocumentDB','databaseAccounts').apiVersions[0]).primaryMasterKey]")
ExpressionSecret
  (ArmExpression
     "[listKeys(resourceId('Microsoft.DocumentDB/databaseAccounts', 'local-db-server'), providers('Microsoft.DocumentDB','databaseAccounts').apiVersions[0]).primaryMasterKey]")
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "outputs": {},
  "parameters": {},
  "resources": [
    {
      "apiVersion": "2018-02-14",
      "location": "uksouth",
      "name": "kv",
      "properties": {
        "accessPolicies": [],
[...]

PR #267, opened 2020-07-02
Ensure ARM Expressions don't double-wrap.

isaacabraham:
Stop allowing us to directly create `ARMExpression` values and instead force us to use a factory method. Also fixed the SignalR key at the same time.


isaacabraham:
Also related to #264

PR #268, opened 2020-07-02
implement file share

### LINKED ISSUE #242: Add support for File Shares to Storage Account
@ThisFunctionalTom mentioned that this would be a useful feature to add and he's going to have a crack at it.

ThisFunctionalTom:
Add support for file shares to storage account.

Closes #242

### DIFF
--- a/src/Farmer/Builders/Builders.Storage.fs
+++ b/src/Farmer/Builders/Builders.Storage.fs
@@ -6,6 +6,7 @@ open Farmer.CoreTypes
 open Farmer.Storage
 open Farmer.Arm.Storage
 open BlobServices
+open FileShares
 
 let internal buildKey (ResourceName name) =
     sprintf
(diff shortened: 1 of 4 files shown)


src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -14,13 +15,17 @@ let internal buildKey (ResourceName name) =
...
 type StorageAccountConfig =
     { /// The name of the storage account.
       Name : ResourceName
       /// The sku of the storage account.
       Sku : Sku
       /// Containers for the storage account.
-      Containers : (ResourceName * StorageContainerAccess) list}
+      Containers : (ResourceName * StorageContainerAccess) list
+      /// File shares
+      FileShares: (ResourceName * ShareQuotaInGb option) list }
```
ninjarobot, line 28: Would it be possible to give a `FileShare` its own independent model so it could be referenced from other resources? For example, if we wanted to add a file share to a storage account and then also reference it in an Azure Container Group, having a separate model makes it possible to reference the resource in both.

isaacabraham: For simple types like this, I think that a nice way to do that could be to change from a tuple to a full record. It would then be available from the `StorageAccountConfig` as a standard field and could be referenced wherever you want. If you want to avoid the record syntax, a static member on the type might be enough e.g. `create(resourceName:string, ?ShareQuoteInGb:int)`.

isaacabraham:
Thank you for this!

PR #271, opened 2020-07-04
Move over simple resourceids to factory method.

isaacabraham:
@ninjarobot I've created an overload of the resource id builder which works for multiple resource names rather groups etc. and applied it around. I've also added tests for as many resources as I could. The main one I've not done it for is network resources / builders. Could you either have a look at the changes to that bunch of resources just to confirm I've not broken anything, and / or write some basic tests?


isaacabraham:
I also added a tiny refactoring to remove the use of `IsSome` and `.Value` (shock!).

PR #276, opened 2020-07-06
Azure functions default version - follow portal behavior

Dzoukr:
Currently, when you click through Azure portal to create Function app, there is no way how to select lower than current (~3) version (can be changed later in configuration, but not during creation). Maybe Farmer should follow that logic - the default is the latest one? What do you think @isaacabraham?


isaacabraham:
This makes sense to me. Something else you could do is just make the field optional in the Builder (and the IArmResource) - if it's left at None, then don't set the property at all and let ARM figure out what to do. Obviously this assumes that the version property is indeed optional at the ARM template level.

The benefit of that approach is that when v4 comes out, we won't have to do anything.

isaacabraham:
@Dzoukr Let me know which way you want to go - happy to go with what we have or for you to make the suggested changes above.

Dzoukr:
I like your approach a little bit more, but I just checked how default ARM template looks like when clicking in portal and the first value of appSettings is runtime version:

```
"appSettings": [
   {
       "name": "FUNCTIONS_EXTENSION_VERSION",
       "value": "~3"
   }
]
```

So Portal does the same as my PR - if not directly set, it silently sets the latest one. Not sure, what would happen without specifing version on ARM level.

isaacabraham:
Try it :-)

Dzoukr:
Ok, now we know. It fallbacks to version 1, if version missing in appSettings.

image

So yes, our current approach is the best we can do now.

PR #281, opened 2020-07-12
Adds network security group support

ninjarobot:
Adds support for generating ARM template content for network security groups and their rules.

### DIFF
--- a/docs/content/api-overview/resources/nsg.md
+++ b/docs/content/api-overview/resources/nsg.md
@@ -1,0 +1,117 @@
+---
+title: "Network Security Group"
+date: 2020-07-19T16:44:00-04:00
+weight: 5
+chapter: false
+---
+
+#### Overview
+The Network Security Group builder creates network security groups with rules for securing network access to resources.
+
+* Network Security Groups (`Microsoft.Network/networkSecurityGroups`)
+* Security Rules (`Microsoft.Network/networkSecurityGroups/securityRules`)
+
+#### Builder Keywords
+
+| Applies To | Keyword | Purpose |
+|-|-|-|
+| nsg | name | Specifies the name of the network security group |
+| nsg | add_rules | Adds security rules to the network security group |
+| securityRule | name | The name of the security rule |
+| securityRule | description | The description  of the security rule |
+| securityRule | service | The service port(s) and protocol(s) protected by this security rule |
+| securityRule | add_source | Specify access from any source protocol, address, and port |
+| securityRule | add_source_any | Specify access from any address and any port |
+| securityRule | add_source_address | Specify access from a specific address and any port |
+| securityRule | add_source_network | Specify access from a specific network and any port |
+| securityRule | add_source_tag | Specify access from a tagged source such as "Internet", "VirtualNetwork", or "AzureLoadBalancer" |
+| securityRule | add_destination | Specify access to any source protocol, address, and port |
+| securityRule | add_destination_any | Specify access to any address and any port |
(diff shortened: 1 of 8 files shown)


docs/content/api-overview/resources/nsg.md
```diff
@@ -0,0 +1,117 @@
...
+
+```fsharp
+open Farmer
+open Farmer.Builders
+open Farmer.NetworkSecurity
+
+// Create a rule for https services accessible from the internet
+let httpsAccess = securityRule {
+    name "web-servers"
+    service (Service ("https", Port 443us))
```
isaacabraham, line 46: Can probably add an overload here to simply take in string and port number?

isaacabraham: Also consider replacing "https" with a DU (HTTP | HTTPS) (if we don't already have one). Or could we can Service itself to being a union of HTTP / HTTPS? Or, have two keywords, `http_service` and `https_service`.

src/Farmer/Arm/NetworkSecurityGroup.fs
```diff
@@ -0,0 +1,63 @@
...
+    }
+    interface IArmResource with
+        member this.ResourceName = this.Name
+        member this.JsonModel =
+            {| ``type`` = securityRules.ArmValue
+               apiVersion = "2020-04-01"
+               name = sprintf "%s/%s" this.SecurityGroup.Name.Value this.Name.Value
+               dependsOn = [ ArmExpression.resourceId(networkSecurityGroups, this.SecurityGroup.Name).Eval() ]
+               properties =
+                    {| description = this.Description |> Option.defaultValue null
```
isaacabraham, line 49: Replace with `Option.toObj`

src/Farmer/Arm/NetworkSecurityGroup.fs
```diff
@@ -0,0 +1,63 @@
...
+               name = this.Name.Value
+               location = this.Location.ArmValue
+            |} :> _
+
+type SecurityRule =
+    { Name : ResourceName
+      Description : string option
+      SecurityGroup : NetworkSecurityGroup
+      Protocol : NetworkProtocol
+      SourcePort : string // Supports a string, like '*'
```
isaacabraham, line 29: Consider making these unions - e.g. `Wildcard | Tag of string`?

docs/content/api-overview/resources/nsg.md
isaacabraham: Ignore that comment - I can see that the HTTPS thing is just the name of the service! However the comment on the overload / tuple is still valid.

ninjarobot: I added an overload accepting a tuple to make that it simpler when adding a single port service.

src/Farmer/Arm/NetworkSecurityGroup.fs
ninjarobot: Good reminder, done.

src/Farmer/Arm/NetworkSecurityGroup.fs
ninjarobot: I adjusted them to give the security rule better type safety.

isaacabraham:
I've merged but I think we should set up a call when you have time. I need your help on the IP stuff but also looking through this in more detail I think there's some re-modelling of the domain that can be done - the `DestinationPort` `DestinationPorts` pattern looks at first glance like a union.

isaacabraham:
Another question: Are the names of the rule services needed? Doesn't *look* like they're used anywhere.

Issue #280, opened 2020-07-08
Better integration for VMs and Gateways


isaacabraham:
I've cleaned up some of this so you can more easily connect a VM to a Gateway:

```fsharp
let developmentEnvironment = vm {
    name "dev-jan"
    username "jan"
    vm_size Vm.Standard_F8s_v2
    operating_system Vm.WindowsServer_2019Datacenter
    os_disk 128 Vm.StandardSSD_LRS
    add_ssd_disk 512
}

let gw = gateway {
    name (developmentEnvironment.Name.Map(sprintf "%s-gateway"))
    gateway_ip_config DynamicPrivateIp developmentEnvironment.IpName
    vnet developmentEnvironment.VNetName.ResourceName
    vpn_gateway_sku VpnGatewaySku.Basic
}
```

However we could go further:

```fsharp
let gw = gateway {
    link_to_vm developmentEnvironment // does the same as the name, gateway_ip_config and vnet keywords above
    vpn_gateway_sku VpnGatewaySku.Basic
}
```

Or even

```fsharp
let developmentEnvironment = vm {
    name "dev-jan"
    username "jan"
    vm_size Vm.Standard_F8s_v2
    operating_system Vm.WindowsServer_2019Datacenter
    os_disk 128 Vm.StandardSSD_LRS
    add_ssd_disk 512
    use_gateway VpnGatewaySku.Basic // don't need a separate gateway builder, just turn it on with sane defaults here...
}
```

Question is, is this the "right" thing to do? Are the defaults above (especially the choice of DynamicPrivateIp) sane? I don't mind if it doesn't suit everyone - I'm thinking of 80% of the people that want a gateway.

@Jandev As a user of Farmer and someone who just did something like this, which of the three approaches above would you find most logical - we could even include all three to fit all use cases.

isaacabraham:
Also cc: @ninjarobot incase you have some thoughts?

Jandev:
If I was to choose between these three options, I'd go for the third one.

The current `vm`-builder adds a public IP to the virtual machine. While this is the default behavior, it's not recommended to expose machines to the internet without any security in place. I get why it's developed like this now, but seeing the gateway is supported now, I think it's better to integrate this stuff with the builder and have a better solution out of the box.  

That said, I think people in actual projects will probably not use the builders a lot as they want to be 'in control' to what gets created.  
Having the possibility to do the first option will probably be something I'd use in an actual project. That way I know what gets created and can finetune it when necessary.

[...]

PR #287, opened 2020-07-19
Improve VNet, Gateway and VM integration.

isaacabraham:
Ignore the accidental commits on `Program.fs` and the `vm.fsx`, I've made a few changes that I think ease working with VNets, Gateways and VMs:

1. Public IP is now optional so you can have a purely private VM.
1. Merge both sku keywords into a single, overloaded `sku` keyword. 
1. Gateway can now be configured with a VNet directly (instead of the name).
1. VNet commands on the VM are now prefixed with `vnet_` for clarity.

Plus some formatting refactoring.

I was hoping to do something more ambitious around them all but I think this is a decent start:

```fsharp
let devVn = vnet {
    name "dev-isaac-vnet"
    add_address_spaces [ "10.0.0.0/16" ]
    add_subnets [
        subnet { name "GatewaySubnet"; prefix "10.0.0.0/24" }
        subnet { name "Frontend"; prefix "10.0.2.0/24" }
    ]
}

let developmentEnvironment = vm {
    name "test-isaac"
    username "isaac"
    no_public_ip
    link_to_vnet devVn
    vnet_subnet_name "Frontend"
}

let gw = gateway {
    name "dev-isaac-gateway"
    vnet devVn
    sku VpnGatewaySku.Basic
}

let deployment = arm {
    add_resource developmentEnvironment
    add_resource gw
    add_resource devVn
}
```

* I would still like it to be easier to create all of these resources together - or at least the VNet and Gateway.
* I don't like the work required to hook up the gateway subnet to the Vnet - you need to know the magic "GatewaySubnet" name (which is nuts) and have to manually connect the vnet subnet to the VM.

This is all boilerplate which should be doable by a builder, IMHO. But this can be a future enhancement.

@ninjarobot could you give this a quick once over?

Thanks

### DIFF
--- a/src/Farmer/Builders/Builders.Vm.fs
+++ b/src/Farmer/Builders/Builders.Vm.fs
@@ -27,103 +27,108 @@ type VmConfig =
       CustomScriptFiles : Uri list
 
       DomainNamePrefix : string option
+      UsePublicIp : bool
 
-      VNetName : ResourceRef
       AddressPrefix : string
       SubnetPrefix : string
+      VNetName : ResourceRef
       SubnetName : ResourceName option
 
       DependsOn : ResourceName list }
 
     member this.NicName = makeResourceName this.Name "nic"
-    member this.IpName = makeResourceName this.Name "ip"
-    member this.Hostname = sprintf "reference('%s').dnsSettings.fqdn" this.IpName.Value |> ArmExpression.create
+    member this.IpName = if this.UsePublicIp then Some (makeResourceName this.Name "ip") else None
+    member this.Hostname =
+        this.IpName
+        |> Option.map(fun name ->
+            sprintf "reference('%s').dnsSettings.fqdn" name.Value |> ArmExpression.create)
 
     interface IBuilder with
         member this.DependencyName = this.Name
         member this.BuildResources location = [
-            // VM itself
-            { Name = this.Name
-              Location = location
-              StorageAccount =
-                this.DiagnosticsStorageAccount
-                |> Option.bind(function
-                    | AutomaticPlaceholder -> None
-                    | AutomaticallyCreated r -> Some r
-                    | External r -> Some r)
-              NetworkInterfaceName = this.NicName
-              Size = this.Size
-              Credentials =
-                match this.Username with
-                | Some username ->
-                    {| Username = username
-                       Password = SecureParameter (sprintf "password-for-%s" this.Name.Value) |}
+            match this.SubnetName with
+            | None ->
+                failwith "Subnet Name must be set"
+            | Some subnetName ->
+                // VM itself
+                { Name = this.Name
+                  Location = location
+                  StorageAccount =
+                    this.DiagnosticsStorageAccount
+                    |> Option.bind(function
+                        | AutomaticPlaceholder -> None
+                        | AutomaticallyCreated r -> Some r
+                        | External r -> Some r)
+                  NetworkInterfaceName = this.NicName
+                  Size = this.Size
+                  Credentials =
+                    match this.Username with
+                    | Some username ->
+                        {| Username = username
+                           Password = SecureParameter (sprintf "password-for-%s" this.Name.Value) |}
+                    | None ->
+                        failwithf "You must specify a username for virtual machine %s" this.Name.Value
+                  Image = this.Image
+                  OsDisk = this.OsDisk
+                  DataDisks = this.DataDisks }
+
+                // Custom Script
+                match this.CustomScript with
+                | Some script ->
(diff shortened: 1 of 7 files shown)


isaacabraham:
Closes #280

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -27,103 +27,108 @@ type VmConfig =
       CustomScriptFiles : Uri list
 
       DomainNamePrefix : string option
+      UsePublicIp : bool
 
-      VNetName : ResourceRef
       AddressPrefix : string
       SubnetPrefix : string
```
ninjarobot, line 33: Maybe these should be `IPAddressCidr` instead of string, and then it would be possible to validate that the subnet is within the address space or to carve a subnet out of the address space.

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -27,103 +27,108 @@ type VmConfig =
...
+                // VNET
+                match this.VNetName with
+                | AutomaticallyCreated _
+                | AutomaticPlaceholder ->
+                    { Name = this.VNetName.ResourceName
+                      Location = location
+                      AddressSpacePrefixes = [ this.AddressPrefix ]
+                      Subnets = [
+                          {| Name = subnetName
+                             Prefix = this.SubnetPrefix
```
ninjarobot, line 105: If generating it, it is probably a useful validation to make sure the subnet is within the address prefixes on the vnet so they don't get a confusing error about it. Could also simplify slightly and make `AddressPrefix` an `option` so if they don't happen to pass one, use the `SubnetPrefix` as the `AddressPrefix`.

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -203,26 +219,32 @@ type VirtualMachineBuilder() =
...
-    /// Sets the IP address prefix of the VM.
-    [<CustomOperation "address_prefix">]
+    /// Sets the address prefix of the vnet of the VM, unless you are linking to an external vnet.
+    [<CustomOperation "vnet_address_prefix">]
     member __.AddressPrefix(state:VmConfig, prefix) = { state with AddressPrefix = prefix }
-    /// Sets the subnet prefix of the VM.
-    [<CustomOperation "subnet_prefix">]
+    /// Sets the subnet prefix of the vnet of the VM, unless you are linking to an external vnet.
+    [<CustomOperation "vnet_subnet_prefix">]
     member __.SubnetPrefix(state:VmConfig, prefix) = { state with SubnetPrefix = prefix }
```
ninjarobot, line 227: This is a little confusing to set by itself as is because if someone _only_ set the `vnet_subnet_prefix` then they would have to be aware of code internals that `vnet_address_prefix` defaults to "10.0.0.0/16".  I could see someone doing this:
```fsharp
let myVm = vm {
    vnet_subnet_prefix "192.168.4.0/24"
    // other stuff, but didn't set `vnet_address_space`
}
```
and this will mean the generated subnet is invalid because the subnet isn't contained in the vnet's address space.

src/Farmer/Builders/Builders.Vm.fs
isaacabraham: I would need your help on this. I actually need to better understand the IP Address space stuff - particularly from the subnet point of view.

src/Farmer/Builders/Builders.Vm.fs
isaacabraham: I suppose people should be aware to change the address prefix as well if they are changing the subnet prefix though? Or should we validate this somehow?

src/Farmer/Builders/Builders.Vm.fs
ninjarobot: This is a [helpful tool](http://www.davidc.net/sites/default/subnets/subnets.html) that can help explain how subnets need to fit into an address space. It even will [save a view](http://www.davidc.net/sites/default/subnets/subnets.html?network=192.168.0.0&mask=16&division=13.db00) after carving out some subnets.

Whatever subnet they use for the VM just needs to be contained within the vnet's address space.  Honestly if doing it on the VM builder like this, if they don't specifically set the `AddressPrefix`, maybe we should just use the same value as the `SubnetPrefix`.

src/Farmer/Builders/Builders.Vm.fs
ninjarobot: The thing about that is that they don't know what default was used, so if they set the `vnet_subnet_prefix` and it didn't happen to fall within "10.0.0.0 - 10.0.255.255" (the default address space) they'll get a deployment error.

I just opened #291 with a utility function to validate this.

PR #293, opened 2020-07-24
Adds basic AKS support, vnet attached AKS, and CNI

ninjarobot:
Builds an AKS cluster with customization over some common settings. Example usage, looking for feedback here:

```fsharp
let myAks = aks {
    name "k8s-cluster"
    dns_prefix "alotta-containers"
    add_agent_pools [
        agentPool {
            name "linuxPool"
            count 3
        }
    ]
    linux_profile "dave" ["public-key-here"]
    sp_client_id "some-spn-client-id"
}
```

Windows pools will need a little more work because they have different container networking requirements so these are created as linux by default.


src/Farmer/Builders/Builders.Aks.fs
```diff
@@ -0,0 +1,118 @@
...
+    [<CustomOperation "add_agent_pools">]
+    member _.AddAgentPools(state:AksConfig, pools) = { state with AgentPools = state.AgentPools @ pools }
+    /// Adds an agent pool to the AKS cluster.
+    [<CustomOperation "add_agent_pool">]
+    member _.AddAgentPool(state:AksConfig, pool) = { state with AgentPools = state.AgentPools @ [ pool ] }
+    /// Sets the linux profile for the AKS cluster.
+    [<CustomOperation "linux_profile">]
+    member _.LinuxProfile(state:AksConfig, username:string, sshKeys:string list) = { state with LinuxProfile = Some (username, sshKeys) }
+    /// Sets the client id of the service principal for the AKS cluster.
+    [<CustomOperation "sp_client_id">]
```
isaacabraham, line 111: Can we have a more human-readable name?

src/Farmer/Arm/Aks.fs
```diff
@@ -0,0 +1,74 @@
+[<AutoOpen>]
+module Farmer.Arm.Aks
```
isaacabraham, line 2: The namespace should be (I think!) `Farmer.ContainerService` (to mirror the namespace of the resource.

src/Farmer/Builders/Builders.Aks.fs
ninjarobot: Yes, but any suggestions?  It's the clientId on the [ManagedClusterServicePrincipalProfile](https://docs.microsoft.com/en-us/azure/templates/microsoft.containerservice/managedclusters#managedclusterserviceprincipalprofile-object) which is quite a mouthful. Is `service_principal_client_id` ok?

src/Farmer/Arm/Aks.fs
ninjarobot: I can change it to that. I called it `Aks` only because that's how its referenced in the [docs](https://docs.microsoft.com/en-us/azure/templates/). I think they adopted that name to differentiate from the [now deprecated Azure Container Service](https://azure.microsoft.com/en-us/updates/azure-container-service-will-retire-on-january-31-2020/) that was replaced by AKS.

ninjarobot: I changed it to `ContainerService` as suggested.

src/Farmer/Builders/Builders.Aks.fs
ninjarobot: Went with `service_principal_client_id`.

src/Farmer/Arm/ContainerService.fs
```diff
@@ -0,0 +1,88 @@
...
+
+open Farmer
+open Farmer.CoreTypes
+open Farmer.Vm
+
+let managedClusters = ResourceType "Microsoft.ContainerService/managedClusters"
+
+type AgentPoolMode = System | User
+
+type ManagedCluster =
```
Aaronontheweb, line 12: Looks good - there's a lot of bells and whistles for AKS but this gets the big stuff

src/Farmer/Builders/Builders.ContainerService.fs
```diff
@@ -0,0 +1,127 @@
...
+        ]
+
+type AgentPoolBuilder() =
+    member _.Yield _ =
+        { Name = ResourceName.Empty
+          Count = 1
+          Mode = System
+          OsDiskSize = 0<Gb>
+          OsType = OS.Linux
+          VmSize = Standard_DS2_v2
```
Aaronontheweb, line 61: That's a good default size

src/Farmer/Builders/Builders.ContainerService.fs
```diff
@@ -0,0 +1,127 @@
...
+              WindowsProfile =
+                  this.WindowsProfileAdminUserName
+                  |> Option.map (fun username -> {| AdminUserName = username; AdminPassword = SecureParameter (sprintf "admin-password-for-%s" this.Name.Value) |})
+            }
+        ]
+
+type AgentPoolBuilder() =
+    member _.Yield _ =
+        { Name = ResourceName.Empty
+          Count = 1
```
Aaronontheweb, line 57: Nitpicking, but I'd default to 2 nodes - lots of interesting edge cases start popping up inside K8s cluster use as soon as node count > 1

src/Tests/ContainerService.fs
```diff
@@ -0,0 +1,42 @@
...
+open Farmer
+open Microsoft.Azure.Management.Compute.Models
+open Microsoft.Azure.Management.ContainerService
+open Microsoft.Rest
+open System
+
+let dummyClient = new ContainerServiceClient (Uri "http://management.azure.com", TokenCredentials "NotNullOrWhiteSpace")
+
+let tests = testList "AKS" [
+    test "Basic AKS cluster" {
```
Aaronontheweb, line 14: Looks very straight forward - I like it!

src/Farmer/Builders/Builders.ContainerService.fs
ninjarobot: I went with what the API defaults are for all of these so it's about the same behavior as if you didn't specify these. The `OsDiskSize` is a little interesting: 

> If you specify 0, it will apply the default osDisk size according to the vmSize specified.

src/Farmer/Builders/Builders.ContainerService.fs
ninjarobot: I agree, that's usually a practical minimum, but going with the default behavior from the API:

> Number of agents (VMs) to host docker containers. Allowed values must be in the range of 0 to 100 (inclusive) for user pools and in the range of 1 to 100 (inclusive) for system pools. The default value is 1.

src/Tests/ContainerService.fs
ninjarobot: Thank you, it's a really short way of saying [this](https://github.com/Azure/azure-quickstart-templates/blob/master/101-aks/azuredeploy.json#L86-L118) 😄

Aaronontheweb:
The VNET support is definitely a must-have - great addition. I think this is a really great start for AKS support - the only other "must have" thing I can think of is some way of linking an ACR instance together with AKS (so private application images can be pulled) - that may not be achievable via ARM though since I think you have to create a Kubernetes secret with the ACR credentials once the cluster is up and running in order to accomplish that. This can be done via the `az` CLI easily enough.

isaacabraham:
@Aaronontheweb we already have the idea of "post deployment" hooks that behind the scenes make calls to the CLI. we could do the same thing here.

isaacabraham:
@ninjarobot is `azureCniNetworkProfile` used elsewhere - could you it into the omit it so that the network profile takes in parameters that it needs directly?

isaacabraham:
@ninjarobot no concerns - I'll be releasing the full 1.0 today so this can be merged in shortly after that and put in for 1.1

isaacabraham:
@ninjarobot ok this is now in :-) Would you be able to create the docs page as well?

PR #299, opened 2020-08-04
Fix incorrect example code in documention for Azure Postgres

tdwhite0:
Small change but this tripped me up for a while.


isaacabraham:
Thanks - these changes may be small but they are really important. Thanks.

Issue #298, opened 2020-08-03
Support for adding resource tags


tdwhite0:
Unless I'm mistaken it doesn't look like there is a way right now to add resource tags to the generated template like shown here:

https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/template-tutorial-add-tags?tabs=azure-powershell

This would be a nice feature to have, I'm thinking about building it myself. Does anyone have any ideas on the best way to implement?

TheRSP:
My company also need support for tags. I was wondering if adding `tag`/`tags` keywords to each of the builders would be sufficient:
```f#
webApp {
  name "my-web-app"

  tag "tier" "UI"
  tags [
    ("Favourite Colour", "Purple")
    ("Favourite Animal", "Mantis Shrimp")
  ]
}
```

isaacabraham:
Yes, that seems reasonable. Only question is that the builder would need to decide which resources were given the tags - I suppose for some builders this would be different.

TheRSP:
Yes, I had assumed in the above example that the tags would apply to the "main" resource (the website) but we actually have three options:
1. apply to the main resource
2. apply to all resources
3. have some way of specifying to which resources to apply the tags
  a. by using different keywords (`site_tags key value`, `server_farm_tags key value`)
  b. by using an argument (`tag Arm.Web.serverFarms key value`) optionally using 1 or 2 as a default value?

tehmufifnman:
Additionally many resources aren't compatible with the tags field, so some way of representing that will be necessary.

isaacabraham:
@tehmufifnman Really? I thought it was some generic thing. In that case, perhaps doing this on an ad-hoc, as-needed basis is indeed the way to go.

tehmufifnman:
@isaacabraham Yup - I am sorta fell into the role of ARM template whisperer around my office LOL

Off the top of my head the one example that comes to mind is a Service Bus Queue/Topic Subscription.
https://docs.microsoft.com/en-us/azure/templates/microsoft.servicebus/namespaces/topics/subscriptions
(notice the lack of "tags")

If you try to add a tags field and deploy you get a validation error.

Here is the full list of tag support from Microsoft:
https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/tag-support

isaacabraham:
[...]

PR #300, opened 2020-08-06
Introduce Tag support for all currently supported resources

TheRSP:
This is an initial step in implementing tag support for resources (#298). At the moment, this only affects the `webApp` and `function` builders. If this implementation is acceptable, I will follow the same pattern for the other builders. and then update docs

**Usage:**

```fsharp
webApp{
  name "my-web-app"
  tag "key" "value"
  tags [ "alpha","a"; "beta","b" ]
}
```

### DIFF
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -75,6 +75,7 @@ type WebAppConfig =
       OperatingSystem : OS
       Settings : Map<string, Setting>
       Dependencies : ResourceName list
+      Tags : Map<string,string>
 
       Cors : Cors option
       Sku : Sku
@@ -119,6 +120,7 @@ type WebAppConfig =
               WebSocketsEnabled = this.WebSocketsEnabled
               Identity = this.Identity
               Cors = this.Cors
+              Tags = this.Tags
               AppSettings =
                 let literalSettings = [
                     if this.RunFromPackage then AppSettings.RunFromPackage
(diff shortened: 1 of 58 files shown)


isaacabraham:
Given the current state of the API, this seems a reasonable way to go. The only concern I have is that *every single* resource and builder will need to implement this by hand. Not a huge issue, but I'm wondering if there's a better way to compose this data within a builder / IArmResource.

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -440,6 +443,12 @@ type WebAppBuilder() =
     member this.EnableCi(state:WebAppConfig) = this.SourceControlCi(state, Enabled)
     [<CustomOperation "disable_source_control_ci">]
     member this.DisableCi(state:WebAppConfig) = this.SourceControlCi(state, Disabled)
+    [<CustomOperation "tags">]
```
isaacabraham, line 446: Rename to `add_tags` and `add_tag` please, unless each operation will overwrite the existing tags that have already been set.

TheRSP:
I believe that all resources that can, now have tag support.
Any resource not explicitly supporting tags on this list has been skipped: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/tag-support

tehmufifnman:
So exciting!

isaacabraham:
Just going through this - was it intentional to add a load of NSG stuff into this PR?

TheRSP:
No, all of that came in as part of a merge. I wasn't sure if I should abandon the PR and start a new one or leave it to keep the comments and continuity of the PR?
My changes are just the tags related stuff

isaacabraham:
@TheRSP I'm just not sure where that came from as it appears to show e.g. NSG docs as something new in this PR, which is confusing me! There are already some merge conflicts in this branch anyway.

TheRSP:
I will create a new PR to replace this one

PR #301, opened 2020-08-08
NSG simplification

isaacabraham:
@ninjarobot Here's some simplifications I've made.

1. I'm not sure about the name on the Service "group" or indeed individual services (within the context of security rules) - they don't appear to be used anywhere? What are they for?
2. I think the Source / Destination Port / Ports could be modelled and brought into the ARM Resource, but we should talk it through so I understand.
3. I've removed the mutable HashSets with the simple Set / List.distincts.

### DIFF
--- a/src/Farmer/Arm/NetworkSecurityGroup.fs
+++ b/src/Farmer/Arm/NetworkSecurityGroup.fs
@@ -34,8 +34,7 @@ type SecurityRule =
       DestinationAddresses : Endpoint list
       Access : Operation
       Direction : TrafficDirection
-      Priority : int
-    }
+      Priority : int }
     interface IArmResource with
         member this.ResourceName = this.Name
         member this.JsonModel =
--- a/src/Farmer/Arm/NetworkSecurityGroup.fs
+++ b/src/Farmer/Arm/NetworkSecurityGroup.fs
@@ -21,18 +21,32 @@ type NetworkSecurityGroup =
                location = this.Location.ArmValue
                tags = this.Tags |} :> _
 
+let (|SingleEndpoint|ManyEndpoints|) endpoints =
+    // Use a wildcard if there is one
+    if endpoints |> Seq.contains AnyEndpoint then SingleEndpoint AnyEndpoint
+    // Use the first tag that is set (only one supported), otherwise use addresses
+    else
+        endpoints
+        |> Seq.tryFind (function Tag _ -> true | _ -> false)
+        |> function
+        | Some (Tag tag) ->
+            SingleEndpoint (Tag tag)
+        | None | Some (AnyEndpoint | Network _ | Host _) ->
+            ManyEndpoints (List.ofSeq endpoints)
+
+let (|SinglePort|ManyPorts|) (ports:_ Set) =
+    if ports.Contains AnyPort
+    then SinglePort AnyPort
+    else ManyPorts (ports |> List.ofSeq)
+
 type SecurityRule =
     { Name : ResourceName
       Description : string option
       SecurityGroup : NetworkSecurityGroup
       Protocol : NetworkProtocol
-      SourcePort : Port option
-      SourcePorts : Port list
-      DestinationPort : Port option
-      DestinationPorts : Port list
-      SourceAddress : Endpoint option
+      SourcePorts : Port Set
+      DestinationPorts : Port Set
       SourceAddresses : Endpoint list
-      DestinationAddress : Endpoint option
       DestinationAddresses : Endpoint list
       Access : Operation
       Direction : TrafficDirection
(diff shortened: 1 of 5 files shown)


src/Farmer/Builders/Builders.NetworkSecurityGroup.fs
```diff
@@ -33,100 +33,95 @@ type SecurityRuleBuilder () =
...
-            // Use the first tag that is set (only one supported), otherwise use addresses
-            else
-                addresses
-                |> Seq.tryFind (function Tag _ -> true | _ -> false)
-                |> function
-                | Some (Tag tag) ->
-                    Some (Tag tag), []
-                | None | Some (AnyEndpoint | Network _ | Host _) ->
-                    None, List.ofSeq addresses
+    let destPorts = rule.Services.Services |> List.map(fun (Service(_, port)) -> port) |> Set
```
ninjarobot, line 100: Nice, good to replace the mutable HashSet with F# Set instead.

src/Farmer/Builders/Builders.NetworkSecurityGroup.fs
```diff
@@ -33,100 +33,95 @@ type SecurityRuleBuilder () =
     member _.Description(state:SecurityRuleConfig, description) = { state with Description = Some description }
     /// Sets the service or services protected by this rule.
     [<CustomOperation("service")>]
-    member _.Service(state:SecurityRuleConfig, service) = { state with Service = service }
```
ninjarobot, line 36: It would be good to keep the ability to set the Service on a rule, because if building both inbound and outbound rules, you will probably want to use the same service.  For example, you would not only restrict your database server to be only accessible from the application server network, you may also restrict the application server network to only have outbound access to database servers.  Doing this, you'll want to reference that service in both the inbound and outbound rules.

src/Farmer/Common.fs
```diff
@@ -764,32 +764,32 @@ module NetworkSecurity =
...
-        | Network cidr -> cidr |> IPAddressCidr.format
-        | Tag tag -> tag
-        | AnyEndpoint -> "*"
-
-    type Service =
-    | Service of Name:string * Port
-    | Services of Name:string * Service list
+        let ArmValue (endpoint:Endpoint) = endpoint.ArmValue
+
+    type NetworkService = NetworkService of name:string * Port
```
ninjarobot, line 792: An NSG rule needs to be about source and destination endpoints and what is allowed/denied rather than each individual port or port range. While it's true that either can work, this requires someone to mentally compile the sequence in which rules are processed, making sure allow and deny rules work as intended for the various ports and endpoints rather than being able to think of a rule set in context of all of the access needs of their sources and destinations.

Flattening this loses the composability of services such that they are only represented at the lowest level, as a raw set of ports. In the context of a single service per rule, that's ok, but there are often many services running in an application, many applications on a machine or cluster, many clusters in a subnet, many different subnets with different applications and consumers. Composing services lets you represent access for a single port/range or an application or a cluster or many clusters - it can scale up to encapsulate that complexity whereas a list keeps the complexity right in front of you.

src/Farmer/Arm/NetworkSecurityGroup.fs
```diff
@@ -46,17 +59,39 @@ type SecurityRule =
...
+                  | SingleEndpoint endpoint ->
+                    "destinationAddressPrefix", endpoint.ArmValue |> box
+                    "destinationAddressPrefixes", box []
+                  | ManyEndpoints endpoints ->
+                    "destinationAddressPrefixes", box [ for endpoint in endpoints -> endpoint.ArmValue ]
+
+                  "access", box this.Access.ArmValue
+                  "priority", box this.Priority
+                  "direction", box this.Direction.ArmValue
+                ] |> Map
```
ninjarobot, line 96: This seems like a step back from an anonymous record in terms of type safety, and I'm not sure what is gained. If we make some change to the dictionary, there's no compiler to help us find mistakes. It's like Python code - easier to write, but harder to refactor because you don't know what you've broken.

It also makes a new pattern that we probably don't want. I could see contributors making maps for building JSON instead of anonymous records and then having to explain this was a special case and convince them of when to use one over the other.

isaacabraham: No problem going back - it was just a bit more verbose. To be honest in terms of type safety we're not gaining anything here - it's just an anonymous record. I suppose the main benefit we got was that we assign everything to a single field at once. I'll revert to that approach.

src/Farmer/Common.fs
isaacabraham: Yep, I agree, but I'm starting to think - at least from a V1 approach - that we should leave the "composability" aspect as a concern for the calling to do i.e. you can easily model that yourself and then when it comes to injecting into Farmer, just flatten the data at that point.

Would that work? We can then consider re-introducing that datatype at a later date.

ninjarobot: It can work, yes, people can manage the lists of ports and group them on their own. Was the recursive definition too difficult to use?

src/Farmer/Arm/NetworkSecurityGroup.fs
ninjarobot: Yes, it is anonymous, but it enforces the presence and structure of those fields for the resulting JSON, which is a fragile part of ARM templates.

src/Farmer/Common.fs
isaacabraham: I started backporting it but the code was a bit more complex. I'm not convinced that we should be this opinionated on the type here - at least not yet. We can always add it as an overload and then flatten (or it could be done as an extension method on the class).

PR #302, opened 2020-08-08
Refactoring around filters

isaacabraham:
@ninjarobot I've made some changes to the Filter code, mostly around the API. Can you have a look at it?

### DIFF
--- a/src/Farmer/Common.fs
+++ b/src/Farmer/Common.fs
@@ -587,16 +587,16 @@ module ServiceBus =
         | Standard
         | Premium of MessagingUnits
     type Rule =
-        | SqlFilter of {| Name : ResourceName; SqlExpression : string |}
-        | CorrelationFilter of {| Name : ResourceName; CorrelationId : string option; Properties : Map<string, string> |}
+        | SqlFilter of ResourceName * SqlExpression : string
+        | CorrelationFilter of Name : ResourceName * CorrelationId : string option * Properties : Map<string, string>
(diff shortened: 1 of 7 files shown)


src/Tests/ServiceBus.fs
```diff
@@ -193,24 +190,29 @@ let tests = testList "Service Bus Tests" [
                             add_subscriptions [
                                 subscription {
                                     name "my-sub"
-                                    add_rules [
-                                        Namespaces.Topics.correlation_property_filter "SuccessfulStatus" ["Status", "Success"]
-                                        Namespaces.Topics.correlation_property_filter "FailedStatus" ["Status", "Fail"]
+                                    add_filters [
+                                        Rule.CreateCorrelationFilter("SuccessfulStatus", ["Status", "Success"])
+                                        Rule.CreateSqlFilter("Thing", "Status = Success")
```
ninjarobot, line 195: Yes, this API is less noisy to use, good improvement.

src/Farmer/Common.fs
```diff
@@ -587,6 +586,17 @@ module ServiceBus =
         | Basic
         | Standard
         | Premium of MessagingUnits
+    type Rule =
```
ninjarobot, line 589: I'm good with this in Common.fs - the only reason I put them in the builder was that I was a little hesitant at first to add more types here and was leaning towards helper functions to try to keep the API simple.

PR #307, opened 2020-08-13
Add support for NuGet Continous Deployment - Fixes #306

### LINKED ISSUE #306: Automate Nuget deployment
Something that is slowing down iterating on the project in terms of more rapid release is that the push to Nuget is not automated.

The build, test and package is, but it's still a manual process for me to log in to release to NuGet. What's the best way to - ideally as a manual button click - we could do for this?

ctolkien:
ADoS admins will need to add:

- [x] Service Connection called 'Nuget' which has the appriorpaite credentials stored
- [x] Pipeline environment, also called 'NuGet'. Put a manual approval gate on this. It will then require an authorised user to click 'Approve' to progress.

Caveats - with every yaml build process there has been a lot of wack-a-mole to get these to work. I'd expect the same.

Fixes #306

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,6 +37,7 @@ jobs:
       testResultsFormat: VSTest
   - task: DotNetCoreCLI@2
     displayName: 'Package'
+    condition: and(succeeded(), eq(variables.vmImage, 'ubuntu-latest'))
     inputs:
       command: 'pack'
       packagesToPack: 'src/Farmer'
@@ -52,7 +53,7 @@ jobs:
 - deployment: "NugetPublish"
   dependsOn: ["Build"]
   condition:  and(succeeded(), eq(variables['build.sourceBranch'], 'refs/heads/master')) #optional, only trigger for master
-  displayName: "Publish to Nuget"
+  displayName: "Publish to NuGet"
   environment: "NuGet" #needs to be configured in ADoS - put a manual gate here
   pool:
     vmImage: 'windows-latest' #task doesn't run on other OS's
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -46,6 +46,7 @@ jobs:
       verbosityPack: 'Minimal'
   - task: PublishBuildArtifacts@1
     displayName: Store NuGet Package
+    condition: and(succeeded(), eq(variables.vmImage, 'ubuntu-latest'))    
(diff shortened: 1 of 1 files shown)


ctolkien:
Based off of this:
https://github.com/ctolkien/TinyPNG/blob/master/azure-pipelines.yml

isaacabraham:
@ctolkien thank you very, very much. I will look into this.

ctolkien:
Should go without saying that the CI/CD bits for this won't pass until the appropriate Service Connections are added :D

ctolkien:
> I currently publish both the Linux and Windows-built version of the packages.

As in, ADoS publish build artifact runs for both? I'm, not actually sure what it does here. It may just overwrite with a "last one wins". We can disable that step running on one of the platforms with a `condition` check (for instance to only run on `ubuntu-latest`):

```
condition: and(succeeded(), eq(variables.vmImage, 'ubuntu-latest'))
```

>there's also the snupkg for the symbols

As the name/id is the same, NuGet will detect and push the snupkg automatically.

isaacabraham:
@ctolkien could you add that condition to the script? I've added the stuff into azdo so I think this is nearly ready to go. Thanks again.

ctolkien:
Sigh... line endings.

@isaacabraham  - have added the 'pack' conditional. There is still a conditional check in there that only commits to master can be considered for release - assume that is OK?

isaacabraham:
Perfect. I guess the only way to test this is to simply "bite the bullet"? :-)

isaacabraham:
@ctolkien I've merged this in and am running it now :-)

Question: what is the workflow supposed to be - what I'm seeing now is that the entire build does not start until I acknowledge the review. This means that we can't do a build and test in master without my pushing through the *whole* pipeline.

What I would like is:

* Any commit to master kick off a full build / test.
* If they pass, the build is green.
* I can *optionally* elect to push to NuGet. This will not happen on every commit, for sure.

Is that kind of workflow possible?

ctolkien:
Hi,
>Question: what is the workflow supposed to be - what I'm seeing now is that the entire build does not start until I acknowledge the review. This means that we can't do a build and test in master without my pushing through the whole pipeline.

Ah no, that's not right. The first step should always run, and it's only the NuGet push should require confirmation (as you've outlined in how you'd like it to work).

I'll take a look.

PR #313, opened 2020-08-17
Add Tag support

TheRSP:
This PR is to replace the abandoned PR #300. It introduces tag support for all resources which support tags
#298


isaacabraham:
Thank you!

Issue #314, opened 2020-08-17
Wrong family for PostgreSQL


VitorRigoni:
Hello! 👋🏻

It's me again 😅
I have a fun little problem this time: I setup an azure pipeline to deploy my infrastructure alongside my application.
To get this to work the way I want, I created an fsx script importing farmer. My pipeline restores the packages, runs the fsx (which in turn generates the ARM template) and deploys the arm template using the `AzureResourceManagerTemplateDeployment@3` task.

One of my resources is a postgresql db.

Everything works fine, the arm template is generated, but I get the following error when applying the generated template:
`##[error]InvalidEditionSloCombination: The edition Basic does not support the service objective B_PostgreSQLFamily.Gen5_2`

I understand what is going on (the single DU case is also printing the type name when it shouldn't), but I don't understand why this is happening. It's not clear to me why `ToString()` would have different implementations

I cannot reproduce on my local machine (Windows, dotnet 3.1.401), but I can reproduce on an ubuntu container with dotnet 3.1.401.

I can reproduce on both windows and ubuntu nodes on azure pipelines. To be honest, I'm not sure if I should be creating this bug here or against the dotnet repo itself or azure devops but regardless I thought I should let you know.

This is the output on the arm template generated on azure (wrong):
```json
"sku": {
    "capacity": 2,
    "family": "PostgreSQLFamily.Gen5",
    "name": "B_PostgreSQLFamily.Gen5_2",
    "size": "5120",
    "tier": "Basic"
},
```

This is the output from my local machine (correct):
```json
"sku": {
        "capacity": 2,
        "family": "Gen5",
        "name": "B_Gen5_2",
        "size": "5120",
        "tier": "Basic"
      },
```

Any input is very much appreciated =)
Thanks!

isaacabraham:
I *think* that this is due to a [breaking change](https://github.com/dotnet/fsharp/issues/9357) that was introduced in the latest F# - previously, this would return just the value, but now is the fully-qualified item. We need to fix this.

cc: @cartermp - what's the plan for that issue - is it going to be reverted?

VitorRigoni:
Oh wow, that's crazy!

I can submit a PR for this, if you'd like. Do you have an idea how you wanna do this?

cartermp:
No, there are no plans to change this.

PR #316, opened 2020-08-18
Add tostring override to postgre family name DU

VitorRigoni:
This PR fixes the issue introduced by the change in ToString() behavior on the latest fsharp version.

Issue: https://github.com/CompositionalIT/farmer/issues/314


isaacabraham:
Thank you very much!

Issue #319, opened 2020-08-23
custom_script_files option doesn't seem to work


pviotti:
Hi there - I was trying to upload a script file to my vm using the `custom_script_files` option as follows:

```fsharp
    let myVm = vm {
        name vmResourceName
        username vmUsername
        vm_size Vm.Standard_B2s
        operating_system (Vm.makeLinuxVm "0001-com-ubuntu-server-focal" "canonical" "20_04-lts")
        os_disk 64 Vm.Premium_LRS
        diagnostics_support
        domain_name_prefix (Some vmHostname) 
        custom_script_files [ "https://gist.githubusercontent.com/blablabla/setup-vm.sh" ]
    }
```

but the script file URI doesn't make it to the ARM template file.  
Does it work only with Windows VMs..? Am I missing something?

(Thanks for Farmer btw, it's a pretty handy project!)

isaacabraham:
@pviotti You have to set the *name* of the custom script:

```fsharp
    let myVm = vm {
        name vmResourceName
        username vmUsername
        vm_size Vm.Standard_B2s
        operating_system (Vm.makeLinuxVm "0001-com-ubuntu-server-focal" "canonical" "20_04-lts")
        os_disk 64 Vm.Premium_LRS
        diagnostics_support
        domain_name_prefix (Some vmHostname) 
        custom_script "setup" // add this line
        custom_script_files [ "https://gist.githubusercontent.com/blablabla/setup-vm.sh" ]
    }
```

Basically, this becomes the name of the script resource in the ARM template. Without it, we ignore the script files.

The fact that you've fallen for this suggests that the API isn't obvious to figure out. We could do one of several things:

1. Throw an error if you provide files without a name (or vice versa). This would have probably been enough for you on this occasion.
2. Change the API so that you have to supply the name *and* the files together:
```fsharp
        custom_script_files "setup" [ "https://gist.githubusercontent.com/blablabla/setup-vm.sh" ]
```
3. We could make this optional and look at setting an automatically generated name for this instead. However, I'm not sure if the resource name has restrictions or must be unique etc.

isaacabraham:
P.S. It should work for both Linux and Windows.

isaacabraham:
@pviotti ignore what I said before. It was nonsense and I've deleted the comment. I've just had another look through the source code and now remember how this works in Azure.

* The custom script files are simply what files to copy up onto the VM.
* You still have to supply a script that is *executed* on the VM, which may use those files.

[...]

PR #322, opened 2020-08-25
[docs] add custom script to the vm docs example

pviotti:
As per #319, here's a small addition to the docs.  
(tbh I have not tried if it works with `powershell scriptname.ps1` on Windows, as I used the feature only on Linux vms, but I didn't want to add a Linux example just for that)  
Perhaps for this kind of non-trivial features it's worth linking to the upstream docs as well (?)

Btw, README.md says to check `docs` branch for the docs, but I found that it's not the most up-to-date version.


isaacabraham:
Thank you!

isaacabraham:
And good spot - I've updated the readme. I should probably delete the docs branch, it's not used any more.

PR #325, opened 2020-09-01
ACI dropped support for assigning static private IP

ninjarobot:
The Azure container instances team changed the service behavior so if you pass a static IP for a private container group, it now returns an error, so deprecation of this capability follows that change.


isaacabraham:
Am I right that this is literally a breaking change now on the resource plane level - it's not about the new version of ARM API has changed but a breaking change to the current version?

isaacabraham:
Ouch. But it's going out sometime soon regardless?

isaacabraham:
Thank you for taking care of this.

PR #339, opened 2020-09-08
Typed resource names

isaacabraham:
This experimental branch uses [phantom types](https://medium.com/@dogwith1eye/a-phantom-costume-for-our-types-694aa385f37b#:~:text=A%20phantom%20type%20is%20simply,type%20without%20an%20apparent%20purpose.) to see how to cheaply add type safety between different resource names which may have different validation rules.

It does not *guarantee* type safety - there is no notion of a "private constructor" to enforce valid creation of a resource. However, it does make it much harder to accidentally assign e.g. a `ResourceName<StorageAccountName>` to a `ResourceName<WebAppName>`.

What do you think of this?

Of interest is the validation code in `Common.fs` which removes the need with custom types that convert to / from resource names to simple generic "interface markers".

The idea would be to create validation functions for each specific resource name that conforms to Azure naming rules e.g. see [here](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules). There is also the Azure HTTP API that we could look at to see if it's possible to reverse engineer some rules from (the github repo eludes me).


isaacabraham:
cc: @hravnx @ninjarobot @Dzoukr - would be interested in your thoughts on this.

hravnx:
I like the idea of having a centralized, discoverable place for validation. Especially if coupled with clear guidance/examples on best practices for resource type implementation.

That last bit was the most challenging aspect for me, when doing the PostgreSQL bits. figuring out what is the 'right' way to implement stuff so that it fits in with the rest.

Dzoukr:
>I like the idea of having a centralized, discoverable place for validation

Same here! Looks great!

Issue #351, opened 2020-09-10
depends_on should support lists


isaacabraham:
This will better support parameterisation.

PR #358, opened 2020-09-12
#351 Depends_on as a list

MNie:
Fixes: #351 

I'm not sure that marking `IBuilder` is also a good idea since we would need then to always cast to an `IBuilder` so instead of writing:

```fsharp
depends_on [ builder ]
```

we would need to write:

```fsharp
depends_on [ builder :> IBuilder ]
```

@isaacabraham any thoughts/hints?

### DIFF
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -47,7 +47,7 @@ let myStorageAccount = storageAccount {
 let myWebApp = webApp {
     name "myTestWebApp"
     setting "storageKey" myStorageAccount.Key
-    depends_on [ myStorageAccount ]
+    depends_on myStorageAccount
 }
 
 // Create an ARM template
--- a/docs/content/_index.md
+++ b/docs/content/_index.md
@@ -47,7 +47,7 @@ let myStorageAccount = storageAccount {
 let myWebApp = webApp {
     name "myTestWebApp"
     setting "storageKey" myStorageAccount.Key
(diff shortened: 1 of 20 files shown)


isaacabraham:
@MNie Yes, we see the same issue with `add_resources` - because F# doesn't implicitly upcast to common interfaces (@cartermp am I right that there's plans to change this in the future?) you have to explicitly do this.

I've thought about moving to an SRTP approach and looking for the appropriate fields on the records as the other way around this, or having all builders return an IBuilder rather than config records - but this strikes me as somewhat more opaque so I would rather avoid doing that if possible.

isaacabraham:
I was assuming that the list option would be added as an overload though, rather than replacing the original single item method.

cartermp:
> am I right that there's plans to change this in the future?

Maybe. But unlikely to happen any time soon.

MNie:
@isaacabraham I go the easy way then. I created overload methods that accepts lists:

- `DependsOn (state: 'config, resources: ResourceName list)`,
- `DependsOn (state: 'config, builders: IBuilder list)`,
- `DependsOn (state: 'config, resources: IArmResource list)`

I added some unit tests to see how it could be used + update documentation. Is there something else left to do here?

isaacabraham:
That's great @MNie thank you. I'll test this out and then merge.

PR #364, opened 2020-09-16
Move to System.Text.Json v5

isaacabraham:
Probably not ready until v5 proper drops but this branch *appears* to work!

I had to include both nuget packages otherwise I got runtime warnings.


isaacabraham:
@ninjarobot yep. Like the NET5 one, I'm going to just create a brand new PR to do this with. I've got a stashed version here that *seems* to work (i.e. all tests run except for that regression with floats -> JSON), but I need to test with e.g. az cli and parsing JSON as well.

Issue #382, opened 2020-09-25
Automatically create dependencies


isaacabraham:
For some keywords e.g. `settings` on WebApp, we should be able to implicitly add the dependency having to set it separately.

PR #385, opened 2020-09-26
Resource ids

isaacabraham:
This PR is related to issue #382. It follows on by creating a dedicated ResourceId type which allows much clearer definition of the resource path. Combined with the work done in #383, this allows for much better specification of resources. It also removes the "hack" of passing resource paths like this as Resource Names for certain dependencies.

This has been applied across all main "DependsOn". We will eventually reapply this to all dependencies within the codebase.

This will also be useful in the future for specifying "unmanaged" resources.

* Creation of a dedicated `ResourceId` type which replaces the `ArmExpression.resourceId` methods.
* Replace all "Depends On" with ResourceIds.
* Slightly updated DependsOn pattern for builders.
* Removed VM DependsOn (wasn't being used anywhere)


isaacabraham:
Have a look at this @ninjarobot - it could (will?) eventually replace ResourceName with a proper ResourceId object. This in effect replaces the `ArmExpression.resourceId` function with something that understands the different components of the resource (type, name, group, elements) as well as an optional "owner" - this is used to derive dependencies when utilising resources for e.g. settings.

It can correctly emit a `resourceId` ARM expression given all valid combinations of the different fields (I believe!).

isaacabraham:
Exactly - currently you had to hack it in by creating an resource ID expression and then converting it to a resource name - totally the wrong way to do it.

PR #388, opened 2020-09-28
Remodel DNS Zone types

isaacabraham:
@raymens when you get ten minutes, could you review this? I've slightly reworked the way that data is modelled for DNS Config:

* Uses the DU cases to store the metadata against each one
* No longer requires option for the type of record

This has the following impacts:
* Can use a single record to model all record types (great here because most fields are the same across all types)
* Less code on builder
* Ability to share validation logic
* Less code in ARM

What do you think?


raymens:
Nice, looks good! Thanks for guiding me along.

PR #394, opened 2020-10-04
Adds managed identity builder and support for ACI and AKS identities.

### LINKED ISSUE #245: Add Identity support for Storage Accounts.

### LINKED ISSUE #330: User-assigned managed identities?
I don't see this supported yet, but I'm wondering what it would take to support user-assigned managed identities on a web app. Right now there's a single enable_managed_identity option. Is it limiting to have that be a simple "enable" toggle? Or perhaps it could be expanded in the future to include a parameter of a ManagedIdentity list of some sort.

Now that I have time and a project to use with Farmer, I'd love to contribute some new resources.

### LINKED ISSUE #380: Add support for managed identity
This is more a place to discuss the right way to add a support for managed identity and how to configure and connect services with it.

Currently there is a way to enable managed identity on web app so it needs to be added to the rest of the services supporting it

We need to come up with a design how to assign identity roles to the services.

ninjarobot:
This PR closes #380, closes #330 and closes #245

* Adds a ManagedIdentity type and builder for userAssignedIdentities - this can be added to other resource types that support it.
* Adds support to container groups for specifying a system or user assigned managed identity.


isaacabraham:
@ninjarobot Apologies, I've not had a chance to review this yet. Hopefully will have some time today.

src/Farmer/Arm/ContainerInstance.fs
```diff
@@ -48,12 +50,30 @@ type ContainerGroup =
...
         member this.JsonModel =
             {| containerGroups.Create(this.Name, this.Location, this.Dependencies, this.Tags) with
+                   identity =
+                       match this.Identity with
+                       | None -> {| ``type`` = "None"; userAssignedIdentities = null |}
+                       | Some (SystemAssigned) -> {| ``type`` = "SystemAssigned"; userAssignedIdentities = null |}
+                       | Some (UserAssigned identities) ->
+                         // Identities are assigned as a dictionary with the user identity resource ID as the key
+                         // and an empty object as the value.
+                         let userAssigned = JObject()
```
ninjarobot, line 71: @isaacabraham I'd rather if I didn't have to use `JObject` directly, just for some reason, they are using data as keys (the managed identity resource ID needs to be the field name).

isaacabraham: Yeah, I think in this case I prefer a Map of string * obj, I think it's much clearer. Let's get this PR in and then we can review some alternatives for this.

src/Farmer/Arm/ContainerInstance.fs
```diff
@@ -48,12 +50,27 @@ type ContainerGroup =
...
         member this.JsonModel =
             {| containerGroups.Create(this.Name, this.Location, this.Dependencies, this.Tags) with
+                   identity =
+                       match this.Identity with
+                       | None -> {| ``type`` = "None"; userAssignedIdentities = null |}
+                       | Some (SystemAssigned) -> {| ``type`` = "SystemAssigned"; userAssignedIdentities = null |}
+                       | Some (UserAssigned identities) ->
+                         // Identities are assigned as a dictionary with the user identity resource ID as the key
+                         // and an empty object as the value.
+                         let userAssigned = identities |> List.map (fun identity -> (UserAssignedIdentity.resourceId identity).Eval(), obj) |> dict
```
ninjarobot, line 71: @isaacabraham it turned out the fix was easy enough. Using a `dict` works for this.

src/Farmer/Arm/ContainerInstance.fs
ninjarobot: Is piping it to `dict` OK?

isaacabraham:
Apologies for the delay - been a bit busy the last week.

isaacabraham:
I think we need some docs around managed identity in Farmer in general - probably good to do here.

src/Farmer/Common.fs
```diff
@@ -598,9 +598,15 @@ module Sql =
...
 
+module ManagedIdentity =
+    /// A user assigned managed identity that can be associated with a resource.
+    type UserAssignedIdentity = UserAssignedIdentity of Name:string * ResourceGroup:string option
+
 module ContainerGroup =
     type PortAccess = PublicPort | InternalPort
     type RestartPolicy = NeverRestart | AlwaysRestart | RestartOnFailure
+    /// Identity settings for a container group.
+    type ContainerGroupIdentity = SystemAssigned | UserAssigned of ManagedIdentity.UserAssignedIdentity list
```
isaacabraham, line 610: I wonder whether this is going to be a common pattern? I assume so, in which case we may wish to consider promoting this type into ManagedIdentity in general.

src/Tests/ContainerGroup.fs
```diff
@@ -138,5 +139,51 @@ let tests = testList "Container Group" [
...
+        }
+        let group =
+            containerGroup {
+                name "myapp-with-msi"
+                add_instances [ nginx ]
+                user_assigned_identity msi.Name
+            }
+        let template = arm {
+            location Location.EastUS
+            add_resource msi
```
isaacabraham, line 175: Again, from a consumption point of view - will people expect that they need to explicitly add this resource to the template? Or should it simply be added when injected into the `containerGroup`?

isaacabraham:
In principle I'm really happy with this. I just wonder whether it's worth taking a step back and looking at identity in general (system & user) and how to apply it consistently and easily across all resources.

We could do that review either:

* Now - before we merge this PR.
* After merging but before release.
* After beta.
* After full release.

My preference would be to try to figure this out now, but if we can't, then to do at at the latest after the beta release.

isaacabraham:
I've included my review comments. I'd like to see how this works in practice (from an API point of view). I've never used the managed user identities before:

1. Do we need to create the identity in e.g. the Portal (if it's a service principal) or does this do that for us?
2. If it's an identity of a real person e.g. you or me - does the name of the identity simply need to match, or is there something else required?

isaacabraham:
As a proof of concept, I'm trying to add this to Storage as well. A few more questions I'm asking myself:

1. Do we actually *need* the Identity builder / config? Could we just make it a simple value and a function that creates them?
2. What's the relationship between the Identity Builder Config / Name and the `UserAssignedIdentity` in `Common.fs`?
3. What type is the identity (just a string currently) - what should it represent? My understanding was it should be the object id (guid) from Azure.

isaacabraham:
We should also review in context of the (crude) work I did to get the app service working with keyvault. What we do here could / should replace that, I would think.

src/Farmer/Common.fs
ninjarobot: Yes, this is likely the case on the majority of resources that support a managed identity, so I agree, moving it to the general `ManagedIdentity` would simplify the effort to add it to different resources.

isaacabraham:
We should probably also think about Roles at the same time.

src/Farmer/Builders/Builders.ContainerGroups.fs
```diff
@@ -139,6 +144,18 @@ type ContainerGroupBuilder() =
...
+    member _.Identity(state:ContainerGroupConfig, identity:ContainerGroupIdentity) =
+        { state with Identity = Some identity }
+    /// Enables a system assigned managed identity to be set for this container group.
+    [<CustomOperation "system_assigned_identity">]
+    member _.SystemAssignedIdentity(state:ContainerGroupConfig) =
+        { state with Identity = Some SystemAssigned }
+    /// Sets the user assigned managed identity on this container group to a user assigned identity in the same resource group.
+    [<CustomOperation "user_assigned_identity">]
+    member _.UserAssignedIdentity(state:ContainerGroupConfig, userIdentity:ResourceName) =
+        { state with Identity = Some (UserAssigned [ ManagedIdentity.UserAssignedIdentity(userIdentity.Value, None) ]) }
```
isaacabraham, line 158: Why do we store the name of the user assigned identity as a string, only to wrap it back into a Resource Name later, when it's actually coming in here as a resource name?

src/Tests/ContainerGroup.fs
ninjarobot: Usually when using a `UserAssigned` identity, it would be applicable to multiple resources. For example, say you have a key vault that contains your connection strings for all your web servers. You could create a managed identity `web-servers` and grant that identity access to read the key vault secrets to get the connection string.  Then whenever you add a new web server, it is assigned that `web-servers` managed identity so that it will have rights.

On the other hand, generating one fits in more with a `SystemAssigned` managed identity where it's generated by Azure for that resource. Then each time you add a web server VM, you might explicitly change the access policy on the key vault to add that web server's `SystemAssigned` identity.

ninjarobot: Worth a mention - there is no ARM resource to create for a `SystemAssigned` identity - it's created implicitly so it isn't defined as a resource in the template, more just "enabled" on the resource and Azure will generate an identity for the resource.

src/Farmer/Builders/Builders.ContainerGroups.fs
ninjarobot: This is to lighten the syntax for the simplest case where there is a user identity resource in the same resource group. More complex scenarios where the user identity is in a different resource group or subscription would need to use the `identity` operation instead.

src/Farmer/Common.fs
```diff
@@ -602,12 +602,14 @@ module ManagedIdentity =
     /// A user assigned managed identity that can be associated with a resource.
     type UserAssignedIdentity =
         | UserAssignedIdentity of Name:string * ResourceGroup:string option
```
ninjarobot, line 604: Make this the `ResourceId`

docs/content/api-overview/resources/container-group.md
```diff
@@ -51,10 +54,15 @@ let nginx = containerInstance {
...
 
+let containerGroupUser = userAssignedIdentity {
+    name "aciUser"
+}
+
 let group = containerGroup {
     name "webApp"
     operating_system Linux
     restart_policy AlwaysRestart
+    user_assigned_identity containerGroupUser.Name
```
ninjarobot, line 65: `identity containerGroupUser.Identity` could get the `ResourceId` constructed from the `ResourceName`.

isaacabraham:
I've started the work as discussed by somewhat unifying how we go from the identity config to a builder (e.g. container group) via the resource identity DU.

src/Farmer/Common.fs
isaacabraham: This is now done.

src/Farmer/Arm/ManagedIdentity.fs
```diff
@@ -0,0 +1,42 @@
...
+
+//TODO: Remove optionality?
+/// List of resource ID's for the managed identities when a resource is using user assigned identities.
+let Dependencies = function
+    | Some (UserAssigned identities) ->
+        identities
+        |> List.map (fun identity -> identity.ResourceId)
+    | None
+    | Some SystemAssigned ->
+        []
```
isaacabraham, line 17: @ninjarobot should we not add the resource that the system assigned identity belongs to as a dependency? If so - this implies that the System Assigned half should *also* create / store the ResourceId of that resource.

isaacabraham:
For this code:

```fsharp
.create(sprintf "reference(resourceId('Microsoft.Web/sites', '%s'), '2019-08-01', 'full').identity.principalId" this.Name.Value) 
```

I understand that the resource id should equal the owning resource e.g. in this case a website. What should this look like for a user assigned identity - the resource id of the identity itself (totally fine) but should we also include the bits at the end e.g. `identity.principalId` etc.?

src/Farmer/Common.fs
```diff
@@ -598,9 +598,20 @@ module Sql =
...
                 c
 
+module ManagedIdentity =
+    /// A user assigned managed identity that can be associated with a resource.
+    type UserAssignedIdentity =
+        | UserAssignedIdentity of CoreTypes.ResourceId
+        member this.ResourceId = match this with UserAssignedIdentity rId -> rId
+    type ResourceIdentity =
+        | SystemAssigned
+        | UserAssigned of UserAssignedIdentity list
```
isaacabraham, line 608: I was thinking of putting the `PrincipalId` member at this level but that doesn't work since User Assigned could be a list. Hmmm.

isaacabraham:
@ninjarobot I'm clocking off now. I've made some good progress on this:

* Made the type changes as discussed
* Moved over to ResourceId (this also got rid of a compiler warning).

I've also got some experimental code that has ported over the App Service to use the new abstractions that we've made and it seems to work!

However, I'm stuck on two things:

1. Creating a PrincipalId for a ResourceIdentity. You can't do this because a ResourceIdentity could contain 0 1 or many User Assigned Identities. What PrincipalId should we return? Put another way: If we provide a Resource Identity to KeyVault as an access policy, which one should be used? All of them?

2. SystemIdentity - it seems to me like we are going to have to capture in this the ResourceId of the owning resource. However, if we do that, we can't "set" the SystemIdentity from the caller. I'm starting to think we will need two different "types" somehow - a SystemIdentity | UserAssignedIdentity that is used by the *caller*. System Identity in this case doesn't contain the resource id. Then, once you have "run" the builder and created the config, this moves over to a type where the SystemIdentity *does* contain the ResourceId.

isaacabraham:
Right. Apologies @ninjarobot for this but I need to write this out to make sure we're grokking the different cases and flows. As I see it, we have a few elements to think about:

# General Background on Azure Identities

## Uses of identities
* Resources can *run as* an identity e.g. a compute resource such as containers, app service, functions, VMs etc. There can be more than one identity in use at a time - an azure resource can "impersonate" multiple identities simultaneously.

    For the purposes of this post, I will call these *impersonating* resources.

* Resources can *grant* an identity access *to them*. This might include Azure Storage or SQL Azure. Typically there are some *roles* or permissions which must also be granted to that identity. There are some generic roles e.g. Contributor, Owner etc. but e.g. Azure Storage has roles unique to that resource as well. A resource can grant access to multiple identities simultaneously, each with their own set of permissions.

    In this document, I will call these resources *granting* resources.

## Identity Types
* Identities can be automatically generated by an impersonating resource itself (a *system identity*). These identities are bound to that resource and cannot be shared across resources.
* Identities can also be generated by the user as a *user assigned identity*. UAIs can be reused across multiple resources and their lifetime is not bound to the impersonating resources that use them.

> Question: Can an impersonating resource run as *both* its own system identity *and* user assigned identities simultaneously?

# Farmer and Identity

There are several abstractions at play here.

## Creating Identities
### User Assigned Identities
We already have the UserAssignedIdentity builder and an associated helper builder.

```fsharp
let userIdentity = userAssignedIdentity { name "user" }
let userIdentity = ResourceId.create("user") |> ManagedIdentity.create
```

> Note that the builder variant doesn't allow support for resource groups etc. - perhaps we should add a `from_resource_id` keyword?

Either way, these are "full" Farmer resources and would be added to the `arm { }` block to be formally created.

## Setting Identities on Impersonating Resources

### Builders
This includes Container Groups, Web Apps etc.. By default these have no identity turned on. The consumer needs to be able to either:
* Activate the system identity on the resource.
* Provide one or more User Assigned Identities (see below for creating them) for the resource to run under. Each of these would be added to the set of impersonated IDs.

Here's how we could allow this consistently across resources (as a pattern):

```fsharp
let myWebApp = webApp { // could be vm, container, webapp etc.
    use_system_identity // activate the system identity
    add_user_identity userIdentity // use an already-created user assigned identity
}

```

### IARMResource
When writing out the templates, we need to know the identities that are required which were set in the builder. You can then write to the template with those details.

#### For System Identities
**You don't need to know anything except for the fact that a System Identity was requested**. You don't need a resource id or any other info.
#### For User Assigned Identities
You need to know the resource information so that:
1. You can correctly emit the path to that resource
2. To set the dependency on the impersonating resource.

## Providing Identities to Granting Resources
A granting resource e.g. Azure Storage or Azure KeyVault needs to be providing with the identities to grant access to. These granting resources generally do not care which type of identity was provided (system or user assigned). In both cases:

* The identity must provide a PrincipalId via an ARM expression. **This arm expression needs to know the resource id of the identity.** Both paths for system and user assigned ids.
* The granting resource is implicitly dependant on the identity.

### Builders
We can provide keywords to grant access and a role (or set of roles?) e.g. for storage:

```fsharp
let myStorage = storageAccount {
    name "mystorage"
    grant_access userIdentity Roles.Contributor // user assigned identity
    grant_access myWebApp.SystemIdentity Roles.Contributor // system identity of the web app
}
```

> The above system identity example implies that the user *must* have called `use_system_identity` on the webapp first.

### IArmResource
I've only looked at the Storage example but I *think* that this might be generic across resources. You ultimately need to provide the name of the granting resource, the role and the principal id of the identity. There's some nonsense with generating deterministic GUIDs but I think I've got that working.

## Current Challenges

### 1. System Identities
There is one problem from a type system point of view. Currently in my experimental branch, I model Managed Identity as follows:

```fsharp
type ManagedIdentity =
    | SystemIdentity of ResourceId option
    | UserAssignedIdentity of ResourceId
````

> I've (temporarily?) removed the list ability of UAIs. We will definitely revisit this.

The concern I have is around the option on System Identity. It's required because of how this type is used differently on Exposing and Granting resources:

| Identity Type | Exposing Resource | Granting Resource |
|-|-|-|
| User Assigned | Needs full resource id | Needs full resource id  |
| System Assigned | Don't have or need resource id | Needs full resource id |

This suggests to me that the ManagedIdentity abstraction is actually wrong. We need two abstractions: one for Impersonating Resources and one for Granting Resources. The Granting one is perhaps just the list of principal ids, but something needs to be cleaned up here.

2. ### Multiple Identities
Since both impersonating and granting resources can have multiple identities, we need to identify a clean way to model this.

**Comment by isaacabraham** [comment]
I've spent a couple of hours working through this and think it's at a point when I can merge into this branch for you to review.

**Comment by isaacabraham** [comment]
@ninjarobot I think this is *very nearly ready*. It now supports:

* All combinations of User Assigned identities (multiple supported again) and System Identity (simultaneously)
* Can retrieve Principal ID (object id) and Client Id as expressions (you'll need this if in code you need to auth with a user assigned identity).
* Storage support for granting access to a role
* All roles are now available in the Roles module

Still to do:

* [ ] Tests for Storage
* [ ] Tests for WebApp
* [ ] Maybe rename some of the types around identity
* [ ] Currently you can't retrieve the ClientId Arm Expression for a System Identity. This is only required if you have a resource running with *both* system *and* user assigned identities.
* [ ] Your review :-)

**Comment by isaacabraham** [comment]
@ninjarobot This is now done as far as I'm concerned. Feel free to test / review it; I'm happy to push this out later this week.

File: docs/content/api-overview/resources/managed-identity.md [hunk 10]
```diff
@@ -0,0 +1,161 @@
...
+            }
+        ]
+        // Also apply it here. All of the containers in this group share this managed identity.
+        add_identity sharedIdentity
+    }
+
+let data = storageAccount {
+    name "dataidentity"
+    // Allow the shared identity blob data reader access to storage.
+    grant_access sharedIdentity Roles.StorageBlobDataReader
```
ninjarobot, line 147: Wow - this is really nice ❤️

ninjarobot (commented):
This looks great, thank you for all the improvements here, looks great to me! I have a couple questions:

* What's the reason for `ManagedIdentity.Empty` rather than it being a `ManagedIdentity option` instead?
* It's great to have all those role ID's - was that code generated off of something?

isaacabraham:
> What's the reason for ManagedIdentity.Empty rather than it being a ManagedIdentity option instead?

Good question. I suppose we could put it as an option but when I was doing this it felt somehow more natural doing it as more of a "null object":

* System Assigned is just on or off
* User Assigned is a list which could be empty

Even with an option, we would still have to do a secondary check e.g. Some identity with a non-empty list of assigned identities or Some identity with System Assigned = Enabled etc.. I wasn't sure what the optionality brought us in terms of better reasoning about things e.g. what's the difference between None and Some (System = Disabled; User = []) ?

> It's great to have all those role ID's - was that code generated off of something?

It took about 15 minutes. Az CLI to generate the list, then some VS Code magic with multi code selection to do some boilerplate changes repeatedly. Sadly there was no way to identify logical "groups" of roles from the output.

isaacabraham:
And it's in!

isaacabraham:
Once I push a new release - probably today :-)

Issue #362, opened 2020-09-15
Add imageRegistryCredentials to containerGroup builder


tdwhite0:
I'd like to be able to use a custom image registry when deploying a Container Group:
https://docs.microsoft.com/en-us/azure/templates/microsoft.containerinstance/containergroups

    "imageRegistryCredentials": [
      {
        "server": "string",
        "username": "string",
        "password": "string"
      }
    ],

ninjarobot:
#395 adds support for this in the builder, also generating a secure parameter in the template so the container registry password is provided securely.

```fsharp
containerGroup {
    add_registry_credentials [
        registry "my-registry.azurecr.io" "user"
    ]
}
```

PR #395, opened 2020-10-05
Adds image registry credential support for container groups.

ninjarobot:
This PR closes issue #362 

* Adds support for specifying registry credentials for container image registries, including a secure password.
* Enables partial CPU core support for containers in a group.


isaacabraham:
Thank you. I started to do this but got stuck!

I did notice a few things about the container registry code that I would like to change though - let me review it properly tomorrow.

isaacabraham:
It might have been actually code outside of this PR - the use of `JObject` and a few other bits and pieces caught my eye. I'll re-review the PR on the basis of what's changed; the rest can be done as a separate PR.

src/Farmer/Builders/Builders.ContainerGroups.fs
```diff
@@ -33,7 +34,7 @@ type ContainerInstanceConfig =
       /// List of ports the container instance listens on
       Ports : Map<uint16, PortAccess>
       /// Max number of CPU cores the container instance may use
-      Cpu : int
+      Cpu : float
```
isaacabraham, line 37: A stupid question perhaps but can you really have half a CPU?

src/Farmer/Arm/ContainerInstance.fs
```diff
@@ -82,6 +90,11 @@ type ContainerGroup =
                             | AlwaysRestart -> "Always"
                             | NeverRestart -> "Never"
                             | RestartOnFailure -> "OnFailure"
+                          imageRegistryCredentials =
+                              this.ImageRegistryCredentials |> List.map (fun cred ->
```
isaacabraham, line 94: Formatting: move the pipeline onto a new line please.

isaacabraham (changes requested):
Happy to merge in aside from the one formatting change and a small question :-)

src/Farmer/Builders/Builders.ContainerGroups.fs
ninjarobot: It's really about resource scheduling on the host for the containers. The total number of cores used to host the container group depends on the requested CPU's, but then actually handing out the CPU to each container process is done by prioritization of each process rather than actually giving a container a whole core. Something asking for 1.0 cores gets twice the priority of something asking for 0.5 cores.

The way it is now with `Cpu : int` then most regions have a 4 core max so you could do at most 4 containers.  Really you could have many more smaller containers on a 4 core machine if they use 0.5 or 0.2 core reservations.

isaacabraham:
Great, thank you!

PR #409, opened 2020-10-20
#404 do not create ip address field if not needed

### LINKED ISSUE #404: It should be possible to create a container group without a public IP address
**Describe the bug**
It should be possible to create a container group without a public IP address.

**To Reproduce**
Do not set any ip address in the containerGroup builder:

```fsharp
let group = containerGroup {
    name "myGroup"
    operating_system Linux
    restart_policy RestartOnFailure
    add_instances [ myContainer ]
}
```

**Actual behavior**

The generated json-template still contains a property for ip-address, with an empty ports list:

```json
        "ipAddress": {
          "ports": [],
          "type": "Public"
        },
```

Trying to deploy this template causes the following ValidationError:

```ini
ValidationError: Deployment failed. Correlation ID: 2f88c005-49b1-4a0e-9731-8d67cdb0244a. {
  "error": {
    "code": "MissingIpAddressPorts",
    "message": "The ports in the 'ipAddress' of container group 'wordcount' cannot be empty."
  }
}
```

**Expected behavior**

If I do not set the IP address in the containerGroup-builder I expect the json-template to not contain the ipAddress property at all. (a template with "ipAddress": null can also be used, but will generate a warning in visual studio code).

MNie:
This PR closes #404

* mark IpAddress in ContainerInstance as option
* when ipAddress is not specified mark it as None and not generate an IpAddress section in ARM template

- Documentation don't need to be updated,
- I don't test my code on Azure but check if the generated example looks the same if we specify ports and if it looks okay if we don't specify them,
- I would update Release_notes.md when the code changes would be approved.

CC: @isaacabraham

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,9 +1,7 @@
 Release Notes
 =============
-## 1.2.0
-* Container Groups: Support for creating group without public IP Address.
-
 ## 1.2.0-beta2
+* Container Groups: Support for creating group without public IP Address.
 * Container Groups: Support for image registry credentials for private registries.
 * Container Groups: Support for partial CPU cores.
 * Log Analytics: Initial support.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,8 @@
 Release Notes
 =============
+## 1.2.0
+* Container Groups: Support for creating group without public IP Address.
+
 ## 1.2.0-beta2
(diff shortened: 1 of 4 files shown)


isaacabraham:
@MNie Great stuff. If you can update the release notes file, then this will naturally fall through when we do the next release. Thanks.

MNie:
@isaacabraham release notes updated : )

PR #414, opened 2020-10-22
#411 add branding

### LINKED ISSUE #411: Brand the console output
When running Farmer, we should emit some basic information at the start e.g.

```bash
==================================================
Farmer 1.2 beta 2
Repeatable Deployments in Azure made Easy!
==================================================
```

MNie:
This PR closes #411 

* add branding

- It couldn't be tested on Azure(?). I test it locally:
image

I'm curious about where to add branding. I add it to `quickWrite` but what with the `toJson`/`toFile` they are visible from the module so we should also add them there?
Also, if this is the right way to do the coloring inside of `Console.Write`?
- Documentation update not needed,
- Unit tests not needed,
- I would update Release Notes when code would be accepted.

CC: @isaacabraham


isaacabraham:
Please just put the release notes in the PR. Then I can accept the PR and the notes will naturally come in.

isaacabraham:
> I'm curious about where to add branding

I would say that any time you "run" the application it should show. You could maybe put it as the first thing in the application startup.

MNie:
@isaacabraham I updated the release notes. I also added branding to the `deploy`.

isaacabraham:
I think if you just put it into the Program.fs startup then that would catch everything rather than having to do it on a case-by-case basis?

MNie:
> I think if you just put it into the Program.fs startup then that would catch everything rather than having to do it on a case-by-case basis?

So we simply want to add branding to the `Sample` only?
image

isaacabraham:
Doh. Good point - I'm being stupid, Farmer is just a class library! Hmmm. Then I would say just in the core functions - deploy, validate, whatif etc. (these will be ones that hit Az CLI).

MNie:
@isaacabraham 

`execute`, `whatif`, `tryExecute`, `tryWhatIf`, `tryValidate` all of them execute `prepareForDeployment` where I add `branding`.

I also add it to the `quickWrite` from `Writer.fs`.

isaacabraham:
Thank you!

PR #415, opened 2020-10-22
#381 add account name validation for cosmos db

### LINKED ISSUE #381: CosmosDB Account name validation
As I tried to create a CosmosDB Account I entered a wrong name which wasnt catched by name validation like in storage account. I would prefer to have a validation in CosmosDB Account naming too.

This is the message from Azure CosmosDB Acount creation process:
"The name can contain only lowercase letters, numbers and the '-' character, and must be between 3 and 44 characters"

MNie:
This PR closes #381 

* add account name validation for cosmos DB

- Documentation update not needed,
- I test my code locally since it is only related to validation,
- I will update Release notes when PR would be accepted.

btw. I saw in a issue description, that we would need validation in couple of other places. We want to do it within this pr or via separate one?

`Hmmm. Not really - I'm not planning on implementing all validation rules, but rather putting in a pattern and set of types to do it. We'll then need to implement it across all resources as needed` ~Isaac

Because of above I create this PR as a Draft.

CC: @isaacabraham

### DIFF
--- a/src/Tests/Cosmos.fs
+++ b/src/Tests/Cosmos.fs
@@ -10,44 +10,70 @@ open Microsoft.Rest
 open System
 open Farmer.CoreTypes
 
-let tests = testList "Cosmos" [
-    test "Cosmos container should ignore duplicate unique keys" {
+let tests =
+    let invalidAccountNameCases =
+        [
+            "", "cannot be empty", "Name too short"
+            "zz", "min length is 3, but here is 2 ('zz')", "Name too short"
+            "abcdefghij1234567890abcde12345678901234567890", "max length is 44, but here is 45 ('abcdefghij1234567890abcde12345678901234567890')", "Name too long"
+            "zzzT", "can only contain lowercase letters ('zzzT')", "Upper case character allowed"
+            "zzz!", "can only contain alphanumeric characters or dash ('zzz!')", "Non alpha numeric (except dash) character allowed"
+        ]
+    let validAccountNameCases =
+        [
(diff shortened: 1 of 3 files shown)


src/Tests/Cosmos.fs
```diff
@@ -10,44 +10,70 @@ open Microsoft.Rest
...
+            let t = arm { add_resource (cosmosDb { name "test" }) }
 
-        t.Template
-        |> Writer.toJson
-        |> ignore
-    }
-    test "Creates connection string and keys with resource groups" {
-        let conn = CosmosDb.getConnectionString(ResourceId.create("db", "group"), PrimaryConnectionString).Eval()
-        let key = CosmosDb.getKey(ResourceId.create("db", "group"), PrimaryKey, ReadWrite).Eval()
+            t.Template
```
isaacabraham, line 56: Why is everything indented here?

isaacabraham (changes requested):

isaacabraham:
@MNie this is good, I think we can adopt this for now. I'm thinking of pushing the validation further back - maybe even to v2 - as it might involve some sizeable changes to the type system.

src/Tests/Cosmos.fs
MNie: I create `invalidAccountNameCases` and `invalidAccountNameCases` inside of `tests` this is why everything is indented. I move those 2 variables outside of `tests` and rollback the formatting changes since they introduce too much noise in here.

isaacabraham (approved):

isaacabraham:
Thanks for this @MNie !

PR #418, opened 2020-10-23
Feature/417/fixed zone

### LINKED ISSUE #417: Builders.Storage WebsitePrimaryEndpoint has a hardcoded zone name
Farmer assumes `z6` for the zone name in the website primary endpoints of Azure Storage. I don't quite know what zones correspond to, but I know [zone names aren't fixed](https://stackoverflow.com/questions/54277273/get-the-primary-static-website-endpoint-from-azure-powershell). For example, one I just deployed is `z8`.

https://github.com/CompositionalIT/farmer/blob/ae1979ff45b869d858b19a3833d3d2c410e4b65f/src/Farmer/Builders/Builders.Storage.fs#L55

This might be fixable by generating a 
```
[reference(parameters('storageAccountName'), '2019-04-01').primaryEndpoints.web]
```
which is what I use in my ARM templates.

MNie:
This PR closes #417

* parametrize WebsitePrimaryEndpoint

- Documentation update not needed.

CC:
@isaacabraham 

I don't know if this is a proper way of passing `reference` in other places I saw that `ArmBuilder` was used but it doesn't give the expected result.

### DIFF
--- a/src/Farmer/Builders/Builders.Storage.fs
+++ b/src/Farmer/Builders/Builders.Storage.fs
@@ -54,8 +54,8 @@ type StorageAccountConfig =
     /// Gets the Primary endpoint for static website (if enabled)
     member this.WebsitePrimaryEndpoint =
         ArmExpression
-            .create(sprintf "reference(%s, '2019-04-01').primaryEndpoints.web" this.Name.ResourceName.Value)
-            .Eval()
+            .reference(storageAccounts, ResourceId.create(storageAccounts, this.Name.ResourceName))
+            .Map(sprintf "%s.primaryEndpoints.web")
     member this.Endpoint = sprintf "%s.blob.core.windows.net" this.Name.ResourceName.Value
     interface IBuilder with
         member this.DependencyName = this.Name.ResourceName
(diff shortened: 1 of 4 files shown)


isaacabraham:
I think that this should change to an ArmExpression rather than a raw string (we can use Eval() to wrap in the []) - this will help show that this isn't a plain string but an expression that gets evaluated later.

MNie:
> I think that this should change to an ArmExpression rather than a raw string (we can use Eval() to wrap in the []) - this will help show that this isn't a plain string but an expression that gets evaluated later.

`Eval()` this is what I'm looking for! Pr updated : )

isaacabraham:
We probably need to document this stuff on the site to make it easier for contributions in future.

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -52,7 +52,10 @@ type StorageAccountConfig =
     /// Gets the ARM expression path to the key of this storage account.
     member this.Key = StorageAccount.getConnectionString this.Name
     /// Gets the Primary endpoint for static website (if enabled)
-    member this.WebsitePrimaryEndpoint = sprintf "https://%s.z6.web.core.windows.net" this.Name.ResourceName.Value
+    member this.WebsitePrimaryEndpoint =
+        ArmExpression
+            .create(sprintf "reference(%s, '2019-04-01').primaryEndpoints.web" this.Name.ResourceName.Value)
+            .Eval()
```
isaacabraham, line 58: I would even consider returning this as an ARM Expression and let the consumer decide when they want to Eval it.

isaacabraham:
Thanks

PR #430, opened 2020-10-24
Update quickstart-2.md

NRKirby:
The storage account connection key defined in the `WebAppBuilder` isn't the same as the key used in the rest of the document. This commit fixes that


isaacabraham:
Thanks a lot!

PR #432, opened 2020-10-25
Resourceids for all

isaacabraham:
This PR is a WIP which adds `ResourceIds` to many more parts of the codebase. There's more testing to be done here as this change affects a lot of the codebase and many of the issues will only be visible when deploying to Azure.


isaacabraham:
There is now at least one regression JSON file which tests the output JSON against a number of resources to check for regressions in terms of emitting ARM JSON. We can add to this file over time as needed.

The benefits of this PR is that many, many more resources are now fully qualified in ARM - although this does add a bit more fluff into the ARM, it means that we massively reduce the possibility of clashes (two resources of different types but same name) and things becomes a lot more consistent.

isaacabraham:
@ninjarobot what do you think of this?

isaacabraham:
There's a lot to look at but at a high level:

1. Dependencies are now ResourceIds.
2. You *cannot* create a ResourceId without a Resource Type (although you *can* manually create an empty resource type if you really want).
3. Builders and ArmResources both now are identified by ResourceId, not by ResourceName.

There are now some extra overloads so that you can create a resourceId from a resource type e.g. `servers.resourceId("myserver")`.

PR #437, opened 2020-10-28
Add support for executing command line arguments in container instance.

### LINKED ISSUE #402: Add command to containerInstance in ACI
I would like to be able to specify additional commands to run when starting a container in Azure Container Instances. 
The `command` in properties in the container instance:

```json
"containers": [
      {
        "name": "string",
        "properties": {
          "image": "string",
          "command": [
            "string"
          ],    
        }
      }
    ],
```

It looks simple enough to add a new custom-operation `command_line` to the containerInstance CE and add the `command` array to the json arm output. I’d be happy to make a PR if there are no other things I might have missed?

viktorvan:
This PR closes #402 

* Add support for command line arguments for container instances in a container group

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,9 +1,7 @@
 Release Notes
 =============
-## 1.3.1
-* Container Instance: Support for command line arguments
-
 ## 1.3.0
+* Container Instance: Support for command line arguments
 * Storage Account: Support for the full set of Storage Account Kind and SKUs (minor breaking change).
 
 ## 1.2.0
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,8 @@
 Release Notes
 =============
+## 1.3.1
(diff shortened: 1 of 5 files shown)


isaacabraham:
Thank you :-) Merged. I wonder how this relates to #403

viktorvan:
I don't know what limitations (if any) there are to the azure cli commands you can run in the deploymentscripts (supported by #403).

It looks like it's generally possible to [execute a command in a container instance using azure cli](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-exec):
`az container exec --resource-group <group-name> --name <container-group-name> --exec-command "<command>"`

But I see that as a different use case from the command_line parameters in this PR.

For example, it is not uncommon for a docker image to have a default start behavior which is then possible to override using the exec command line argument when creating the container instance. 
If you were to deploy the container without setting any command line arguments and then run a deployment script I think you might not get the behavior you expect, maybe the default startup runs and then your command from the deploymentscript resource. I am not sure how the timing works out.

PR #436, opened 2020-10-28
#434 add premium V3,

MNie:
- update cognitive services SKU

This PR closes #434

The changes in this PR are as follows:

* Add premiumV3 to WebApp
* Add S5/S6/S7/S8 for CognitiveServices

I have read the [contributing guidelines](CONTRIBUTING.md) and have completed the following:

* [ ] **Tested my code** end-to-end against a live Azure subscription.
* [X] **Updated the documentation** in the docs folder for the affected changes.
* [X] **Written unit tests** against the modified code that I have made.
* [X] **Updated the [release notes](RELEASE_NOTES.md)** with a new entry for this PR.
* [X] **Checked the coding standards** outlined in the [contributions guide](CONTRIBUTING.md) and ensured my code adheres to them.

If I haven't completed any of the tasks above, I include the reasons why here:

CC: @isaacabraham 

I saw that we don't give an ability to select S5/S6/S7/S8 for Bing.Search as noted [here](https://docs.microsoft.com/en-us/rest/api/cognitiveservices/accountmanagement/resourceskus/list). I add this option but I also include a validation regarding the kind & sku because S5/S6/S7/S8 are available only for Bing.Search.

I also update the plans for WebApp in the same PR.

Questions I have:
- Is it okay to do it in a single PR or should I create separate PRs?
- Validation for kind & sku seems reasonable or we don't want to do this?

Because of the above questions, I created the PR as a draft.

PR #440, opened 2020-11-02
Update Cognitive services sku

### LINKED ISSUE #526: Remove BingSearch from Cognitive Services.
[BingSearch API would be moved out from the Cognitive Services](https://azure.microsoft.com/en-us/updates/bing-search-apis-will-transition-from-azure-cognitive-services-to-azure-marketplace-on-31-october-2023/). We could align this change in the #440.

CC: @isaacabraham?

MNie:
This PR closes #526 

* update Cognitive services SKU (extracted from #436 + fixes after the first review)


isaacabraham:
Any chance that you can do an end-to-end test of this please?

isaacabraham:
If you're not able to test this out end-to-end, let me know. Cheers.

MNie:
@isaacabraham sorry I missed your comments :(. I will try to test it by the end of this week.

isaacabraham:
@MNie I've merged the latest from master into this branch and fixed the build - it looks like you had inadvertently rolled back the changed to Storage that had happened?

src/Farmer/Common.fs
```diff
@@ -474,6 +479,26 @@ module CognitiveServices =
         | TextAnalytics
         | TextTranslation
 
+    type CognitiveServicesSku =
```
isaacabraham, line 482: I wonder whether we really need this validation? Perhaps we can encode this rule within the type system in a similar way that we have done with Storage accounts.

isaacabraham: In fact, if Bing Search is to be removed anyway (per #526) then we can remove this entire validation.

isaacabraham:
@MNie Am I correct that if we remove Bing services from this, the new SKUs are no longer required anyway? In which case, this PR if simply about removing the obsolete Bing services from Cognitive Services and that's it? :-)

MNie:
@isaacabraham yes, you're right :). I will update tomorrow morning the pr. I could provide a separate pr with a fix to #526 when it would be ready. 

> it looks like you had inadvertently rolled back the changed to Storage that had happened?

Strange, I thought I did a simple merge and resolve conflicts (I have a single one in the readme.md). I will take a look tomorrow morning at what happens.

MNie:
@isaacabraham I remove BingSearch options from CognitiveServices CE and add a new BingSearch CE.

It seems that BingSearch has only one kind and then all options are available via Rest calls:
image

And I see that a free tier is `F1` instead of `F0`.

src/Farmer/Arm/BingSearch.fs
```diff
@@ -0,0 +1,23 @@
...
+let accounts = ResourceType ("Microsoft.Bing/accounts", "2020-06-10")
+[<Literal>]
+let private kind = "Bing.Search.v7"
+
+type Accounts =
+    { Name: ResourceName
+      Location: Location
+      Sku: BingSearch.Sku
+      Tags: Map<string,string>
+      Properties: {| statisticsEnabled: bool |} }
```
isaacabraham, line 15: Just move statistics to the top level. Consider replacing bool with Enabled / Disabled.

isaacabraham: The ARM type shouldn't be a 1:1 mapping with the JSON but an F# view on it.

src/Farmer/Builders/Builders.BingSearch.fs
```diff
@@ -0,0 +1,50 @@
...
+    /// Gets an ARM Expression key for any Bing Search instance.
+    static member getKey (resourceId: ResourceId) =
+        ArmExpression.create(sprintf "listKeys(%s, '%s').key1" resourceId.ArmExpression.Value accounts.ApiVersion, resourceId)
+    static member getKey (name: ResourceName) = BingSearch.getKey (accounts.resourceId name)
+
+type BingSearchConfig =
+    { Name: ResourceName
+      Sku: Sku
+      Tags: Map<string,string>
+      StatisticsEnabled: bool }
```
isaacabraham, line 18: Same comment here re: stats field.

src/Farmer/Builders/Builders.BingSearch.fs
```diff
@@ -0,0 +1,50 @@
...
+        { Name = ResourceName.Empty
+          Sku = F1
+          Tags = Map.empty
+          StatisticsEnabled = false }
+    [<CustomOperation "name">]
+    member _.Name (state:BingSearchConfig, name) = { state with Name = ResourceName name }
+    [<CustomOperation "sku">]
+    member _.Sku (state:BingSearchConfig, sku) = { state with Sku = sku }
+    [<CustomOperation "enable_statistics">]
+    member _.EnableStatistics (state:BingSearchConfig) = { state with StatisticsEnabled = true }
```
isaacabraham, line 42: Consider passing in a boolean for this - this enables better programmatic access.

src/Farmer/Builders/Builders.BingSearch.fs
```diff
@@ -0,0 +1,50 @@
...
+    [<CustomOperation "add_tags">]
```
isaacabraham, line 43: Remove Tag keywords and replace with `ITaggable`.

src/Tests/BingSearch.fs
```diff
@@ -0,0 +1,46 @@
...
+        }
+        let baseArm = (swa :> IBuilder).BuildResources(Location.WestEurope).[0]
+        let bsArm = baseArm :?> BingSearch.Accounts
+        let model = baseArm.JsonModel |> convertTo<{| kind: string |}>
+        Expect.equal bsArm.Name (ResourceName "test") "Name"
+        Expect.equal bsArm.Location Location.WestEurope "Location"
+        Expect.isTrue bsArm.Properties.statisticsEnabled "Statistics enabled"
+        Expect.equal bsArm.Sku S0 "Sku"
+        Expect.equal model.kind "Bing.Search.v7" "kind"
+        Expect.equal bsArm.Tags (tags |> Map.ofList) "Tags"
```
isaacabraham, line 28: x |> Map.ofList can be replaced with just Map x.

isaacabraham (changes requested):

src/Farmer/Builders/Builders.BingSearch.fs
MNie: I pass a `FeatureFlag` so we would have:

```fsharp
bingSearch {
  name "test"
  statistics Enabled
}
```

src/Farmer/Arm/BingSearch.fs
MNie: Good hint, fixed : )

src/Tests/BingSearch.fs
```diff
@@ -38,7 +38,7 @@ let tests = testList "Bing Search" [
         let model = baseArm.JsonModel |> convertTo<{| kind: string |}>
         Expect.equal bsArm.Name (ResourceName "test") "Name"
         Expect.equal bsArm.Location Location.WestEurope "Location"
-        Expect.isFalse bsArm.Properties.statisticsEnabled "Statistics enabled"
+        Expect.equal bsArm.Statistics Disabled "Statistics not enabled"
```
isaacabraham, line 41: Revert back to `isFalse`; there's a member on the Enabled / Disabled called `toBool` or something.

isaacabraham (approved):

src/Tests/BingSearch.fs
MNie: But `equal Statistics Disabled` isn't more meaningful than `isFalse Statistics`?

isaacabraham: hmmm. I thought:

```fsharp
Expect.isFalse bsArm.Statistics.AsBoolean "Statistics not enabled"
```

That's what the `isFalse` function is for - to test against booleans. I suppose I'm thinking of Enable / Disabled as a bool here, but maybe you're right.

The alternative (maybe preferably) would be to pull at the JSON boolean value into `model` (like you've done with `kind`) and compare to the boolean directly. That's actually my preference for all tests wherever possible because that then tests the actually emitted JSON.

MNie: @isaacabraham I updated the JSON model to include also a `statisticsEnabled` field and I test it against `isFalse`/`isTrue`. I keep also a check for `bsArm.Statistics`.

isaacabraham:
Hi @MNie. I've merged up from master and I think the only thing outstanding for this is doing the docs - not much fun but needs to be done in order to get it in.

Will you be able to do that? If not, just let me know. Ta.

MNie:
@isaacabraham sorry for the long delay in responding here. I updated the docs + added a note in the `cognitive-services` doc (if we don't want to include information about extracting BingSearch, please give me the info I will revert this one).

PR #450, opened 2020-11-08
Adds flexible types for depends_on IBuilder list

### LINKED ISSUE #449: Make it easier to add a list of resources
Currently this does not work in Farmer:

```fsharp
let storageAccounts = [ storageAccount { name "aaa" }; storageAccount { name "bbb" } ]
let app = webApp { depends_on storageAccounts }
```

You have to *explicitly* mark `storageAccounts` as an `IBuilder list`:

```fsharp
let storageAccounts : IBuilder list = [ storageAccount { name "aaa" }; storageAccount { name "bbb" } ]
let app = webApp { depends_on storageAccounts }
```

or 

```fsharp
let storageAccounts = [ storageAccount { name "aaa" } :> IBuilder; storageAccount { name "bbb" } :> _ ]
let app = webApp { depends_on storageAccounts }
```

This catches out a lot of people and is not immediately obvious, especially if you're coming from C#.

This is actually possible using [F# flexible types](https://docs.microsoft.com/en-us/dotnet/fsharp/language-reference/flexible-types) so we should create a prototype of this for testing.

isaacabraham:
This PR closes #449.

* All Builders that provide `depends_on` now use `#IBuilder list` instead of `IBuilder list`.
* Extra tests

There are several limitations here:

1. It won't work simultaneously if you make the `IArmResource list` overload flexible as well.
2. The type signature of this overload is a bit weird as it's a generic type that has a constraint that implements the `IBuilder` interface.
3. You still can't implicitly create an `IBuilder list` if the contents of the list are multiple builders. In this case, you need to explicitly add a type annotation.


isaacabraham:
Putting this PR on hold. It causes some existing combinations to fail that previously worked, because of overload resolution nonsense.

PR #460, opened 2020-11-14
Deployment script improvements

ninjarobot:

* Specifies cleanup on expiration when retention interval is set, and enables cleanup on success only.
* Support for running a deployment script after other resources are deployed.

### DIFF
--- a/src/Farmer/Builders/Builders.DeploymentScript.fs
+++ b/src/Farmer/Builders/Builders.DeploymentScript.fs
@@ -17,9 +17,9 @@ type OutputCollection(owner) =
 
 type DeploymentScriptConfig =
     { Name : ResourceName
-      AdditionalDependencies : ResourceId list
+      Dependencies : ResourceId Set
       Arguments : string list
-      CleanupPreference : Cleanup option
+      CleanupPreference : Cleanup
       Cli : CliVersion
       EnvironmentVariables: Map<string, EnvVar>
       ForceUpdate : bool
@@ -61,7 +61,7 @@ type DeploymentScriptConfig =
             // Deployment Script
             { Location = location
               Name = this.Name
-              AdditionalDependencies = this.AdditionalDependencies
+              Dependencies = this.Dependencies
               Arguments = this.Arguments
               CleanupPreference = this.CleanupPreference
               Cli = this.Cli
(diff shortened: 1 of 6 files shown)


isaacabraham:
@ninjarobot I've merged from master up to this and also made some minor changes:

* `Cleanup` is no longer optional. As it was defaulting to Always at the very end, I've simply moved that defaulting up the chain to the builder level, so it's clear to callers what the Cleanup model is by default.
* Replaced `run_after` with the standard `depends_on` pattern. This also means that you can simply pass a builder or IArmResource directly in. I suspect that this will be the common pattern because if you're relying on an external resource (outside of Farmer) it almost certainly will already exist anyway, in which case the `depends_on` isn't necessary anyway.
* Slight refactoring of the ARM generation (readability / clarity).

Only one question left which I'm just about to add as a comment here, otherwise I think this is good to go.

src/Farmer/Builders/Builders.DeploymentScript.fs
```diff
@@ -126,8 +136,8 @@ type DeploymentScriptBuilder() =
     /// Time to retain the container instance that runs the script - 1 to 26 hours.
     [<CustomOperation "retention_interval">]
     member _.RetentionInterval(state:DeploymentScriptConfig, retentionInterval) =
-        let maxRetention = min retentionInterval 30<Hours>
-        { state with RetentionInterval = Some (TimeSpan.FromHours (float maxRetention)) }
+        let maxRetention = min retentionInterval 26<Hours>
```
isaacabraham, line 135: More and more I'm starting to think that we should get rid of these implicit "fixing bad data" issues and simply fail fast with e.g. a failwith - it's clearer to the caller what's gone wrong (and the error message can / should make it clear how to fix it), rather than them assume 100 hours has worked and then realised later on that 26 hours is the max. What do you think?

ninjarobot: Yes, that probably would be better. Part of me really was frustrated that this was buried in docs while I was just trying to get a valid template and I wanted to minimize the chance someone else would trip on that same issue. As long as we fail fast, I think that is good.

isaacabraham:
Thanks for this!

PR #465, opened 2020-11-18
Secure parameters for environment vars in ACI and deployment scripts

ninjarobot:

* Adds support for secure parameters for environment variables in container groups and deployment scripts

### DIFF
--- a/src/Farmer/Common.fs
+++ b/src/Farmer/Common.fs
@@ -686,7 +686,9 @@ module ContainerGroup =
         | PublicAddressWithDns of DnsName:string
         | PrivateAddress
     /// A secret file which will be encoded as base64 and attached to a container group.
-    type SecretFile = SecretFile of Name:string * Secret:byte array
+    type SecretFile =
+            | SecretFile of Name:string * Secret:byte array
+            | SecretParameter of Name:string * Secret:SecureParameter
     /// A container group volume.
     [<RequireQualifiedAccess>]
     type Volume =
--- a/src/Farmer/Common.fs
+++ b/src/Farmer/Common.fs
@@ -69,13 +69,12 @@ type IsoDateTime =
 type TransmissionProtocol = TCP | UDP
 type TlsVersion = Tls10 | Tls11 | Tls12
 type EnvVar =
+    /// Use for non-secret environment variables to be surfaced in the container. These will be stored in cleartext in the ARM template.
     | EnvValue of string
-    | SecureEnvValue of string
-    | SecureParamEnvValue of SecureParameter
+    /// Use for secret environment variables to be surfaced in the container securely. These will be provided as secure parameters to the ARM template.
+    | SecureEnvValue of SecureParameter
     static member create (name:string) (value:string) = name, EnvValue value
-    static member createSecure (name:string) (value:string) = name, SecureEnvValue value
-    static member createSecureParameter (name:string) (paramName:string) =
-        name, SecureParamEnvValue (SecureParameter paramName)
+    static member createSecure (name:string) (paramName:string) = name, SecureEnvValue (SecureParameter paramName)
 
 module Mb =
     let toBytes (mb:int<Mb>) = int64 mb * 1024L * 1024L
--- a/src/Farmer/Common.fs
+++ b/src/Farmer/Common.fs
@@ -679,10 +679,12 @@ module ContainerGroup =
         | PublicAddress
         | PublicAddressWithDns of DnsName:string
         | PrivateAddress
-    /// A secret file which will be encoded as base64 and attached to a container group.
+    /// A secret file that will be attached to a container group.
     type SecretFile =
-        | SecretFile of Name:string * Secret:byte array
-        | SecretParameter of Name:string * Secret:SecureParameter
+        /// A secret file which will be encoded as base64 data.
+        | SecretFileContents of Name:string * Secret:byte array
+        /// A secret file which will provided by an ARM parameter at runtime.
+        | SecretFileParameter of Name:string * Secret:SecureParameter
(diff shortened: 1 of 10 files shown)


src/Farmer/Common.fs
```diff
@@ -71,8 +72,11 @@ type TlsVersion = Tls10 | Tls11 | Tls12
 type EnvVar =
     | EnvValue of string
     | SecureEnvValue of string
+    | SecureParamEnvValue of SecureParameter
     static member create (name:string) (value:string) = name, EnvValue value
     static member createSecure (name:string) (value:string) = name, SecureEnvValue value
+    static member createSecureParameter (name:string) (paramName:string) =
```
ninjarobot, line 78: I added a helper here because it didn't seem reasonable to try to generate the parameter name like in other places. We could generate it off the environment variable name, but not knowing what container or container group or deployment script it's in, this seemed like it could get really confusing for the user.

isaacabraham:
Thanks. I hope it wasn't too bad!

isaacabraham:
Haven't done a review yet - too busy at the moment I'm afraid.

src/Farmer/Common.fs
```diff
@@ -71,8 +71,11 @@ type TlsVersion = Tls10 | Tls11 | Tls12
...
```
isaacabraham, line 77: What's the difference between a `SecureEnvValue` and a `SecureParamEnvValue`?

src/Farmer/Common.fs
```diff
@@ -677,7 +680,9 @@ module ContainerGroup =
         | PublicAddressWithDns of DnsName:string
         | PrivateAddress
     /// A secret file which will be encoded as base64 and attached to a container group.
-    type SecretFile = SecretFile of Name:string * Secret:byte array
+    type SecretFile =
+            | SecretFile of Name:string * Secret:byte array
```
isaacabraham, line 684: I think that we should consider renaming either the one (or both) of the cases or the overall type - something doesn't feel right here. Either it's a `SecretFile` and the first case needs renaming, or the first case is a `SecretFile` and the top type name should be something different.

Or something else entirely :-)

isaacabraham (changes requested):
Looks good apart from the above questions (which are probably just my lack of knowledge).

One other thing: There is now an ARM regression test in `Template.fs` which contains a number of resources that are added to a template; the template file is saved and compared on every build to ensure that we don't accidentally modify the ARM template output. You could maybe create another one for containers? Or add some container details into that test?

src/Farmer/Common.fs
ninjarobot: Both of these can be used to populate an environment variable that will be provided to the container but aren't visible in the UI or API when viewing an existing container group. A `SecureParamEnvValue` will generate an ARM template parameter to be used for the environment variable so that it won't be hardcoded into the generated template. This is helpful because you can use a key vault secret reference for a parameter and ARM can pull it from the key vault and then no human ever interacts with the secret value.

ninjarobot: This will create a `SecureParameter` to populate the secret environment variable value - this means it will generate an ARM template parameter for that environment variable using the `paramName` for the name of that parameter. This allows you to pass in a key vault reference so nobody every interacts with the secrets when performing the deployment.

src/Farmer/Common.fs
ninjarobot: In either case, it ends up creating a [secret file that can be mounted in the container(s) in the group](https://docs.microsoft.com/en-us/azure/container-instances/container-instances-volume-secret). It's just that in one case, you can provide the data for the file directly and in another case, it's passed in as an ARM deployment parameter, again so that you can use a key vault reference.

isaacabraham:
@ninjarobot apologies for not looking at this sooner. Thanks for clarifying - I think my question really is: do we need all three ways of setting environment variables? Can / should we be a little bit more prescriptive here and only offer one of them (or at most two) - I get the parameterised one (so not to put the value in the ARM template - we do that everywhere e.g. SQL password) but then do we need the `SecureEnvValue` version: should we allow people to put secure values directly in the ARM template?

If I understand it correctly, my instincts say "no". What do you think?

src/Farmer/Common.fs
isaacabraham: That makes sense - thanks for explaining this.

isaacabraham: In the keyvault version, what would be stored in the secret - the UTF8-encoded byte array?

isaacabraham: Is the parameter directly designed for KeyVault (in which case we can supply something like a KeyVault ref) or is it simply a parameter that could be used to supply the encoded data directly as an argument?

isaacabraham: I'm now thinking that perhaps we shouldn't allow people to supply the secret data directly in the template and *force* it to be a parameter!

ninjarobot: I just meant [this](https://docs.microsoft.com/en-us/azure/azure-resource-manager/templates/key-vault-parameter?tabs=azure-cli#reference-secrets-with-static-id) where you can reference a key vault secret in the parameter file so ARM will pull it. This isn't anything specific to this PR, just something you can do with deployment parameters in general.

isaacabraham:
Would you have time for a call this week? I really want to get this in given that you've taken the time to do it and it looks a good feature - I just want to make sure that I understand the considerations correctly and that the ergonomics are right going forward.

If we need to make a breaking change in terms of compilation, so be it, if it means that we're making things more secure and that there's a simple path forwards.

isaacabraham (approved):

isaacabraham:
@ninjarobot it's in :-)

PR #471, opened 2020-11-27
Add `default_blob_access_tier` operator to storage builder

### LINKED ISSUE #454: Storage account access tier 'cool'
Is there any way to set the access tier on a storage account object? My understanding is that there is a significant cost difference between coold and hot tier access so would like to configure the tier to reduce cost.

```fsharp
  let storageAcc = storageAccount {
    name "myStorageAccount" 
    sku Storage.Sku.Standard_LRS
    
    use_static_website contentPath "index.html"
  }
```
produces
```json
    {
      "apiVersion": "2019-06-01",
      "dependsOn": [],
      "kind": "StorageV2",
      "location": "uksouth",
      "name": "myStorageAccount",
      "properties": {},
      "sku": {
        "name": "Standard_LRS"
      },
      "type": "Microsoft.Storage/storageAccounts"
    }
```
I believe the difference I would like is
```diff
      "location": "uksouth",
      "name": "myStorageAccount",
-     "properties": {},
+     "properties": {
+        "accessTier": "Cool"
+     },
      "type": "Microsoft.Storage/storageAccounts"
    }
```

TheRSP:
This PR closes #454

* Added `default_blob_access_tier` operator to storage builder

### DIFF
--- a/src/Farmer/Arm/Storage.fs
+++ b/src/Farmer/Arm/Storage.fs
@@ -16,7 +16,6 @@ type StorageAccount =
     { Name : StorageAccountName
       Location : Location
       Sku : Sku
-      DefaultBlobAccessTier: BlobAccessTier Option
       Dependencies : ResourceId list
       EnableHierarchicalNamespace : bool option
       StaticWebsite : {| IndexPage : string; ErrorPage : string option; ContentPath : string |} option
@@ -30,7 +29,7 @@ type StorageAccount =
                         let performanceTier =
                             match this.Sku with
                             | GeneralPurpose (V1 (V1Replication.LRS performanceTier))
-                            | GeneralPurpose (V2 (V2Replication.LRS performanceTier)) ->
+                            | GeneralPurpose (V2 (V2Replication.LRS performanceTier, _)) ->
                                 performanceTier.ToString()
                             | Files _
                             | BlockBlobs _ ->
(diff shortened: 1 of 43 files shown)


src/Tests/Storage.fs
```diff
@@ -183,4 +183,21 @@ let tests = testList "Storage Tests" [
...
         Expect.equal resource.Sku.Name "Premium_LRS" "Sku Name"
     }
+    test "Sets blob access tier correctly different SKU kinds correctly" {
+        let account = storageAccount { default_blob_access_tier Cool }
+        let resource = arm { add_resource account } |> getStorageResource
+        Expect.equal resource.AccessTier (Nullable AccessTier.Cool) "Access Tier"
+
+        let account = storageAccount { default_blob_access_tier Hot }
+        let resource = arm { add_resource account } |> getStorageResource
+        Expect.equal resource.AccessTier (Nullable AccessTier.Cool) "Access Tier"
```
isaacabraham, line 193: Shouldn't this be Hot?

src/Farmer/Arm/Storage.fs
```diff
@@ -16,6 +16,7 @@ type StorageAccount =
     { Name : StorageAccountName
       Location : Location
       Sku : Sku
+      DefaultBlobAccessTier: BlobAccessTier Option
```
isaacabraham, line 19: I'm not sure at the ARM level that we should be moving this (probably makes sense at the Builder level though). Does that default access tier apply across all different storage kinds e.g. General Purpose, Blobs, Files etc. or just a subset of them?

src/Tests/Storage.fs
TheRSP: Yes. Don't know how that slipped through :(

src/Farmer/Arm/Storage.fs
TheRSP: Looking into this further, while it's valid to set for all kinds except StorageV1, it only seems to affect StorageV2 and Blob Storage. 
Would you prefer adding the accessTier to the storageV2 sku?
Given that the SKU is used at both arm and builder levels, I'm not sure how best to separate those?

isaacabraham: Mmmm. Either we'll need a new type just for the builder (seems overkill) or we put that default flag against just those two kinds at both builder + ARM levels. We can still expose it as a keyword at the builder level - if someone tries to set it against an invalid storage account kind, throw an exception.

isaacabraham:
Hi @TheRSP. I've pushed to the PR branch per the discussion above. Can you have a look through if you're happy with this? The only real changes are moving the default access tier into those two storage kinds, and also renaming the DU (there are three access tiers, but only two for default settings).

TheRSP:
@isaacabraham This all looks good to me. Thanks.

isaacabraham:
I'll merge this in once 1.3.x is released.

isaacabraham:
Thanks again for doing this.

PR #494, opened 2020-12-13
Add support for creating Databricks workspaces

### LINKED ISSUE #480: Add Support for Databricks Workspaces
It would be nice to be able to deploy Databricks workspaces using both the managed network and Bring Your Own Vnet model. Databricks workspaces are a core component of all the ARM templates that I work on. 

Has this been requested and denied previously? Are there any known limitations to implementing this support?

I'd be happy to work on this feature. It would be my first contribution to an OSS project.

cmarshall10450:
This PR closes #480

* Adds databricksWorkspace builder for adding Databricks Workspaces to ARM templates

- [x] DataBricks SKU (name + tier - name mandatory, tier is optional?)
- [x] Nicer integration with VNets
- [ ] Storage account integration (with managed identity)


src/Farmer/Builders/Builders.Databricks.fs
```diff
@@ -0,0 +1,124 @@
...
+    member _.PrepareEncryption(state:WorkspaceConfig) =
+        { state with 
+            PrepareEncryption = Some true
+            Encryption = Some {
+                KeySource = "Microsoft.Keyvault";
+                KeyName = "";
+                KeyVersion = "latest";
+                KeyVaultUri = ""} }
+    ///Set encryption key name
+    [<CustomOperation "key_name">]
```
isaacabraham, line 82: Might it be worth merging some of these keywords together - it looks like you should never set the Key Vault without setting the name / version. I also wonder whether we can just omit the `enable_encryption` keyword - if you supply a keyvault, it enables encryption implicitly?

isaacabraham (changes requested):
Thank you for doing this. I've put some review comments in. Some are boilerplate formatting but a few might require a more detailed discussion - perhaps a screenshare / call.

docs/content/api-overview/resources/databricks-workspace.md
```diff
@@ -0,0 +1,49 @@
...
+
+* Workspace (`Microsoft.Databricks/workspaces`)
+
+#### Builder Keywords
+| Keyword | Purpose |
+|-|-|
+| name | Sets the name of the workspace. |
+| pricing_tier | Sets the pricing tier of the workspace. Defaults to Standard Tier |
+| enable_encryption | Turns on workspace encryption encryption. |
+| key_vault | Sets the name of the key vault where the encryption key is stored |
```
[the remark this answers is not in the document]
cmarshall10450: Would you be able to give me some more information on what ResourceId and ResourceName are used for and where/when each of them should be used. I'm not sure I fully understand them.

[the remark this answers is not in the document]
isaacabraham: Sure.

A `ResourceName` represents the raw name of a resource (as the name suggests ;-)). In F# it's common to try to avoid working with raw strings to avoid accidentally mixing two strings that mean semantically different things - thing of it is a wrapper around a string.

A `ResourceId` is a type that contains the full *path* to a resource - so it contains the name, but also optionally the resource group and the full path to the resource. We'll be adding subscription id to this at some point soon, too.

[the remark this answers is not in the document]
isaacabraham: If this is confusing, I'm happy to clean this up myself and you can see the diff of what I've done as a result.

cmarshall10450:
Thanks for all of your comments and advice. They were very helpful in improving the code. I hope I've managed to resolve most, if not all, of your comments. Please let me know if you see something that I've missed, still needs to be improved, or if I've misunderstood what you meant.

docs/content/api-overview/resources/databricks-workspace.md
```diff
@@ -0,0 +1,49 @@
...
+| key_name | Sets the name of the encryption key secret|
+| key_version | Sets the version of the encryption key secret. Defaults to latest |
+| disable_public_ip | Turn off public IP addresses for cluster virtual machines. Defaults to false|
+| enable_byov_mode | Turn on Bring Your Own VNet networking mode. Defaults to false |
```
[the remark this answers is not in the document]
cmarshall10450: I've removed this in favour of setting the necessary defaults if byov_vnet, byov_public_subnet or byov_private_subnet  is set.

docs/content/api-overview/resources/databricks-workspace.md
```diff
@@ -0,0 +1,49 @@
...
+#### Overview
+The Databricks Workspace builder is used to create Azure Databricks Workspaces
+
+* Workspace (`Microsoft.Databricks/workspaces`)
+
+#### Builder Keywords
+| Keyword | Purpose |
+|-|-|
+| name | Sets the name of the workspace. |
+| pricing_tier | Sets the pricing tier of the workspace. Defaults to Standard Tier |
```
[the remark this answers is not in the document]
cmarshall10450: I've changed this to use sku to be more consistent with other resources.

docs/content/api-overview/resources/databricks-workspace.md
```diff
@@ -0,0 +1,49 @@
...
+| enable_encryption | Turns on workspace encryption encryption. |
```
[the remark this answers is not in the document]
cmarshall10450: I've changed this to be a feature flag that can be Enabled or Disabled

docs/content/api-overview/resources/databricks-workspace.md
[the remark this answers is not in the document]
cmarshall10450: Thanks for offering to clean it up. I've made an attempt at using ResourceRef and ResourceName. Hopefully I've used them as intended. The operation should now take a KeyVaultConfig, an Arm.KeyVault.Vault or a string as the name.

docs/content/api-overview/resources/databricks-workspace.md
```diff
@@ -0,0 +1,49 @@
...
+| key_vault | Sets the name of the key vault where the encryption key is stored |
+| key_name | Sets the name of the encryption key secret|
+| key_version | Sets the version of the encryption key secret. Defaults to latest |
+| disable_public_ip | Turn off public IP addresses for cluster virtual machines. Defaults to false|
+| enable_byov_mode | Turn on Bring Your Own VNet networking mode. Defaults to false |
+| byov_vnet | Sets the name of the VNet to use for BYOV mode |
```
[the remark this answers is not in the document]
cmarshall10450: This now accepts a string, an Arm.Network.VirtualNetwork or VirtualNetworkConfig

src/Farmer/Arm/Databricks.fs
```diff
@@ -0,0 +1,101 @@
...
+                    {| managedResourceGroupId = 
+                        let expr = 
+                            sprintf
+                                "concat(subscription().id, '/resourceGroups/', '%s')"
+                                this.ManagedResourceGroupId
+                        ArmExpression.create(expr).Eval()
+                       parameters = 
+                        {| customVirtualNetworkId = 
+                                match this.ByovMode with
+                                    | Some _ ->
```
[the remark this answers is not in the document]
cmarshall10450: The formatting has been addressed.

src/Farmer/Arm/Databricks.fs
```diff
@@ -0,0 +1,101 @@
...
+                                        match this.ByovConfig with
+                                            | Some config -> 
+                                                {| value = 
+                                                    let expr = 
+                                                        sprintf 
+                                                            "resourceId('Microsoft.Network/virtualNetworks', '%s')"
+                                                            config.Vnet
+                                                    ArmExpression.create(expr).Eval() |} |> box
+                                            | None -> null
+                                    | None -> null
```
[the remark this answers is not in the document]
cmarshall10450: This has been changed to use `Option.map` and `Option.toObj` to remove the match statements

src/Farmer/Arm/Databricks.fs
```diff
@@ -0,0 +1,101 @@
...
+                           customPublicSubnetName = 
+                                match this.ByovMode with
+                                    | Some _ -> 
+                                        match this.ByovConfig with
+                                            | Some config -> {| value = config.PublicSubnet |} |> box
```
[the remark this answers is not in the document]
cmarshall10450: I've changed it to use the box the way you suggested.

src/Farmer/Builders/Builders.Databricks.fs
```diff
@@ -0,0 +1,124 @@
...
+              Encryption = this.Encryption
+              ByovMode = this.ByovMode
+              ByovConfig = this.ByovConfig
+              Tags = this.Tags  }
+        ]        
+
+type WorkspaceBuilder() =
+    member __.Yield _ =
+      { 
+        ///Name of the workspace
```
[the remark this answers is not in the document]
cmarshall10450: The formatting has been addressed

src/Farmer/Builders/Builders.Databricks.fs
```diff
@@ -0,0 +1,124 @@
...
+        PrepareEncryption = None
+        ///Encryption configuration
+        Encryption = None
+        ///Should use Bring Your Own VNet networking model
+        ByovMode = None
+        ///Bring Your Own VNet configuration
+        ByovConfig = None
+        ///List of tags
+        Tags = Map.empty
+      }
```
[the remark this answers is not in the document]
cmarshall10450: The formatting has been addressed

src/Farmer/Builders/Builders.Databricks.fs
```diff
@@ -0,0 +1,124 @@
...
+    ///Set workspace name  
+    [<CustomOperation "name">]
+    member __.Name (state:WorkspaceConfig, name) =
```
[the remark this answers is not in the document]
cmarshall10450: The extra underscore has been removed.

src/Farmer/Builders/Builders.Databricks.fs
cmarshall10450: This has been changed to an operation called "encryption_key" that takes in a name and version. The version also defaults to "latest".

docs/content/api-overview/resources/databricks-workspace.md
```diff
@@ -0,0 +1,49 @@
...
+#### Builder Keywords
+| Keyword | Purpose |
+|-|-|
+| name | Sets the name of the workspace. |
+| pricing_tier | Sets the pricing tier of the workspace. Defaults to Standard Tier |
+| enable_encryption | Turns on workspace encryption encryption. |
+| key_vault | Sets the name of the key vault where the encryption key is stored |
+| key_name | Sets the name of the encryption key secret|
+| key_version | Sets the version of the encryption key secret. Defaults to latest |
+| disable_public_ip | Turn off public IP addresses for cluster virtual machines. Defaults to false|
```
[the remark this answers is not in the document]
cmarshall10450: This has been changed to enable_public_ip. There will always be some double negative logic as the ARM property is a double negative itself - "enableNoPublicIp" but the logic has been moved to the JsonModel

src/Farmer/Arm/Databricks.fs
```diff
@@ -0,0 +1,65 @@
...
+
+let workspaces = ResourceType ("Microsoft.Databricks/workspaces", "2018-04-01")
+
+type ByovConfig =
+    { Vnet: ResourceRef<ByovConfig>
+      PublicSubnet: ResourceName
+      PrivateSubnet: ResourceName }
+
+type EncryptionConfig =
+    { KeySource: string
```
isaacabraham, line 17: This should be made into a DU. See [here](https://docs.microsoft.com/en-us/azure/templates/microsoft.databricks/workspaces#encryption-object) - there are only a couple of cases.

samples/scripts/databricks.fsx
```diff
@@ -0,0 +1,29 @@
...
+open Farmer.Builders
+open Farmer.Databricks
+
+let workspace = databricksWorkspace {
+    name "my-databricks-workspace"
+    sku Databricks.Sku.Standard
+    
+    use_public_ip Enabled
+    
+    key_vault "databricks-kv"
```
isaacabraham, line 14: Am I correct that for KeyVault integration to work, you would need to supply all the four KV values (source, name, version and vault)? If so, we should probably feed that through in one of two ways:

1. Force the caller to supply all four values at once (this would also simplify the builder code).
2. At least have some validation in the Builder if they only entered some of the values. (see other builders - they do this in the `Run()` method.

samples/scripts/databricks.fsx
```diff
@@ -0,0 +1,29 @@
...
+    encryption_key "workspace-encryption-key" "latest"
+    
+    byov_vnet "databricks-vnet"
+    byov_public_subnet "databricks-pub-snet"
```
isaacabraham, line 18: Design question: Would you ever have public subnet and a private subnet at the same time? Is this permitted (if we're not sure, let's leave it as is).

src/Tests/Databricks.fs
```diff
@@ -0,0 +1,53 @@
+module Databricks
+
+open Expecto
+open Farmer
+open Farmer.Builders
+open Farmer.Databricks
+open System
+
+let tests = testList "Databricks Tests" [
```
isaacabraham, line 10: I think these tests could (should) go down to the ARM level at least i.e. you can upcast `workspace` to an `IBuilder` and then call `buildResources` to get the raw ARM output. I've modified the first test to illustrate this.

There's a final test we should add to which takes the *generated ARM* json into a big file and is used for regression testing - let's not forget about that, we can do it at the end.

src/Farmer/Builders/Builders.Databricks.fs
```diff
@@ -0,0 +1,143 @@
...
+    member this.ByovPublicSubnet(state:WorkspaceConfig, publicSubnet) = this.ByovPublicSubnet(state, ResourceName publicSubnet)
+
+    /// Sets the existing private subnet
+    [<CustomOperation "byov_private_subnet">]
+    member _.ByovPrivateSubnet (state:WorkspaceConfig, privateSubnet:ResourceName) =
+        let config =
+            state.ByovConfig
+            |> Option.map(fun vnetConfig -> { vnetConfig with PrivateSubnet = privateSubnet })
+            |> Option.orElse
+                (Some { Vnet = External(Unmanaged(virtualNetworks.resourceId ResourceName.Empty))
```
isaacabraham, line 127: I think this code is problematic. External Unmanaged means a resource that is created *outside* of Farmer. However, if you are creating a VNet as part of the Farmer template, it'll need to be specified as External Managed, and you'll need to provide this as a dependency onto the workspace to ensure that the VNet is created *before* the workspace.

Have a look at how some other resources do this - if you get stuck or it's not going, no worries - let me know, one of me or the team will get this over the line.

src/Farmer/Arm/Databricks.fs
```diff
@@ -0,0 +1,65 @@
...
+                    {| name =
+                        match this.Sku with
+                        | StandardTier -> "standard"
+                        | PremiumTier -> "premium" |}
+                properties =
+                    {| managedResourceGroupId =
+                        let expr = sprintf "concat(subscription().id, '/resourceGroups/', '%s')" this.ManagedResourceGroupId.Value
+                        ArmExpression.create(expr).Eval()
+                       parameters = Map [
+                        "enableNoPublicIp", box this.EnablePublicIp.AsBoolean
```
cmarshall10450, line 46: I'm not sure that this will work as intended. Enabling this option actually disables the use of public IPs so the boolean value of EnablePublicIp should be negated.

src/Farmer/Arm/Databricks.fs
```diff
@@ -0,0 +1,65 @@
...
+                sku =
+                    {| name =
+                        match this.Sku with
+                        | StandardTier -> "standard"
+                        | PremiumTier -> "premium" |}
+                properties =
+                    {| managedResourceGroupId =
+                        let expr = sprintf "concat(subscription().id, '/resourceGroups/', '%s')" this.ManagedResourceGroupId.Value
+                        ArmExpression.create(expr).Eval()
+                       parameters = Map [
```
isaacabraham, line 45: I have moved this code over to a Map (think: key / value pair store) which allows us to programmatically create the fields. This removes the need for the "null" case and repeated pattern matching / option handling.

I also added a new helper, `Option.mapBoxed` which basically does `Option.map`, then boxes the result, and then does `Option.toObj`. This is a pattern I've seen in a few places and we could retrospectively apply this as a tech debt task.

cmarshall10450 (commented):
Thank you for cleaning up the PR and updating the names. I don't think the "enableNoPublicIp" parameter is set quite correctly though.

isaacabraham (changes requested):
Hey. I think this is looking pretty good, there's just a few things more that need resolving. Some of them are a bit tricky for a first resource (vnet and keyvault integration with Farmer) so I'm happy for you to leave them for us if you prefer.

Otherwise I think just a few smallish tasks and this will be good to go!

samples/scripts/databricks.fsx
cmarshall10450: Having both a public and private subnet is required for using Bring Your Own VNet with Databricks. Only having one is not supported as the subnets are used for deploying Network Interfaces into and are attached to the machines in the clusters. I'm not sure if the public subnet is needed if enableNoPublicIp is set to true.

src/Farmer/Builders/Builders.Databricks.fs
cmarshall10450: I tried looking at how a Virtual Machine is linked to a Virtual Network but I couldn't work out how to handle all cases such as the VNet already existing or the VNet being deployed by Farmer. I wasn't totally happy with this part and knew that it didn't handle all of the cases

samples/scripts/databricks.fsx
cmarshall10450: I'm not sure what affect using Default or Microsoft.Keyvaults as the KeySource has. The other 3 values do need to be supplied but the version can default to "latest" if its not supplied, otherwise a guid needs to be give. The version should then probably be in the validations.

src/Farmer/Builders/Builders.Databricks.fs
isaacabraham: Let me have a bash at this - I think there are a few combinations that need to be considered.

src/Farmer/Arm/Databricks.fs
isaacabraham: Doh. Great spot!

isaacabraham: Sorry about that.

cmarshall10450: No worries. I blame Microsoft's naming for the confusion

src/Tests/Databricks.fs
cmarshall10450: This doesn't seem to compile for me. It gives me the error: `Type constraint mismatch. The type 'unit' is not compatible with type 'IBuilder'`

isaacabraham: Yeah, looks like an Isaac special - broke the build.

isaacabraham: It's the `Run()` method - it needs to return the `state` back out if the validation passes. I'll fix it up.

samples/scripts/databricks.fsx
isaacabraham: Ok, [this article](https://docs.microsoft.com/en-gb/azure/databricks/security/secrets/secret-scopes) explains that you can either use DataBricks directly for secret management or KeyVault. I've updated the model so that there's no more enum for the secret type. Either you supply the Key Vault configuration or you don't. If you don't, we set the secret type as Default. If you do, we set it as MicrosoftKeyVault. I've also added the "latest" thing you mentioned - where are the docs on that "feature"?

isaacabraham: Ok I *think* that the encryption is more or less good to go, just need a keyword for databrick encryption. I've rolled back on my thoughts to "implicitly" infer this - there are three states: no encryption at all (is this a good idea?), DataBricks, or KeyVault.

cmarshall10450: I'm not sure that this property refers to the secret scopes. I think it's for encrypting the DBFS root file system, which is a Microsoft managed storage account. It is encrypted by a Microsoft managed key by default but can be encrypted by a customer managed key instead. The details of that can be seen in [this article](https://docs.microsoft.com/en-us/azure/databricks/security/keys/customer-managed-keys-dbfs/)

cmarshall10450: I think I was wrong about using "latest". I must have been thinking of AWS Secrets Manager. It looks like, [from here](https://blog.eldert.net/reference-key-vault-secret-latest-version/), that leaving the version empty gives the latest version instead.

cmarshall10450: I've found some more information on the encryption key properties. It looks like using a secret from a Key Vault requires the "keySource" property to be set to Microsoft.Keyvaults because setting it to Default means that it uses the Microsoft managed key instead. This can be seen [here](https://docs.microsoft.com/en-us/azure/databricks/security/keys/customer-managed-keys-dbfs/cmk-dbfs-azure-cli#disable-customer-managed-keys). With that, there isn't much point allowing any other value than "Microsoft.Keyvaults" for the key source if the key vault, key name or key version are set.

isaacabraham:
@cmarshall10450 Would you like to review the latest version of this - and do some end-to-end testing? I've done a basic test (no BYOV or encryption) and it deploys correctly.

@ninjarobot would you like to check that the vnet integration looks reasonable?

samples/scripts/databricks.fsx
isaacabraham: 1. Ok, so I'll change the `latest` thing to just omit the field.
2. Re: DBFS - ok, that makes sense - I went down the rabbit hole trying to figure that all out. I'll make the naming reflect this.
3. The key source stuff has already been updated to be modelled as you describe - either you will supply the key name for customer-managed (and an optional version) or you'll specify the Databricks (internal managed) option.

Open questions:

* When would you *not* want the `prepareEncryption` field set to true? Or is this that the the case that you don't *have* to encrypt the DBFS stuff?
* Also, how does this interact with the `requireInfrastructureEncryption` in the reference docs?
* There is also a mysterious `storageAccountIdentity` field on the JSON but no real docs as to what it does. Any ideas?

isaacabraham: (I've just pushed with some renames to keywords and also removed the "latest" in favour of just omitting the field entirely.

PR #500, opened 2020-12-16
Add Always On for Function Builder

### LINKED ISSUE #498: Add "always_on" member in functions builder
Currently, when deploying a function app, the "always on" setting defaults to false. Adding this member gives more control on the specifics when deploying a function app.

Iamabdul:
This PR closes #498 

* ...
* ...
* ...

This is my first try at this


Iamabdul:
Hello @isaacabraham, new to the game here, is this what you mean?

isaacabraham:
Looks spot on!

isaacabraham:
Hi @Iamabdul quick question. I'm ready to merge this in but I'm thinking, do we even expose the ability to run Functions in app service (or premium) mode? If so, I would suggest that we consider simply not exposing this keyword at all and simply turn it on whenever someone picks an app service plan model and leave it off if someone leaves it as dynamic / consumption mode.

What do you think?

Iamabdul:
> Hi @Iamabdul quick question. I'm ready to merge this in but I'm thinking, do we even expose the ability to run Functions in app service (or premium) mode? If so, I would suggest that we consider simply not exposing this keyword at all and simply turn it on whenever someone picks an app service plan model and leave it off if someone leaves it as dynamic / consumption mode.
> 
> What do you think?

Hello, I see what you mean. As we want to ensure farmer does most of the work for us as developers, maybe we do want to have it turned on automatically (depending on the plan that we choose). I can also confirm, in the azure portal, that by running the function on the app service or premium plan it provides the always on option and is turned on by default (supporting your suggestion).

isaacabraham:
@Iamabdul hmmm. So are you actually seeing the issue right now with functions generated in Farmer? Currently we only support the consumption plan I believe, which as a result I think means that there's no need to add always on?

Maybe this should be about adding the premium / app service plans and then setting always on as part of that.

Iamabdul:
@isaacabraham  Apologies I didn't mean say "running the function", it was supposed to be "creating the function app manually" in the portal as opposed to running the farmer function. I currently do have a premium / app service plan which is used in my current farmer template, and when the json is generated, I see the "always on" property as false. The function app created by running this farmer template results in the "always on" property as turned off in it's settings when I go to the portal.

isaacabraham:
@Iamabdul How are you setting a functions instance with a premium app service plan? The functions instance doesn't let you change the SKU of the host it runs on?

Iamabdul:
I linked the functions app to an external service plan (Currently the function doesn't have this ability but I created a custom operation based on the webapp's "link_to_unmanaged_service_plan"). I should have mentioned this first, I might be confusing things, sorry!

Iamabdul:
Hello @isaacabraham , I created a pull request [here](https://github.com/CompositionalIT/farmer/issues/524) for the link_to_unmanaged_service_plan custom operation. 👍

isaacabraham:
Ok. So we have a couple of choices here:

1. Expose the `always_on` keyword as a shared keyword between Web App and Functions. If you are linking to an external service plan (which will come in with #527) then it's your responsibility to manually set this keyword yourself.
2. Properly support the two Functions SKUs that aren't yet supported; if you set the SKU to either of those, it will automatically turn it on.

I think that we can (and should) support both of these two scenarios. What do you think?

PR #543, opened 2021-01-23
Update global.json, additional scripts for .NET 5.0.

ninjarobot:
@isaacabraham Farmer seems to build and test well with .NET 5, but needs this change to global.json. This also removes the Newtonsoft.Json.dll from the repo since the `#r "nuget:Farmer"` reference will pull it.


isaacabraham:
Thanks for this - merged

PR #544, opened 2021-01-24
Add in new network libraries

admodras:
This PR closes #

* ...
* ...
* ...


isaacabraham:
@admodras Thanks for putting this in - really appreciate the new network resources (as will @ninjarobot)!

However, we really need to separate these different features into separate PRs in order to deal with them in a manageable fashion - each of them needs to be reviewed separately, documented and tested - plus we need to think about how to integrate them with existing builders etc.

Thanks!

admodras:
Yes - this is going to be split up so they are each in separate PRs and we need to update tests and documentation. Still working through a few issues and unfortunately time hasn't allowed us to quite finish off this PR.  We need to integrate with existing builder moving forward and will see if I can carve out time with @ninjarobot to at least get the first library completed and then i can step through the rest after i have something to work off of.

admodras:
We really should have a separate PR that upgrades to dotnet 5.0 so i don't need to have that as part of this PR. We added it for now because that is what we are using and would be cleaner if Farmer as a whole was running off of dotnet5.0 so i can just focus on the new network features.

isaacabraham:
@admodras there was a long-standing NET 5 PR already. I actually canned it this weekend because it was doing too much and have made a lighter one that will go in today.

isaacabraham:
@admodras Farmer now supports .NET 5. Happy to work with you when you have time to get the other network resources in.

23min:
Seems like `dotnet new farmer` still creates a netcorapp3.1 project (installed the template a few days ago). 
Updating the libs manually to .net5 breaks the Minecraft sample.
This .net churn is just so frustrating. Everything's perpetually broken. Sorry for the frustration ;)

isaacabraham:
@ninjarobot what should we do with this PR - I imagine that some of the features in this PR have ended up in Farmer already by some means or other.

js-mode:
@ninjarobot - you can go ahead a close this PR (I don't have permissions to).  I no longer need it for reference and should be covered by the work from Eric.

PR #546, opened 2021-01-24
Adds initContainer support in container groups

ninjarobot:

* Adds initContainer support in Azure Container Instances (container groups).

### DIFF
--- a/Farmer.sln
+++ b/Farmer.sln
@@ -14,7 +14,7 @@ ProjectSection(SolutionItems) = preProject
 	samples\scripts\appinsights.fsx = samples\scripts\appinsights.fsx
 	samples\scripts\container-registry.fsx = samples\scripts\container-registry.fsx
 	samples\scripts\cosmos.fsx = samples\scripts\cosmos.fsx
-	samples\scripts\loganalytics.fsx=samples\scripts\loganalytics.fsx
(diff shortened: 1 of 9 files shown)


isaacabraham:
Looks good to me. Just out of interest, what exactly does initContainer support do? :-)

isaacabraham:
It can merge. I'm just waiting until 1.4 is closed off - which will realistically just be the diagnostics settings which I've basically completed now. Once that's in and 1.4 is out, we'll pull this in for 1.5.

isaacabraham:
@ninjarobot I've reviewed this and updated in line with the latest commits. Only two questions:

1. Do we *have* to change the version of the container group? I only ask because it's theoretically a potential breaking change from a deployment / ARM generation point of view. It's not a huge issue, it should be fine, but I just want to check.
1. Do we absolutely need to have another builder for the `containerInit` thing? Other options that I'm just thinking of (but these may be completely wrong):
    a. Can we use just a function with optional arguments e.g. `ContainerInit.create("init", "busybox", [ "my"; "cmd"; "args" ])`. This is useful when you need to create resources with just a few arguments and they're virtually all mandatory.
    b. Can we reuse the existing Container Builder? This one probably not.

isaacabraham:
Sounds like a builder is the way to go!

PR #556, opened 2021-02-10
Make namespace UUID algorithm compliant with RFC4122:

### LINKED ISSUE #552: Deterministic Guid Algorithm Incorrect
The algorithm for deterministic guid might be slightly incorrect.  As such, it is incompatible with deterministic guids created with other libraries (e.g. [here](https://github.com/Faithlife/FaithlifeUtility/blob/master/src/Faithlife.Utility/GuidUtility.cs)).

The issue seems to be that the namespace guid is not byte swapped to network order.  [See implementation here](https://github.com/Faithlife/FaithlifeUtility/blob/master/src/Faithlife.Utility/GuidUtility.cs#L94).

If update the algorithm to include the swap bytes on the namespace guid, it makes identical guids and becomes compatible.  I made the change locally and tested it to prove this does fix it.

The fix is to simply call `swapBytes` at [this line](https://github.com/CompositionalIT/farmer/blob/64e415902a8fc883739d9b30383a9cdbf324d532/src/Farmer/Types.fs#L308)

Edit:  I looked up the RFC and indeed, you should be swapping the namespace guid into network order. [See here](https://tools.ietf.org/html/rfc4122#section-4.3).

dunnry:
https://tools.ietf.org/html/rfc4122#section-4.3

This PR closes #552

* ...update DeterministicGuid algorithm

I haven't tested this code against a live subscription.  I haven't actually used Farmer yet, I just happened to see that the algorithm was wrong in reviewing the code.  I don't believe there is any documentation for this to update.

The one thing that might occur is that any GUIDs that were calculated using the older algorithm will now not match.  So, if you are using this with an existing deployment, this might be problematic - I hope that someone can better understand the impact.  However, if you are using this with new deployments, there should be no issue.

I have compared the output of this code with a known C# equivalent (see linked issue) and they now match.


isaacabraham:
> The one thing that might occur is that any GUIDs that were calculated using the older algorithm will now not match. So, if you are using this with an existing deployment, this might be problematic

Mmmm. That's a really good point. I'm not sure what to do about this.

isaacabraham:
And what's the "cost" of not implementing this - by using a non-standard way of generating them?

dunnry:
It's hard for me to know the cost of not using this.  If users only calculate deterministic GUID with Farmer, it's probably fine to ignore this fix.  However, if you are using any other implementation of the standard, it will all break.  I just don't know how other people are using the resource IDs created by Farmer.  I know that a C#, Java, or other algorithm (there are many) will not be able to calculate the same ID.  This will culminate in problems for other code when attempting to find or reference the same resource (or create one that Farmer can use).

I found the problem by trying to use DeteministicGuid with our existing ones, created outside of Farmer.  Of course they were different - hence the issue.

Possible approaches to help:

1. Write additional checks to see if the existing, non-compliant GUIDs are present, and then upgrade them to the compliant spec.
2. Write logic that uses the older algorithm for deployments created by older Farmer versions (assuming you keep that data somewhere).  This is maybe not a great idea because you would forever have two versions of DeterministicGuid.
3. Ignore this PR.  This might only impact a small number of people that will be frustrated when their GUIDs calculated outside of Farmer don't match.

isaacabraham:
@dunnry I've tested it - on the second deployment, you get an error saying that "The role assignment already exists". This isn't *as bad* as I feared - at least ARM won't redeploy. The important thing is to make sure that we capture this and show a nicer error, I think, with some guidance for how to resolve it.

isaacabraham:
@dunnry What do you think about this:

`
"A role assignment defined in this template already exists in Azure, but with a different GUID. If you have recently upgraded to Farmer 1.5, please be aware of a breaking change in the generation of role assignment GUIDs. To resolve this, view the resource group in the Azure portal, remove the existing role assignment from IAM and redeploy your Farmer template."
`

PR #580, opened 2021-03-16
Allow connecting ACI to existing vnet

### LINKED ISSUE #577: ACI NetworkProfileConfig automatically sets up DependsOn for vnet
I'd like to be able to specify vnet and only see it setup the dependency if the vnet is defined in the same deployment.
As it is, if I have vnet already deployed as part of another group the template validation fails with 
> The resource 'Microsoft.Network/virtualNetworks/vnetName' is not defined in the template.

ninjarobot:
This PR closes #577 

* Adds `link_to_vnet` in `networkProfile` for connecting a container group to an existing virtual network.

### DIFF
--- a/src/Farmer/Builders/Builders.ContainerGroups.fs
+++ b/src/Farmer/Builders/Builders.ContainerGroups.fs
@@ -375,17 +375,25 @@ type ContainerNetworkInterfaceConfiguration = { IpConfigs : ContainerNetworkInte
 type NetworkProfileConfig =
     { Name : ResourceName
       ContainerNetworkInterfaceConfigurations : ContainerNetworkInterfaceConfiguration list
-      VirtualNetwork : ResourceName
+      VirtualNetwork : ExternalKind
       Tags: Map<string,string>  }
     interface IBuilder with
         member this.ResourceId = networkProfiles.resourceId this.Name
         member this.BuildResources location = [
             { Name = this.Name
               Location = location
+              Dependencies = [
+                  match this.VirtualNetwork with
+                  | Managed resId -> resId // Only generate dependency if this is managed by Farmer (same template)
+                  | _ -> ()
+              ] |> Set.ofList
               ContainerNetworkInterfaceConfigurations =
-                this.ContainerNetworkInterfaceConfigurations
-                |> List.map (fun ifconfig -> {| IpConfigs = (ifconfig.IpConfigs |> List.map (fun ipConfig -> {| SubnetName = ResourceName ipConfig.Subnet |})) |})
-              VirtualNetwork = this.VirtualNetwork
+                  this.ContainerNetworkInterfaceConfigurations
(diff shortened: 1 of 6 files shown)


isaacabraham:
Hey. I've had a look at this (and made a small improvement for the way network profiles are added to container groups ;-). I think this is something that needs documenting but in general:

There are four ways a resource (let's call it Source) can be linked to another one (let's call it Destination):

* External Managed: Source and Destination are both explictly created in your Farmer app. Source is "linked" to Destination through e.g. a keyword or similar.
* External Unmanaged: Destination is created in your Farmer app. Source is created *outside* of Farmer - it probably already exists in Azure.
* AutoCreate Named: Farmer will magically create Source when it creates the Destination resource. The user has supplied the name for the auto created resource.
* AutoCreate Derived: Farmer will magically create Source when it creates the Destination resource. The name of the auto created resource will be derived based on the attributes of the Destination resource. 

A good example for this is to look at WebApp and how we connect the Service Plan to it (ignore the fact that WebApp uses a "common" type across functions and web apps!):

Your config should contain a reference to the "source" resource (in your case, Network Profile):

```fsharp
NetworkProfile : ResourceRef<ResourceName>
```

1. The user doesn't specify anything, and the Farmer Container Group builder should automatically create the Network Profile (if this is this supported?!). If so, set NetworkProfile to AutoCreate Derived.
2. The user wants to specify the *name* of the Network Profile, but still wants Farmer to automatically create the Profile when creating the Container Group. If so, set it to AutoCreate Named.
3. The user is creating the Profile (using e.g. the NetworkProfileBuilder) and just wants to link it. Therefore, use External Managed.
4. The user already has a Profile and just wants this Container Group to use it. Therefore, use External Unmanaged.

Now, if you *don't* want to offer e.g. 1 and 2 - let's say that Container Groups can't create Network Profiles, they can just link to them - then you can just use ExternalKind instead of ResourceRef I suppose. I've not had to do that yet, but if that's the behaviour that we want, then that seems reasonable to me.

src/Farmer/Builders/Builders.ContainerGroups.fs
```diff
@@ -406,7 +413,12 @@ type NetworkProfileBuilder() =
     member _.AddIpConfigs(state:NetworkProfileConfig, configs) = { state with ContainerNetworkInterfaceConfigurations = state.ContainerNetworkInterfaceConfigurations @ configs }
     /// Sets the virtual network for the profile
     [<CustomOperation "vnet">]
-    member _.VirtualNetwork(state:NetworkProfileConfig, vnet) = { state with VirtualNetwork = ResourceName vnet }
+    member _.VirtualNetwork(state:NetworkProfileConfig, vnet) =
+        { state with VirtualNetwork = Managed(virtualNetworks.resourceId(ResourceName vnet)) }
+    /// Links to an existing vnet.
+    [<CustomOperation "link_to_vnet">]
+    member _.LinkToVirtualNetwork(state:NetworkProfileConfig, vnet) =
```
isaacabraham, line 420: Just passing a name probably won't always be sufficient: if the vnet is in another resource group, ARM won't accept this.

In such a case, you should have an overload to supply a full `resourceId`, which can be created with the resource group of, in this case, the vnet.

PR #587, opened 2021-03-19
Modifications to when net5 is exposed in ARM.

isaacabraham:
@ninjarobot minor changes to app service on net5.


ninjarobot (approved):
Thanks :)

PR #590, opened 2021-03-21
Only publish nuget package on new release tags

ninjarobot:

* Triggers CI on new tags as well.
* *Only* publishes NuGet packages when a new tags is created (such as drafting a GitHub Release)

This is a CI change only.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -2,6 +2,9 @@ trigger:
   paths:
     exclude:
     - 'docs/'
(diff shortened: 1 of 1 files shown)


isaacabraham:
@ninjarobot is it possible to specify *what* tag is required or something like that? Not that it should be required - I can't think what other tags we'd need - but I'm curious :-)

isaacabraham:
It's in :-) Thank you for doing this!

PR #592, opened 2021-03-22
Turn on System Identity by default.

### LINKED ISSUE #591: Turn on System Identity on App Service and Functions by default
We should consider simply turning on identity by default for App Service and Functions. There's (as far as I'm aware) no real cost to doing this, and it would make things easier for the default use case.

We could / should replace this with the ability to turn it on / off with a boolean argument instead.

isaacabraham:
This PR closes #591

* Turn on System Identity by default for Web App, Functions, Container Service and Container Group
* Parameterise system_identity keyword (*breaking change*).
* Add system_identity keyword for Container Service.


ninjarobot (commented):
The code all looks great, but what do you think of specifying the default for resources in the `arm` builder so users can "opt in" on this for everything in a given deployment and it won't just start doing this automatically after a package upgrade? It would save them from having to add it on each resource, but also not change their behavior just by upgrading.

isaacabraham:
I think we could support threading some arbitrary state through all builders by changing the IBuilder interface.

isaacabraham:
@ninjarobot I think given the limits on identities, this PR should be closed - otherwise I can see those identity limits being hit without notice.

PR #598, opened 2021-03-30
Add support for dotnet-isolated runtime.

### LINKED ISSUE #596: Add runtime-option "dotnet-isolated" for Azure Functions
It's now possible to run Azure functions as an [isolated process](https://docs.microsoft.com/en-us/azure/azure-functions/dotnet-isolated-process-guide) and use .NET 5.

To get this to work we needed to make the following changes:

Unrelated code is omitted...

```fsharp
//Builders.Functions.fs

type FunctionsRuntime = DotNet_Isolated | DotNet | Node | Java | Python

interface IBuilder with
    member this.BuildResources location = [
        { AppSettings = [
            "FUNCTIONS_WORKER_RUNTIME", (string this.Runtime).ToLower().Replace("_", "-")
          ] } ]
// 
``` 

If it's ok to add this, I'd be happy to add a PR.

viktorvan:
This PR closes #596

* Add RuntimeKind for DotNet runtime for Azure Functions to enable running as dotnet-isolated.

No docs updated, the runtime options are not currently listed in the docs.

### DIFF
--- a/src/Farmer/Builders/Builders.Functions.fs
+++ b/src/Farmer/Builders/Builders.Functions.fs
@@ -12,7 +12,10 @@ open Farmer.Arm.KeyVault
 open Farmer.Arm.KeyVault.Vaults
 open System
 
-type FunctionsRuntime = DotNet | Node | Java | Python
+type RuntimeKind =
+    | Isolated
+    | InProcess
+type FunctionsRuntime = DotNet of RuntimeKind | Node | Java | Python
 type FunctionsExtensionVersion = V1 | V2 | V3
 
 module private FunctionsConfig =
@@ -142,6 +145,11 @@ type FunctionsConfig =
             | None ->
                 ()
 
+            let functionsRuntime =
+                match this.Runtime with
+                | DotNet Isolated -> "dotnet-isolated"
+                | DotNet InProcess -> "dotnet"
+                | other -> (string other).ToLower()
             { Name = this.Name
               ServicePlan = this.ServicePlanId
               Location = location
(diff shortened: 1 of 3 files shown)


viktorvan:
The default runtime, when no runtime is explicitly set is still "dotnet". In the future maybe it will make sense to make "dotnet-isolated" the default.

src/Farmer/Builders/Builders.Functions.fs
```diff
@@ -12,7 +12,10 @@ open Farmer.Arm.KeyVault
 open Farmer.Arm.KeyVault.Vaults
 open System
 
-type FunctionsRuntime = DotNet | Node | Java | Python
+type RuntimeKind =
+    | Isolated
+    | InProcess
+type FunctionsRuntime = DotNet of RuntimeKind | Node | Java | Python
```
ninjarobot, line 18: Rather than introducing the `RuntimeKind` type that is only specific to the `DotNet` case, it would be easier to use if `DotNetIsolated` is just another case altogether that represents the "dotnet-isolated" option. Then the usage remains simple like

```fsharp
use_runtime DotNet
```
or
```fsharp
use_runtime DotNetIsolated
```

ninjarobot (changes requested):
I recommend just adding the `DotNetIsolated` case to keep the `use_runtime` call simple.

src/Farmer/Builders/Builders.Functions.fs
viktorvan: Yes, that's a good point, `use_runtime DotNetIsolated` feels better than `use_runtime (DotNet Isolated)`. 

Since the RuntimeKind was an idea from @isaacabraham maybe he wants to weigh in as well? Should we just go with a `DotNetIsolated` case?

ninjarobot: Yes, I am thinking from the perspective of the `use_runtime` in the DSL, but I would definitely like to get perspective from @isaacabraham as well.

ninjarobot: @isaacabraham can you please comment on this as to whether or not to break out the `RuntimeKind` for these?

isaacabraham: Yes, I think merging them will be better from a public API consumption point of view.

PR #611, opened 2021-04-23
Adds tutorial for custom output expressions.

ninjarobot:

* Adds a tutorial for building an ARM expression to access resource properties and using it in a custom output.

Tutorial only, no unit test surface changed, no new features for release notes.

### DIFF
--- a/docs/content/tutorials/custom-output.md
+++ b/docs/content/tutorials/custom-output.md
@@ -1,0 +1,54 @@
+---
+title: "Custom Output with ARM Expressions"
+date: 2021-04-23
+draft: false
+---
+
+#### Introduction
+
+Many resources have properties that are only set once the resource is created, such as a public IP's address or an ExpressRoute's circuit service key. It is often helpful to have these as output from the deployment so they are available to any downstream automation tasks.
+
+In this tutorial, you will deploy an ExpressRoute circuit, create a reference to the `serviceKey` property on the newly deployed circuit, and provide that as the ARM deployment output.
+
+#### Define the ExpressRoute circuit to deploy
+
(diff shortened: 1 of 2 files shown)


docs/content/tutorials/custom-output.md
```diff
@@ -0,0 +1,54 @@
...
+ARM templates support expressions that are evaluated when the template is executed by ARM. These have many different capabilities, but in this case, we want to reference a newly deployed resource - the ExpressRoute circuit.
+
+First, you can use the type and name of the resource to create a `ResourceId`. Then, that `ResourceId` can be used to build a `reference` expression and retrieve a property of the resource.
+
+```fsharp
+// Build an ARM resourceId type for the circuit.
+let erId = ResourceId.create(Arm.Network.expressRouteCircuits, er.Name)
+
+// Use that ID to build a reference expression and get a property of the referenced resource.
+let serviceKeyRef = ArmExpression.create ($"reference({erId.ArmExpression.Value}).serviceKey")
```
isaacabraham, line 39: I wonder if we can simplify this using the `+` overload and stuff like that?

isaacabraham: Up until now most of this was considered "contributor" level stuff. If we want to make the user-friendly, we might need to think about the API.

isaacabraham (approved):
Approved this but we should probably schedule something in to review how to do this the "right" way.

docs/content/tutorials/custom-output.md
ninjarobot: I think `ArmExpression.reference` could use just an overload to specify a property path and I wouldn't have to do the interpolated string with a nested expression here.

Any type safe wrappers around those ARM expressions end up being very useful, although it may be a pretty big project to cover all of them.

isaacabraham:
It's also useful as an exercise to see how we can make this API simpler :-)

PR #612, opened 2021-04-26
Support for CORS on Storage Accounts

### LINKED ISSUE #575: Add CORS for Storage Account
Support for enabling CORS in the storageAccount so that linked webApps are able to connect without the need to go into the azure portal

isaacabraham:
This PR closes #575

* Create a new `add_cors_rules` keyword to the Storage builder.
* Support for CORS across all four storage services.

### DIFF
--- a/src/Farmer/Arm/Storage.fs
+++ b/src/Farmer/Arm/Storage.fs
@@ -39,7 +39,7 @@ type StorageAccount =
                             match this.Sku with
                             | GeneralPurpose (V1 (V1Replication.LRS performanceTier))
                             | GeneralPurpose (V2 (V2Replication.LRS performanceTier, _)) ->
-                                performanceTier.ToString()
+                                performanceTier.ArmValue
                             | Files _
                             | BlockBlobs _ ->
                                 "Premium"
@@ -48,13 +48,10 @@ type StorageAccount =
                                 "Standard"
                         let replicationModel =
                             match this.Sku with
-                            | GeneralPurpose (V1 (V1Replication.LRS _)) -> "LRS"
-                            | GeneralPurpose (V2 (V2Replication.LRS _, _)) -> "LRS"
-                            | GeneralPurpose (V1 replication) -> replication.ToString()
-                            | GeneralPurpose (V2 (replication, _)) -> replication.ToString()
-                            | Blobs (replication, _) -> replication.ToString()
-                            | Files replication -> replication.ToString()
-                            | BlockBlobs replication -> replication.ToString()
+                            | GeneralPurpose (V1 replication) -> replication.ReplicationModelDescription
+                            | GeneralPurpose (V2 (replication, _)) -> replication.ReplicationModelDescription
+                            | Blobs (replication, _) -> replication.ReplicationModelDescription
+                            | Files replication | BlockBlobs replication -> replication.ReplicationModelDescription
                         $"{performanceTier}_{replicationModel}"
                     |}
                 kind =
@@ -82,7 +79,7 @@ type StorageAccount =
             this.StaticWebsite
             |> Option.map(fun staticWebsite -> result {
                 let! enableStaticResponse = Deploy.Az.enableStaticWebsite this.Name.ResourceName.Value staticWebsite.IndexPage staticWebsite.ErrorPage
-                printfn "Deploying content of %s folder to $web container for storage account %s" staticWebsite.ContentPath this.Name.ResourceName.Value
+                printfn $"Deploying content of %s{staticWebsite.ContentPath} folder to $web container for storage account %s{this.Name.ResourceName.Value}"
                 let! uploadResponse = Deploy.Az.batchUploadStaticWebsite this.Name.ResourceName.Value staticWebsite.ContentPath
                 return enableStaticResponse + ", " + uploadResponse
             })
@@ -117,10 +114,8 @@ type StorageService =
                                 for rule in this.CorsRules do
                                     {| allowedOrigins = rule.AllowedOrigins.Emit (fun r -> r.AbsoluteUri)
                                        allowedMethods = [
-                                           let head, tail = rule.AllowedMethods
-                                           string head
-                                           for httpMethod in tail do
-                                               string httpMethod
+                                           for httpMethod in rule.AllowedMethods.Value do
+                                               httpMethod.ArmValue
                                        ]
                                        maxAgeInSeconds = rule.MaxAgeInSeconds
                                        exposedHeaders = rule.ExposedHeaders.Emit id
--- a/src/Farmer/Arm/Storage.fs
+++ b/src/Farmer/Arm/Storage.fs
@@ -1,12 +1,12 @@
 [<AutoOpen>]
 module Farmer.Arm.Storage
 
-open System
 open Farmer
 open Farmer.Storage
 
 let storageAccounts = ResourceType ("Microsoft.Storage/storageAccounts", "2019-06-01")
 let containers = ResourceType ("Microsoft.Storage/storageAccounts/blobServices/containers", "2018-03-01-preview")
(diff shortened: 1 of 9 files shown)


src/Farmer/Arm/Storage.fs
```diff
@@ -79,6 +87,49 @@ type StorageAccount =
...
+            this.ResourceType.resourceId (this.StorageAccount.ResourceName/"default")
+        member this.JsonModel =
+            {| this.ResourceType.Create(this.StorageAccount.ResourceName/"default", dependsOn = [ storageAccounts.resourceId this.StorageAccount.ResourceName ]) with
+                properties =
+                    {| cors =
+                        {| corsRules =
+                            [
+                                for rule in this.CorsRules do
+                                    {| allowedOrigins = rule.AllowedOrigins.Emit (fun r -> r.AbsoluteUri)
+                                       allowedMethods = [
```
ninjarobot, line 119: What's the difference in what is done here vs
```fsharp
rule.AllowedMethods |> List.map string
```

src/Farmer/Builders/Builders.ContainerGroups.fs
```diff
@@ -200,15 +200,15 @@ type ContainerGroupBuilder() =
     [<CustomOperation "restart_policy">]
     member _.RestartPolicy(state:ContainerGroupConfig, restartPolicy) = { state with RestartPolicy = restartPolicy }
     member private _.SetIpAddress(state:ContainerGroupConfig, ipAddressType, ports) =
-        { state with
+        { state with
```
ninjarobot, line 203: Some extra whitespace got added here during this cleanup.

src/Farmer/Common.fs
```diff
@@ -2,6 +2,15 @@
 
 open System
 
+type NonEmptyList<'T> =
+    'T * 'T List
+module NonEmptyList =
+    let create l =
+        match l with
+        | [] -> failwith "Not allowed!"
```
ninjarobot, line 10: It might be a little better to give some information about what is not allowed here. If they are familiar with .NET stack traces, they will probably see the exception came from `NonEmptyList.create`, but something like, "Cannot create an empty list" would be a better error.

src/Farmer/Common.fs
```diff
@@ -2,6 +2,15 @@
 
 open System
 
+type NonEmptyList<'T> =
+    'T * 'T List
```
ninjarobot, line 6: Any reason to store the tupled head and tail? I didn't see that used, but maybe missed it.

src/Tests/Storage.fs
```diff
@@ -32,7 +34,7 @@ let tests = testList "Storage Tests" [
                 sku Sku.Premium_LRS
                 enable_data_lake true
             }
-            arm { add_resource account }
+            arm { add_resource account }
```
ninjarobot, line 37: Extra whitespaced ended up here, too.

src/Farmer/Arm/Storage.fs
theprash: I'm commenting for Isaac as we wrote some of this code as a group within our company.

There are a few subtle differences between the two methods mentioned [here](https://stackoverflow.com/questions/7624987/whats-the-difference-between-uri-tostring-and-uri-absoluteuri) but they're seemingly not important for this use case. We generally prefer something explicit over `string`/`ToString()` where you're never sure if you'll get something useful or just the type name. It's easier to know if this code is correct by reading it.

src/Farmer/Common.fs
theprash: This already existed in `Builders.KeyVault.fs` and it was moved here for re-use. It's a way of creating a type like list where the type system enforces that there is always at least one item in the list. So the list `[ 1 ]` would be represented as `1, []`. The usages are directly below in the `NonEmptyList` module.

src/Tests/Storage.fs
```diff
@@ -233,4 +235,33 @@ let tests = testList "Storage Tests" [
...
+    test "Sets CORS correctly" {
+        let account = storageAccount {
+            add_cors_rules [
+                StorageService.Blobs, CorsRule.AllowAll
+                StorageService.Blobs, { CorsRule.AllowAll with AllowedOrigins = Specific [ Uri "https://compositional-it.com" ] }
+                StorageService.Queues, CorsRule.create([ "https://compositional-it.com" ], [ GET ], 15, [ "exposed1"; "exposed2" ], [ "ALLOWED1"; "ALLOWED2" ] )
+            ]
+        }
+
+        let rules = (account |> findCorsResource "blobServices").properties.cors.corsRules
```
Ryan-Palmer-CIT, line 247: It would be a bit clearer if these values had different names instead of shadowing each other, e.g. 'blobRules', 'blobRule1' 'blobRule2' and 'queueRules'

src/Farmer/Common.fs
ninjarobot: Sorry for the confusion. I meant uses of the head and/or tail that have been deconstructed from the list. It appears this is really just an implementation detail of the non-empty list (it must have a head).

The reason I ask is that most of the time spent reviewing this PR was when I was trying to understand why `rule.AllowedMethods` used elsewhere was a tuple of some item and then a list of that item. Looking at the implementation detail here, I can understand the reason was that this thing that is called a list and used similarly to a list is really a tuple. Maybe this is an implementation detail that should be hidden so it can be used as a normal list.

src/Farmer/Arm/Storage.fs
ninjarobot: I agree, using `string` can give unexpected results if you're not careful about the type, but the code was already using `string head` and `string httpMethod`. It might be a good idea to have a static member on the type that does the conversion to string safely and will get a warning if someone adds a DU case to make sure it's implemented explicitly.

```fsharp
static member Format = function
    | DELETE -> "DELETE"
    | GET -> "GET" 
    ...
```

I'm was more confused when initially seeing this code because `rule.AllowedMethods` isn't able to just map like a list and instead has to apply a function to the head and then each item in the tail. I see now that it's because this is a `NonEmptyList`. Maybe a `map` function in the `NonEmptyList` module would help hide this implementation detail.

isaacabraham: Yep, we need to remove the use of string here and replace with a Description / ArmValue member.

src/Farmer/Common.fs
isaacabraham: Sounds reasonable - what's the best way to solve this though?

ninjarobot: Instead of representing it as a tuple of the head and tail, you could wrap the list in a single case DU and make the case private:
```fsharp
type NonEmptyList<'t> = private NonEmptyList of 't List
module NonEmptyList =
    let create = function
        | [] -> invalidArg "list" "Not allowed to create an empty list."
        | lst -> NonEmptyList lst
    let toList (NonEmptyList l) = l
```
Since this is defined in `Common.fs`, separate modules have to use the `NonEmptyList.create` function which will validate input. This doesn't do anything magical, but unless the tuple is used for something, this hides the implementation a little better IMHO.

src/Farmer/Arm/Storage.fs
isaacabraham: I've fixed this up and also removed a few other `ToString()`s as well.

src/Farmer/Common.fs
isaacabraham: Fixed

src/Farmer/Common.fs
isaacabraham: Fixed.

isaacabraham:
I think that this is ready to go - @ninjarobot can you confirm you're ok with the Non Empty List implementation now?

ninjarobot (approved):
The usage of `NonEmptyList` looks much clearer, thank you!

Issue #618, opened 2021-04-29
Bug in ServiceBus Namespace IArmResource Premium.Sku


ericsoler1:
When a premium sku is used an invalid arm template is generated.

```
let sb = serviceBus {
            name "servicebusname"
            add_topics topics
            sku (Sku.Premium MessagingUnits.OneUnit) 
        }
```

results in the invalid arm resource

```
{
      "apiVersion": "2017-04-01",
      "dependsOn": [],
      "location": "westus",
      "name": "servicebusname",
      "sku": {
        "capacity": 1,
        "name": "Premium OneUnit",
        "tier": "Premium OneUnit"
      },
      "tags": {},
      "type": "Microsoft.ServiceBus/namespaces"
    }
```

https://docs.microsoft.com/en-us/azure/templates/microsoft.servicebus/namespaces?tabs=json#sbsku-object

ninjarobot:
Fixed by #620

Issue #604, opened 2021-04-13
Error adding filters for Service Bus  


Akash-Mair:
When I try add filters for a subscription under a topic I get the following error:

```
System.Exception: ERROR: {"error":{"code":"InvalidTemplate","message":"Deployment template validation failed: 'The template resource 'turn' at line '45' and column '9' is not valid: Unable to evaluate template language function 'resourceId': the type 'Microsoft.ServiceBus/namespaces/topics/subscriptions' requires '3' resource name argument(s). Please see https://aka.ms/arm-template-expressions/#resourceid for usage details.. Please see https://aka.ms/arm-template-expressions for usage details.'.","additionalInfo":[{"type":"TemplateViolation","info":{"lineNumber":45,"linePosition":9,"path":"properties.template.resources[2].resources[0]"}}]}}
```

Farmer code:
```fsharp
let topicName = "game" 

let myServiceBus = serviceBus {
    name "fridayprep"
    sku Standard
    add_topics [
        topic { 
            name topicName
            add_subscriptions [
                subscription { 
                    name "playerone"
                    add_sql_filter "turn" "sys.To = '1' OR sys.To = 'All'"
                    }
                subscription { 
                    name "playertwo"
                    // add_sql_filter "turn" "sys.To = '1' OR sys.To = 'All'"
                    }
                subscription { 
                    name "playerthree" 
                    // add_sql_filter "turn" "sys.To = '1' OR sys.To = 'All'"
                    } 
            ]
        }
    ]
}

let deployment = arm {
    location Location.WestEurope
    add_resource myServiceBus
    output "connectionString" myServiceBus.NamespaceDefaultConnectionString
    output "defaultSharedAccessPolicyPrimaryKey" myServiceBus.DefaultSharedAccessPolicyPrimaryKey
}
```

Generated ARM template:

```json
{
  "$schema": "https://schema.management.azure.com/schemas/2019-04-01/deploymentTemplate.json#",
  "contentVersion": "1.0.0.0",
  "outputs": {
    "connectionString": {
      "type": "string",
      "value": "[listkeys(resourceId('Microsoft.ServiceBus/namespaces/authorizationRules', 'fridayprep', 'RootManageSharedAccessKey'), '2017-04-01').primaryConnectionString]"
    },
    "defaultSharedAccessPolicyPrimaryKey": {
      "type": "string",
[...]

PR #620, opened 2021-04-29
Ersoler/service bus

### LINKED ISSUE #619: Bug in ServiceBusNamespace/Topics/Subscriptions/Rules dependsOn clause generation
Rules generate an invalid arm template expression for its depends on clause

```
"dependsOn": [
            "[resourceId('Microsoft.ServiceBus/namespaces/topics/subscriptions', 'subscriptionName')]"
          ],
```
the namespace name and topic name must also be provided when using this template expression.

ericsoler1:
This PR closes #619 #618  #604

* ServiceBus Subscription IArmResource JsonModel now uses a relative depends on clause for the Rule resources
* ServiceBus Namespace IArmResource JsonModel now emits the correct value for Sku for the Premium case

* [x ] **Tested my code** end-to-end against a live Azure subscription.


ninjarobot (approved):
Thanks for fixing these issues and adding the test coverage over them.

src/Farmer/Arm/ServiceBus.fs
```diff
@@ -44,7 +44,7 @@ module Namespaces =
                          |}
                         resources = [
                          for rule in this.Rules do
-                            {| rules.Create(rule.Name, dependsOn = [ subscriptions.resourceId this.Name ]) with
+                            {| rules.Create(rule.Name, dependsOn = [ ResourceId.create (ResourceType("", ""), this.Name)]) with
```
ninjarobot, line 47: @isaacabraham this is a new scenario where we shouldn't implement the type-qualified resource Id in the `dependOn`. These rules are resources that are nested in the subscription subresource, so they just reference it by name. This fixes the regression itself, but maybe we should use a different type for `dependsOn` so that it can be both a `ResourceId` and just a name.

isaacabraham: Can't you fully-qualify a subscription-level resource? As far as I know we don't do any more "nesting" of resources, everything is flat nowadays.

src/Farmer/Arm/ServiceBus.fs
```diff
@@ -109,6 +109,14 @@ module Namespaces =
                            enablePartitioning = this.EnablePartitioning |> Option.toNullable |}
                 |} :> _
 
+module private Sku =
```
isaacabraham, line 112: I would suggest making this a member on the `Sku` type (consistency).

src/Farmer/Arm/ServiceBus.fs
isaacabraham: And if not, we should change the ARM Writer for service bus to conform to that (if possible).

src/Farmer/Arm/ServiceBus.fs
ericsoler1: I think I fixed this here. https://github.com/CompositionalIT/farmer/pull/625

PR #584, opened 2021-03-19
Add support for webApp and functionsApp deployment slots

TheRSP:
This PR closes #198

The changes in this PR are as follows:

* Add slot builder `appSlot`
* Add `add_slot` to functions and webApp
* Add `zip_deploy_slot` to functions and webApp

I have read the [contributing guidelines](CONTRIBUTING.md) and have completed the following:

* [x] **Tested my code** end-to-end against a live Azure subscription.
* [x] **Updated the documentation** in the docs folder for the affected changes.
* [x] **Written unit tests** against the modified code that I have made.
* [ ] **Updated the [release notes](RELEASE_NOTES.md)** with a new entry for this PR.
* [ ] **Checked the coding standards** outlined in the [contributions guide](CONTRIBUTING.md) and ensured my code adheres to them.

If I haven't completed any of the tasks above, I include the reasons why here:

PR #629, opened 2021-05-13
New slot support

### LINKED ISSUE #198: Add support for web app slots
https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2018-02-01/sites/slots

Would also be useful to be able to make app settings "slot sticky".  I can't find any official documentation or reference but I've been using the following in my ARM templates.

```
{
  "name": "[variables('webAppName')]",
  "type": "Microsoft.Web/sites",
  "apiVersion": "2016-08-01",
  "location": "[resourceGroup().location]",
  "properties": {
    "serverFarmId": "[variables('appServiceResourceId')]",
    "siteConfig": {
      "connectionStrings": [
        {
          "name": "Database",
          "connectionString": "[variables('sqlConnectionString')]",
          "type": "SQLAzure"
        },
        {
          "name": "ServiceBus",
          "connectionString": "[variables('serviceBusConnectionString')]",
          "type": "Custom"
        }
      ],
      "appSettings": [
        {
          "name": "ASPNETCORE_ENVIRONMENT",
          "value": "[parameters('aspNetEnvironment')]"
        },
        {
          "name": "AllowedOrigins",
          "value": "[variables('allowedOrigins')]"
        }
      ]
    }
  },
  "resources": [
    {
      "type": "config",
      "name": "slotconfignames",
      "apiVersion": "2016-08-01",
      "properties": {
        "connectionStringNames": [
          "Database",
          "ServiceBus"
        ],
        "appSettingNames": [
          "ASPNETCORE_ENVIRONMENT",
          "AllowedOrigins"
        ]
      },
      "dependsOn": [
        "[variables('webAppName')]"
      ]
    }
  ]
}
```

KieranChandler:
This PR closes #198
* Add slot builder `appSlot`
* Add `add_slot` to functions and webApp
* Add `zip_deploy_slot` to functions and webApp


viktorvan:
Have you tested to deploy this to azure multiple times and verified that the production slot does not go down during the deployment. I have previously made some comments on the previous PR #584.

What I found happens is that the webapp builder add some “default” app settings to the webapp (which is in the production slot by default). And when the ARM template is deployed it will overwrite and remove all settings that has previously been swapped into the production slot. I.e. the production webapp will break. 

So I think one solution is to make sure that the web-app does not have any appsettings when the arm-template is deployed. But maybe there are better ways to address this issue.

viktorvan:
I am not sure this is actually a bug. Maybe it can be solved by using “slot settings” for the settings that the web builder adds by default?

“To configure an app setting or connection string to stick to a specific slot (not swapped), go to the Configuration page for that slot. Add or edit a setting, and then select deployment slot setting. Selecting this check box tells App Service that the setting is not swappable.” https://docs.microsoft.com/en-us/azure/app-service/deploy-staging-slots#automate-with-arm-templates

That sounds like a better solution than my hack of clearing the settings that I posted in another issue discussion. 

Would there be any consequences to making them “slot settings” even for non slotted scenarios? Or do we need to make conditionally slot settings when you add slots in the web builder?

viktorvan:
It is a bit complicated to test, perhaps. But I agree that we need to have a test in place so we can check that the implementation works.

Pre-requisites
- A simple webapp with an endpoint that reads and validates value from appSettings and then returns 200.
- The webapp needs to be deployed and running in the production slot.

Test
During the steps below the test needs to poll the endpoint above of the production slot webapp with a short timeout and assert that it is up and returns 200 all the time during steps 1-3 below.

1. Run arm-deployment
2. Deploy updated webapp to staging slot
3. Swap slots

I expect the downtime to be neglible in this scenario when slotted deployment works.

As a verification of the test we should run this without slotted deployment and then we should be able to detect some downtime during the deployment.

isaacabraham:
The CI process already has a single test that does a basic ARM deployment into a subscription and then destroys it - see `endToEndTests` - so we could theoretically add that in as an integration test, if required.

On a side note - are you using deployment slots only as a way to have near-zero downtime deployments? Using ZipDeploy with Run From Package on enables a similar experience but without the complexity of deployment slots.

isaacabraham:
> What I found happens is that the webapp builder add some “default” app settings to the webapp (which is in the production slot by default). And when the ARM template is deployed it will overwrite and remove all settings that has previously been swapped into the production slot. I.e. the production webapp will break.

> So I think one solution is to make sure that the web-app does not have any appsettings when the arm-template is deployed. But maybe there are better ways to address this issue.

Mmmm. That needs a bit of investigating I think - this isn't really a Farmer thing but just making sure there's a clear understanding of "how do settings work in a multi-slot environment with ARM" - once that's done we can define what the behaviour for Farmer should be.

viktorvan:
Yes, I agree some investigation is needed, I think slot settings may be key here. But realistically I do not think I have any time to spare to investigate this the upcoming week.

Interesting side note @isaacabraham, we are using slots just for zero downtime. Our process is: 
1. deploy webapp to staging slot. 
2. Run warmup on staging slot.
3. Swap staging slot with production slot.

Can that be simplified with zip-deploy / Run From Package? I need to investigate this as well I think 😁

isaacabraham:
If you use [run from package](https://docs.microsoft.com/en-us/azure/azure-functions/run-functions-from-deployment-package), it avoids much of that. I'm not sure exactly *how* good at warm-start it is, but it's a much quicker and simpler deployment model that means you don't need to have two instances of your site, and it avoids the potential "file in use" that you get with standard zip deploy. As soon as you deploy the new version of your app (as a zip), the app service starts using that instead.

The only real limitation with run from package is that you can't write any files to the local FS (because the app is run from the zip).

isaacabraham:
@viktorvan of course this is already supported in Farmer, just set the `run_from_package` flag ;-)

viktorvan:
For me the next step would be to run the tests I outlined above:
>Test
>During the steps below the test needs to poll the endpoint above of the production slot webapp with a short timeout and assert that it is up and returns 200 all the time during steps 1-3 below.

>1. Run arm-deployment
>2. Deploy updated webapp to staging slot
>3. Swap slots

>I expect the downtime to be neglible in this scenario when slotted deployment works.

I haven't had any time to look at this, but today I am starting my summer vacation, yay! So I think I will finally have some time for this next week.

KieranChandler:
We've been using slot deployment for zero-downtime for the past couple of months and noticed some issues with the appsettings as you mentioned and have made some fixes. I'm currently working on getting them merged into this branch and getting the tests written

KieranChandler:
@ninjarobot Yes, definitely. I'm working on it and the end to end testing at the moment and hopefully will finish tomorrow

ninjarobot (approved):

Issue #405, opened 2020-10-13
Support for Azure Monitor logic apps in Farmer


MoeJw:
Sometimes we need To get richer debugging information about logic apps during runtime, we can set up and use Azure Monitor logs to record and store information about runtime data and events, such as trigger events, run events, and action events in a Log Analytics workspace and get something like this

![image](https://user-images.githubusercontent.com/42002662/95859700-ae2e6200-0d67-11eb-830d-e3cb861a533a.png)

To do this in the Azure portal  we have to add the following:
 1- Log Analytics workspace
 2- Add apps Management solution in the  Log Analytics workspace for existing logic apps.
 3-enable diagnostics settings for an existing logic app.

![image](https://user-images.githubusercontent.com/42002662/95861660-6f4ddb80-0d6a-11eb-83f2-20ce847def4a.png)

It's tedious to do all of this in the Azure portal. It is so much easier to do it in Farmer.
I already added Log Analytics workspace to Farmer just waiting for @isaacabraham  to review the last changes  and I want to also to add   apps Management solution and diagnostics settings

isaacabraham:
Sounds good :-) I've been very busy the last week but will address the PR you've added as soon as I can.

ninjarobot:
Fixed by #920

PR #632, opened 2021-05-18
Add LogAnalytics Operations Management Solution resource

MoeJw:
This PR closes [405](https://github.com/CompositionalIT/farmer/issues/405)

* add solution resource

### DIFF
--- a/src/Farmer/Arm/Solution.fs
+++ b/src/Farmer/Arm/Solution.fs
@@ -1,0 +1,41 @@
+﻿[<AutoOpen>]
+module Farmer.Arm.Solution
+
+open Farmer
+
+let solution = ResourceType("Microsoft.OperationsManagement/solutions", "2015-11-01-preview")
+
+type Solution =
+    { Name : ResourceName
+      Location : Location
+      WorkspaceResourceId : ResourceId
+      ContainedResources : ResourceId Set
+      Publisher : string
+      PromotionCode : string option
+      ReferencedResources : ResourceId Set
+      Product : string 
+      Dependencies : ResourceId Set
+      Tags : Map<string,string> }
+      
+
+    interface IArmResource with
+        member  this.ResourceId = solution.resourceId this.Name
+        member  this.JsonModel =
+            {| solution.Create(this.Name, this.Location,this.Dependencies, tags = this.Tags ) with
+                plan = 
+                    {| 
+                       name = this.Name.Value
+                       publisher = this.Publisher
+                       promotionCode = this.PromotionCode |> Option.toObj
+                       product = this.Product
+                    
+                    |}
+                properties = 
+                    {| workspaceResourceId = this.WorkspaceResourceId.ArmExpression.Value
+                       containedResources = this.ContainedResources
+                       referencedResources = this.ReferencedResources |}
+            |} :> _
+
+
+
+
--- a/src/Farmer/Builders/Builders.Solution.fs
+++ b/src/Farmer/Builders/Builders.Solution.fs
@@ -1,0 +1,79 @@
+﻿[<AutoOpen>]
+module Farmer.Builders.Soltuion
+
+open Farmer
+open Farmer.Arm.Solution
+type SolutionConfig = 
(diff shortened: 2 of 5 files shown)


MoeJw:
Hey @ninjarobot I closed the last PR, because I didn't fetch the last changes to my fork.
I created new one and added my changes to it.
I didn't create tests for it because there is no defined solution type in the Microsoft.Azure.Management.OperationalInsights.Models.

MoeJw:
I'm ready for any changes you want to do

src/Farmer/Builders/Builders.Solution.fs
```diff
@@ -0,0 +1,79 @@
+﻿[<AutoOpen>]
+module Farmer.Builders.Soltuion
```
ninjarobot, line 2: `Solution` is misspelled here (`Soltuion`).

src/Farmer/Arm/Solution.fs
```diff
@@ -0,0 +1,41 @@
+﻿[<AutoOpen>]
+module Farmer.Arm.Solution
+
+open Farmer
+
+let solution = ResourceType("Microsoft.OperationsManagement/solutions", "2015-11-01-preview")
+
+type Solution =
```
ninjarobot, line 8: Since `Solution` often refers to an MSBuild solution file in the .NET world, it would be best to call this module something that distinguishes it, like `OperationsManagementSolution`.

src/Farmer/Builders/Builders.Solution.fs
```diff
@@ -0,0 +1,79 @@
...
+        | {Product = ""} ->  failwith "You must specify a product."
+        | _ -> state
+
+    [<CustomOperation "name">]
+    member _.Name(state: SolutionConfig, name) = { state with Name = ResourceName name }
+
+    [<CustomOperation "workspace">]
+    member _.WorkspaceResourceId(state: SolutionConfig, workspace:WorkspaceConfig) = { state with WorkspaceResourceId  = (workspace :> IBuilder).ResourceId}
+
+    [<CustomOperation "containedResources">]
```
ninjarobot, line 58: We use a `snake_case` convention for the custom operations in the Farmer DSL, such as the operations in the [Log Analytics builder](https://compositionalit.github.io/farmer/api-overview/resources/loganalytics/). Can you please change these from `camelCase` to `snake_case`? For example, this would be `contained_resources`.

ninjarobot (changes requested):
Thanks for the contribution @MoeJw - I made a few recommendations for adjustments to follow conventions and also avoid confusion with a .NET solution file. Can you please make those changes and add test coverage to ensure the DSL sets the properties you expect on the configuration record?

isaacabraham:
@MoeJw how does this align with the existing Log Analytics builder that Farmer has - can we incorporate this into that (this is just my lack of knowledge of how OMS works in relation to the existing LA builder - so feel free to tell me "no, they are completely different things" :-))?

And, if not, I would like to pick a namespace (and builder) more descriptive than simply `solution`, which is not only a little too generic but has other meaning in the .NET world as well. Perhaps `OMS` or similar?

MoeJw:
>  Can you please make those changes and add test coverage to ensure the DSL sets the properties you expect on the configuration record?


@ninjarobot  Sorry I didn't create Unit Test for this because I couldn't find any class about it in Microsoft.Azure.Management

MoeJw:
@isaacabraham OMS will allow you to have a single point of management for your hybrid, multi-cloud and on premises environments by allowing to visualize he most important data out of it.

Log Analytics Collect, store and analyze log data from virtually any source 

image
 So as you can see here oms consider the log analytics as data source the collects the data and provide it to the oms to visualize and manage, Therefore OMS can't exists  without log analytics

MoeJw:
Yeah I think OMS is better than solution.

isaacabraham:
@MoeJw There are some examples of unit tests where there is no Azure SDK; in such cases, we simply test the emitted JSON directly. The Databricks one is an example of that.

PR #633, opened 2021-05-20
Adds support for load balancers.

ninjarobot:

* Adds support for [Azure Load Balancers](https://docs.microsoft.com/en-us/azure/load-balancer/load-balancer-overview)

### DIFF
--- a/docs/content/api-overview/resources/load-balancer.md
+++ b/docs/content/api-overview/resources/load-balancer.md
@@ -1,0 +1,120 @@
+---
+title: "Load Balancer"
+date: 2021-05-20T13:04:00-04:00
+chapter: false
+weight: 21
+---
+
+#### Overview
+The Load Balancer builder (`loadBalancer`) creates load balancers that can distribute load amongst healthy services in a backend pool on public or private networks.
+
+* Load balancers (`Microsoft.Network/loadBalancers`)
+* Load balancer frontend IP configurations (`Microsoft.Network/loadBalancers/frontendIPConfigurations`)
+* Load balancer backend address pools (`Microsoft.Network/loadBalancers/backendAddressPools`)
+* Load balancer health probes (`Microsoft.Network/loadBalancers/probes`)
+
+#### Builder Keywords
+
+| Applies To | Keyword | Purpose |
+|-|-|-|
+| loadBalancer | name | Specifies the name of the load balancer |
+| loadBalancer | sku | Specifies the SKU of the load balancer - default is 'Basic'. |
+| loadBalancer | tier | Specifies the tier of the load balancer - default is 'Regional'. |
+| loadBalancer | add_frontends | Adds frontend IP configurations as defined by the `frontend` builder. |
+| loadBalancer | add_backend_pools | Adds backend address pool configurations as defined by the `backendAddressPool` builder. |
+| loadBalancer | add_rules | Adds load balancing rules as defined by the `loadBalancingRule` builder. |
+| loadBalancer | add_probes | Adds probes for the address pool as defined by the `loadBalancerProbe` builder. |
+| loadBalancer | add_dependencies | Adds the resource ID's of additional dependencies that must be provisioned before the load balancer. |
+| frontend | name | Name of the frontend IP configuration. |
+| frontend | private_ip_allocation_method | Specifies how the private IP is assigned on an internal load balancer. |
+| frontend | public_ip | Specifies the name of a public IP to generate. It will be generated with the same SKU as the load balancer. |
+| frontend | link_to_public_ip | The name of an existing public IP to link to. |
+| backendAddressPool | name | The name of the backend address pool. |
+| backendAddressPool | load_balancer | The name of a load balancer these should be added to (used when adding to a pool for an existing load balancer). |
+| backendAddressPool | link_to_vnet | Specifies an existing virtual network where the backend services are connected. |
+| backendAddressPool | add_ip_addresses | Adds IP addresses to the backend pool. |
+| loadBalancerProbe | name | The name of the load balancer probe. |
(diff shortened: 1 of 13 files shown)


docs/content/api-overview/resources/load-balancer.md
```diff
@@ -0,0 +1,120 @@
...
+                    frontend_ip_config "lb-frontend"
+                    backend_address_pool "lb-backend"
+                    frontend_port 80
+                    backend_port 8080
+                    protocol TransmissionProtocol.TCP
+                    probe "httpGet"
+                }
+            ]
+            add_dependencies [
+                Farmer.Arm.Network.virtualNetworks.resourceId "my-vnet"
```
isaacabraham, line 115: This shouldn't be required I think - should just be able to take in the VNet as an IBuilder and work from that?

src/Farmer/Arm/Network.fs
```diff
@@ -113,8 +113,8 @@ type VirtualNetworkGateway =
                                    properties =
                                     let allocationMethod, ip =
                                         match ipConfig.PrivateIpAllocationMethod with
-                                        | DynamicPrivateIp -> "Dynamic", null
-                                        | StaticPrivateIp ip -> "Static", string ip
+                                        | PrivateIpAddress.DynamicPrivateIp -> "Dynamic", null
```
isaacabraham, line 116: Since these are fully qualified now, perhaps the two values / cases should just be `Dynamic` and `Static` i.e. `PrivateIpAddress.Dynamic` etc.

src/Tests/JsonRegression.fs
```diff
@@ -201,4 +201,66 @@ let tests =
...
+                    add_frontends [
+                        frontend {
+                            name "lb-frontend"
+                            public_ip "lb-pip"
+                        }
+                    ]
+                    add_backend_pools [
+                        backendAddressPool {
+                            name "lb-backend"
+                            link_to_vnet "my-vnet"
```
isaacabraham, line 233: This should be overloaded to take in a VNet directly.

src/Tests/JsonRegression.fs
```diff
@@ -201,4 +201,66 @@ let tests =
...
+                            load_balancer "lb"
```
isaacabraham, line 234: We shouldn't (ideally) need to specific the name of the load balancer - it's contained by the parent anyway.

src/Tests/JsonRegression.fs
```diff
@@ -201,4 +201,66 @@ let tests =
...
+                            frontend_ip_config "lb-frontend"
+                            backend_address_pool "lb-backend"
+                            frontend_port 80
+                            backend_port 8080
+                            protocol TransmissionProtocol.TCP
+                            probe "httpGet"
+                        }
+                    ]
+                    add_dependencies [
+                        Farmer.Arm.Network.virtualNetworks.resourceId "my-vnet"
```
isaacabraham, line 261: Same here.

src/Tests/LoadBalancer.fs
```diff
@@ -0,0 +1,167 @@
...
+    test "Complete load balancer" {
+        let found =
+            arm { add_resource completeLoadBalancer }
+                |> findAzureResources<Microsoft.Azure.Management.Network.Models.LoadBalancer> client.SerializationSettings
+                |> List.tryFind (fun r -> r.Name = "lb")
+        let resource = Expect.wantSome found "No 'lb' resource found in template."
+        Expect.hasLength resource.BackendAddressPools 1 "Incorrect number of backend address pools"
+        Expect.equal resource.BackendAddressPools.[0].Name "lb-backend" "Incorrect name for backend address pool"
+        Expect.hasLength resource.Probes 1 "Incorrect number of probes"
+        Expect.equal resource.Probes.[0].Name "httpGet" "Incorrect name for httpGet probe"
```
isaacabraham, line 143: For readability, consider an assignment:

```fsharp
let probe = resources.Probes.[0]
```

src/Tests/LoadBalancer.fs
```diff
@@ -0,0 +1,167 @@
...
+        Expect.equal resource.Probes.[0].Protocol "Http" "Incorrect protocol for httpGet probe"
+        Expect.equal resource.Probes.[0].Protocol "Http" "Incorrect protocol for httpGet probe"
+        Expect.equal resource.Probes.[0].RequestPath "/" "Incorrect request path for httpGet probe"
+        Expect.equal resource.Probes.[0].Port 8080 "Incorrect port for httpGet probe"
+        Expect.equal resource.Probes.[0].IntervalInSeconds (Nullable 15) "Incorrect interval for httpGet probe"
+        Expect.equal resource.Probes.[0].NumberOfProbes (Nullable 2) "Incorrect number of probes for httpGet probe"
+        Expect.hasLength resource.LoadBalancingRules 1 "Incorrect number of load balancing rules"
+        Expect.equal resource.LoadBalancingRules.[0].Name "rule1" "Incorrect name for rule"
```
isaacabraham, line 151: Same here for the LB rule.

isaacabraham (changes requested):
Looks good to me. I would just like to see either overloads and the samples updated to use the previously defined values to tie up the relationships between resources, and where possibly to omit them completely.

src/Farmer/Arm/Network.fs
ninjarobot: So it doesn't break existing code, I adjusted it not to be fully qualified. Thanks for catching this.

docs/content/api-overview/resources/load-balancer.md
ninjarobot: I fixed this so it will work based on what is passed in the builder so this is not required. I left `add_dependencies` because you may still need this depending on what else may be required for ARM to deploy first that wouldn't be explicitly defined in the load balancer or backend pool.

Issue #168, opened 2020-05-21
Support for Traffic Manager


evilpilaf:

isaacabraham:
@evilpilaf Good idea. Can you provide a sample template as a basis to develop the implementation? 

Here's the ARM resource spec: https://docs.microsoft.com/en-us/azure/templates/microsoft.network/2018-04-01/trafficmanagerprofiles

evilpilaf:
A sample template could be as such:

```
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "profileName": {
            "defaultValue": "<ProfileName>",
            "type": "String"
        },
        "endpointOne": {
            "defaultValue": "/subscriptions/<GUID>/resourceGroups/endpointOneResourceGroupName/providers/Microsoft.Web/sites/<endpointOne>",
            "type": "String"
        }
		// ...
    },
    "variables": {},
    "resources": [
        {
            "type": "Microsoft.Network/trafficManagerProfiles",
            "apiVersion": "2018-04-01",
            "name": "[parameters('profileName')]",
            "location": "global",
            "properties": {
                "profileStatus": "Enabled",
                "trafficRoutingMethod": "Performance",
                "dnsConfig": {
                    "relativeName": "[parameters('profileName')]",
                    "ttl": 60
                },
                "monitorConfig": {
                    "profileMonitorStatus": "CheckingEndpoints",
                    "protocol": "HTTP",
                    "port": 80,
                    "path": "/",
                    "intervalInSeconds": 30,
                    "toleratedNumberOfFailures": 3,
                    "timeoutInSeconds": 10
                },
                "endpoints": [
                    {
                        "id": "[concat(resourceId('Microsoft.Network/trafficManagerProfiles', parameters('profileName')), '/azureEndpoints/<FirstEndpointName>')]",
                        "name": "east",
                        "type": "Microsoft.Network/trafficManagerProfiles/azureEndpoints",
                        "properties": {
                            "endpointStatus": "Enabled",
                            "endpointMonitorStatus": "CheckingEndpoint",
                            "targetResourceId": "[parameters('endpointOne')]",
                            "target": "<TargetUrl>",
                            "weight": "<Weigth>",
                            "priority": "<Priority>",
                            "endpointLocation": "<TargetLocation>"
                        }
                    },
					// ...
                ],
[...]

PR #635, opened 2021-05-25
Traffic Manager support

BrianVallelunga:
This PR references https://github.com/CompositionalIT/farmer/issues/168

This is not ready to deploy, just the start of support for Traffic Manager. It's a feature my project currently uses and is required for me to move from my current ARM template to Farmer.

I have not had experience contributing a new resource before, nor building a computation expression, but the existing ones have given me a good starting point as examples. I have some basic JSON being generated already, so that's a good start.

My first attempt at this will focus on providing support for basic Traffic Manager with simple endpoints and monitoring configuration.

Work in progress.

### DIFF
--- a/src/Farmer/Arm/TrafficManager.fs
+++ b/src/Farmer/Arm/TrafficManager.fs
@@ -14,19 +14,19 @@ type Endpoint =
       Target : EndpointTarget
       Weight: int
       Priority: int
-      Location: Location }
+      Location: Location option }
     interface IArmResource with
         member this.ResourceId = azureEndpoints.resourceId (this.Name)
         member this.JsonModel =
-            {| location = this.Location
-               name = this.Name.Value
+            {| name = this.Name.Value
                properties =
                 {| endpointStatus = this.Status.ArmValue
                    weight = this.Weight
                    priority = this.Priority
-                   endpointLocation = this.Location.ArmValue
+                   endpointLocation = this.Location |> Option.map (fun l -> l.ArmValue)
+                                                    |> Option.defaultValue null
                    targetResourceId = match this.Target with
-                                      | ExternalDomain _ -> null
+                                      | External _ -> null
                                       | Website resourceName -> sites.resourceId(resourceName).Eval()
                    target = this.Target.ArmValue |} |} :> _
 
@@ -44,8 +44,6 @@ type Profile =
         member this.ResourceId = profiles.resourceId (this.Name)
         member this.JsonModel =
             {| profiles.Create(this.Name, Location.Global, this.Dependencies, tags = this.Tags) with
-                   name = this.Name.Value
-                   tags = this.Tags
                    properties =
                        {| profileStatus = this.Status.ArmValue
                           trafficRoutingMethod = this.RoutingMethod.ArmValue
--- a/src/Farmer/Arm/TrafficManager.fs
+++ b/src/Farmer/Arm/TrafficManager.fs
@@ -7,25 +7,6 @@ let profiles = ResourceType ("Microsoft.Network/trafficManagerProfiles", "2018-0
 let endpoints = ResourceType ("Microsoft.Network/trafficManagerProfiles/azureEndpoints", "2018-04-01")
 let externalEndpoints = ResourceType ("Microsoft.Network/trafficManagerProfiles/externalEndpoints", "2018-04-01")
 
-type RoutingMethod =
-    | Performance
-    | Weighted
-    | Priority
-    | Geographic
-    | Subnet
-
-type MonitorProtocol =
-    | Http
-    | Https
-
-type MonitorConfig =
-    { Protocol : MonitorProtocol
-      Port: int
-      Path: string
-      IntervalInSeconds: int
-      ToleratedNumberOfFailures: int
-      TimeoutInSeconds: int }
-
 type Endpoint =
     { Name : ResourceName
       Status: FeatureFlag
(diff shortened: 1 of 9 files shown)


BrianVallelunga:
I'm making good progress, but I just realized my fork was 800 commits behind master, so I have some fix-up work to do. I may have to force push a rebase and open a new PR.

src/Farmer/Builders/Builders.TrafficManager.fs
```diff
@@ -0,0 +1,199 @@
...
+        { state with
+            Tags = pairs |> List.fold (fun map (key,value) -> Map.add key value map) state.Tags }
+
+    /// Adds a tag to the Traffic Manager profile
+    [<CustomOperation "add_tag">]
+    member this.Tag(state:TrafficManagerConfig, key, value) = this.Tags(state, [ (key,value) ])
+
+    /// Adds an Endpoint to the Traffic Manager profile
+    [<CustomOperation "add_endpoint">]
+    member __.AddEndpoint(state:TrafficManagerConfig, endpoint:EndpointConfig) =
```
ninjarobot, line 138: Most of the builders let you add to a list all at once, so this could be 
```fsharp
[<CustomOperation "add_endpoints">]
member _.AddEndpoints(state:TrafficManagerConfig, endpoints:EndpointConfig list) =
    { state with Endpoints = state.Endpoints @ endpoints }
```

src/Farmer/Builders/Builders.TrafficManager.fs
```diff
@@ -0,0 +1,199 @@
...
+        { state with Endpoints = endpoint :: state.Endpoints }
+
+    /// Sets the DNS TTL of the Traffic Manager profile, in seconds (default 30)
+    [<CustomOperation "dns_ttl">]
+    member __.DnsTtl(state:TrafficManagerConfig, ttl) = { state with DnsTtl = ttl }
```
ninjarobot, line 143: Even if this takes an integer, please provide an overload that can take a TimeSpan since that can be a little clearer.

src/Farmer/Arm/TrafficManager.fs
```diff
@@ -0,0 +1,61 @@
...
+                   endpointLocation = this.Location |> Option.map (fun l -> l.ArmValue)
+                                                    |> Option.defaultValue null
+                   targetResourceId = match this.Target with
+                                      | External _ -> null
+                                      | Website resourceName -> sites.resourceId(resourceName).Eval()
+                   target = this.Target.ArmValue |} |} :> _
+
+type Profile =
+    { Name : ResourceName
+      DnsTtl: int
```
ninjarobot, line 35: This needs either a unit of measure or to be a `System.TimeSpan` so it's clear what the unit is for the TTL.

src/Farmer/Builders/Builders.TrafficManager.fs
BrianVallelunga: I'll change it to `add_endpoints` and add an overload that also just takes a single Endpoint.

src/Farmer/Builders/Builders.TrafficManager.fs
```diff
@@ -0,0 +1,199 @@
...
+              Priority = this.Priority
+              Location = match this.Target with
+                         | External (_, l) -> Some l
+                         |_ -> None
+              }
+        ]
+
+type TrafficManagerConfig =
+    { Name : ResourceName
+      DnsTtl : int
```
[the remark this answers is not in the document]
BrianVallelunga: I'm going to go ahead and make all of the "Seconds" fields use measures. This is probably a best practices for all of Farmer.

src/Farmer/Arm/TrafficManager.fs
BrianVallelunga: Done. See above comments about units and TimeSpans.

src/Farmer/Builders/Builders.TrafficManager.fs
ninjarobot: Many do, many don't, as it's evolved over time, but when enhancements come in, we ask that contributors use type safety. They were recently added on [service bus TTL's](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Builders/Builders.ServiceBus.fs#L39), for example. 

It's better to let the type system enforce it so we are definitely trying to move to that instead of naming conventions (I didn't suggest you use a name like `dns_ttl_in_seconds`).

Adding overloads would be a welcome improvement in a separate PR, just again, be wary of breakage.

BrianVallelunga:
I'll get back to this next week, but I think the basic code is in good shape. I still need to do testing and documentation. The code currently supports external and web apps as endpoints. I have not supported Cloud Services or Nested endpoints as both of those seem more complex and probably used much less frequently. That can be a future addition.

The main area I'm unsure about relates to dependencies. The main builder creates a profile and adds the endpoints as children of the profile. This is what I've seen in every ARM example. However, it should also possible to add the endpoints as separate resources since the builder creates an IArmResource. If this is done though, I would think they would need a dependency reference to the profile.

I can think of two other dependency issues. There might need to be work done with managed/unmanaged resources similar to the `link_to_XXX` members on other builders for web app endpoints. Additionally, I still probably need to add any webapps as dependencies to the profile resource, but only if they're not being built in the same template.

Any advice on the outstanding issues mentioned above would be great. Thanks.

BrianVallelunga:
Yes, I should be able to finish it up tomorrow. Adding the tests should be easy, but I'm still trying to figure out what I want to do with standalone endpoints, if anything.

BrianVallelunga:
I think this is ready for an initial release. I just included a couple of super-basic unit tests in the commit, but could work on some additional ones if needed.

I opted to not create Endpoints as separate resources for the moment since I couldn't work out all of the dependency permutations just yet. If this is needed, I don't think anything would break if I later added that functionality.

Finally, I tweaked the defaults to match what the Azure portal UI defaults to, which make sense to me.

PR #640, opened 2021-05-31
#355 add publish as docker container option to FunctionApp

### LINKED ISSUE #355: Azure function from docker container
Is there a way to configure [AzureFunction from a docker container](https://docs.microsoft.com/en-us/azure/azure-functions/functions-create-function-linux-custom-image?tabs=bash%2Cportal&pivots=programming-language-csharp)?

CC: @isaacabraham

MNie:
This PR closes #355 

* adding `publish_as` to support publishing Azure Functions as a Docker Container.

CC: @ninjarobot I'm curious about the following things:
1. Do we want to pass user/pass/URI as a part of `docker` DU?
2. If we want to pass them do we want to model them as `.NET types` or maybe introduce some of our types like `type DockerUser = User of string`, `ype DockerPass = Pass of string`, etc.?


MNie:
> @MNie it would probably be good to have similar syntax to the image registry credential record that used used with container groups. A [helper function](https://github.com/CompositionalIT/farmer/blob/2def995fe49c35b084f571ee704c6e7b5bb86120/src/Farmer/Builders/Builders.ContainerGroups.fs#L246) keeps it simple to use and generates a `SecureParameter` for the password so registry credentials aren’t embedded in deployments.
> 
> Do you think this is doable for these registry credentials as well?

@ninjarobot thanks for the hint, I will give it a try : )

MNie:
@ninjarobot I replace string with SecureParameter and add a helper function to generate a `DockerInfo` class. The question is should I also add it somewhere to some `IParameters` interface implementation to a `SecureParameters` property implementation (in `Arm/WebApp.fs` perhaps?)?

ninjarobot (changes requested):
@MNie this looks great, thanks again! 

Can you please add one test in `JsonRegression` that can make sure it generates the parameters correctly? I want to ensure that if we change code in other areas, it doesn't create a regression with the parameters here.

MNie:
> @MNie this looks great, thanks again!
> 
> Can you please add one test in `JsonRegression` that can make sure it generates the parameters correctly? I want to ensure that if we change code in other areas, it doesn't create a regression with the parameters here.

@ninjarobot sure : ), test added.

ninjarobot (approved):

PR #642, opened 2021-06-17
Add support for private endpoints

### LINKED ISSUE #535: Add resource and builder for private endpoints
A private endpoint is used to restrict access from a virtual network subnet to a certain resource such that the resource is given an address on the private subnet.

Implementing a builder for [`privateEndpoints`](https://docs.microsoft.com/en-us/azure/templates/microsoft.network/privateendpoints) would enable establishing this connectivity during the deployment process.

TheRSP:
This PR closes #535

* Added `PrivateEndpoint` ARM model
* Added `add_private_endpoint[s]` operators to WebApp
* Added option to allow private endpoints to vnet (disable `privateEndpointSecurityPolicies`)
* Added `SubnetIds` property to `VirtualNetworkConfig` as convenient source of Subnet `ResourceId` instances


ninjarobot (commented):
Awesome, I'm really glad to see this one @TheRSP! I'm going to try it out with a few of the vnet scenarios. Only feedback so far is that it might be clearer to call the subnet builder function `allow_private_endpoints` instead of `private_endpoints Enabled` or `private_endpoints Disabled` just because it looks a little like would be where you actually add the private endpoints rather than just enabling them for a subnet.

TheRSP:
I've renamed the operation to `allow_private_endpoints` as suggested. I've left in the `Enabled|Disabled` (reason below) for now but can remove if you want.

Why I prefer the FeatureFlag variant:
_In our use company, for a few resources, we extend the farmer builders with our own and overwrite the Yield method to enable us to set defaults for all resources within our company (e.g. always_on for web apps) this does lead us into the problem that we can't turn off these features if we turn them on iun our overrides unless they have a disable option. This is why I have included the feature flag._

src/Farmer/Arm/Network.fs
```diff
@@ -85,6 +87,7 @@ type VirtualNetwork =
                                        else
                                            subnet.AssociatedServiceEndpointPolicies
                                            |> List.map (fun policyId -> {| id = policyId.ArmExpression.Eval() |})
+                                   privateEndpointNetworkPolicies = if subnet.AllowPrivateEndpoints then "Disabled" else "Enabled"
```
ninjarobot, line 90: Is this logic backwards? This reads that if they are allowed, then the are "Disabled", but that seems like the opposite of what we would want.

Also, it would be best to not emit a value when it is the default to keep down the size of the resulting template.

src/Tests/test-data/load-balancer.json
```diff
@@ -26,7 +26,8 @@
                     "serviceName": "Microsoft.ContainerInstance/containerGroups"
                   }
                 }
-              ]
+              ],
+              "privateEndpointNetworkPolicies": "Enabled"
```
ninjarobot, line 30: I would not expect them to be "Enabled" here since they weren't explicitly allowed in this subnet.

src/Farmer/Arm/Network.fs
TheRSP: This is intentional. The `privateEndpointNetworkPolicies ` property prevents the use of private endpoints. However, this seemed confusing to me so I inverted the logic such that, in Farmer, we have `allow_private_Endpoints` rather than 'Disable the thing that prevents private endpoints".
I've moved the inversion to the builder and added a comment to this effect

src/Tests/test-data/load-balancer.json
TheRSP: I've modified the builder and ARM model to now take a `FeatureFlag option` so that if no value is set (and the value is therefore `None`) we will emit `null` into the JsonModel so it won't be serialized into the model

ninjarobot (approved):

PR #647, opened 2021-06-21
Add AuthorizationRule to service bus queue

### LINKED ISSUE #646: Feature request: Add authorization rules to Service Bus
I see authorization rules are already present on EventHub resources, would it be possible to fit that into Service Bus resources as well?

biltongza:
This PR closes #646 

* add authorization rule support to service bus queues
* add authorization rule support to service bus namespaces

Unit tests: the tests for service bus seems coupled to the MS azure library service bus model which does not expose the authorization rules on a queue. Next best thing would be to check the output of the generated json template I think?


ninjarobot (commented):
Thanks @biltongza for adding this authorization control. Can these apply to subscriptions as well as queues?

For tests, it would be helpful to test that the builder creates the expected resources in the resulting template.

biltongza:
Yes, it looks like they can. I'll look into this when I get a moment, I have a busy day today.

biltongza:
Turns out I actually need subscription support :) so I'll definitely try get it in plus tests, I should get some time later today for it.

biltongza:
Ok, I've added the namespace support and tests, but one of the tests is failing and I'm a bit stuck. It seems it is using the wrong type for the authorization rules on queues, because the generated resource has the resource type for namespace auth rules instead of the expected queue authorization rules. I'm not sure why.

biltongza:
Does the CI build not run tests? The queue test should definitely be failing but I see CI says all checks passed?

biltongza:
Cool, it's working, thanks so much for the assist! Wish there were a more elegant way of doing that.

I've just tested the changes and made a service bus with auth rules both at the namespace level and the queue level, so I believe this is ready to be merged.

ninjarobot (approved):
Tests look great, thanks for adding those to the PR.

PR #659, opened 2021-06-28
Updating VirtualMachines with linux config

Aaron-Gill:
This PR closes #

* Adds 2 fields for Virtual Machines both for linux configuration
* disable_password_authentication : when true disables password authentication onto the vm
* add_additional_keys: adds a ssh key and path for a virtual machine.

### DIFF
--- a/src/Farmer/Arm/Compute.fs
+++ b/src/Farmer/Arm/Compute.fs
@@ -46,7 +46,6 @@ type CustomScriptExtension =
                                 |> Convert.ToBase64String |}
                         |} :> _
             |} :> _
-
 type VirtualMachine =
     { Name : ResourceName
       Location : Location
@@ -54,13 +53,19 @@ type VirtualMachine =
       Size : VMSize
       Credentials : {| Username : string; Password : SecureParameter |}
       CustomData : string option
+      DisablePasswordAuthentication: bool option
+      PublicKeys : (string  * string ) list option
       Image : ImageDefinition
       OsDisk : DiskInfo
       DataDisks : DiskInfo list
       NetworkInterfaceName : ResourceName
       Tags: Map<string,string>  }
     interface IParameters with
-        member this.SecureParameters = [ this.Credentials.Password ]
+        member this.SecureParameters = 
+            if this.DisablePasswordAuthentication.IsSome && this.DisablePasswordAuthentication.Value then
+                []
+            else
+                [ this.Credentials.Password ]
     interface IArmResource with
         member this.ResourceId = virtualMachines.resourceId this.Name
         member this.JsonModel =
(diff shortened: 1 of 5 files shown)


src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -245,6 +254,10 @@ type VirtualMachineBuilder() =
     [<CustomOperation "custom_data">]
     member _.CustomData(state:VmConfig, customData:string) = { state with CustomData = Some customData }
 
+    [<CustomOperation "linux_configuration">]
```
ericsoler1, line 257: I think other builders usually have `add_linux_configurations` for lists and `add_linux_configuration` for single objects.

Aaron-Gill: 👍

src/Farmer/Arm/Compute.fs
```diff
@@ -75,7 +76,13 @@ type VirtualMachine =
...
                         adminUsername = this.Credentials.Username
                         adminPassword = this.Credentials.Password.ArmExpression.Eval()
-                        customData = this.CustomData |> Option.map (System.Text.Encoding.UTF8.GetBytes >> Convert.ToBase64String) |> Option.toObj |}
+                        customData = this.CustomData |> Option.map (System.Text.Encoding.UTF8.GetBytes >> Convert.ToBase64String) |> Option.toObj 
+                        linuxConfiguration = 
+                            match this.LinuxConfiguration with
+                            | Some linuxConfiguration -> 
+                                {| disablePasswordAuthentication = linuxConfiguration.DisablePasswordAuthentication
+                                   ssh = {| publicKeys = linuxConfiguration.PublicKeys |> List.map (fun k -> {| path = k.path;keyData = k.keyData |}) |} |}
+                            | None -> Unchecked.defaultof<_> |}
```
ericsoler1, line 85: Check on some of the other ArmResource but I'm pretty sure most dont use  Unchecked.defaultof<_>.

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -12,6 +12,10 @@ open System
 
 let makeName (vmName:ResourceName) elementType = ResourceName $"{vmName.Value}-%s{elementType}"
 
+type SshObject = {
```
ericsoler1, line 15: This probably belongs in the Farmer/Arm/Compute.fs file.

src/Farmer/Arm/Compute.fs
ninjarobot: It's needed to turn this anonymous record into a null when the Option isn't set.

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -29,6 +33,7 @@ type VmConfig =
       DomainNamePrefix : string option
 
       CustomData : string option
+      LinuxConfiguration : (bool * SshObject list) option
```
ninjarobot, line 36: It would be better to separate out the two different members on the `VmConfig` record since you'll usually set them with independent builder functions. Also, if you would always need to set the public key and the path together, a tuple of them would ensure both are passed. In any case, make sure the `VmConfig` record is a good structure for the way your builder methods work, because it's easier to deal with mappings in the `BuildResources` implementation. 

Maybe something like

```
DisablePasswordAuthentication : bool option
SshPathAndPublicKeys : (string * string) list
```

src/Farmer/Builders/Builders.Vm.fs
ninjarobot: Prefer simpler types on the builder methods because these can help make the resulting DSL more readable.

src/Farmer/Builders/Builders.Vm.fs
Aaron-Gill: ended up just removing it

src/Farmer/Builders/Builders.Vm.fs
Aaron-Gill: done

src/Farmer/Arm/Compute.fs
Aaron-Gill: done

src/Farmer/Builders/Builders.Vm.fs
Aaron-Gill: done

src/Farmer/Arm/Compute.fs
```diff
@@ -75,7 +76,21 @@ type VirtualMachine =
                      {| computerName = this.Name.Value
                         adminUsername = this.Credentials.Username
                         adminPassword = this.Credentials.Password.ArmExpression.Eval()
```
ninjarobot, line 78: I just tried this out, and it works well, but if password authentication is disabled, it shouldn't emit a password parameter or the `adminPassword` element. You could do something like this so the parameter isn't emitted:

```fsharp
member this.SecureParameters =
    match this.DisablePasswordAuthentication with
        | Some true -> []
        | _ -> [ this.Credentials.Password ]
```

and this so the `adminPassword` ARM element isn't emitted.
```fsharp
adminPassword =
    match this.DisablePasswordAuthentication with
    | Some true -> null
    | _ -> this.Credentials.Password.ArmExpression.Eval()
```

ninjarobot (approved):
Looks great, thanks for fixing that @Aaron-Gill.

PR #671, opened 2021-07-02
VM Identity

isaacabraham:

* Adds Identity support to VM.
* Adds a new `IIdentity` interface for reusable Identity builder support.
* Adds a VM `PasswordParameterARM` member.


isaacabraham:
@ninjarobot I tried adding this to the one other resource that uses Identity (I think Deployment Scripts) but that seems to work differently to e.g. AKS / VM so I left that blank.

src/Tests/VirtualMachine.fs
```diff
@@ -136,4 +137,17 @@ let tests = testList "Virtual Machine" [
...
+
+    test "Handles identity correctly" {
+        let machine = arm { add_resource (vm { name ""; username "isaac" }) } |> findAzureResources<VirtualMachine> client.SerializationSettings |> Seq.head
+        Expect.isNull machine.Identity "Default managed identity should be null"
+
+        let machine = arm { add_resource (vm { system_identity; username "isaac" }) } |> findAzureResources<VirtualMachine> client.SerializationSettings |> Seq.head
+        Expect.equal machine.Identity.Type (Nullable ResourceIdentityType.SystemAssigned) "Should have system identity"
+        Expect.isNull machine.Identity.UserAssignedIdentities "Should have no user assigned identities"
+
+        let machine = arm { add_resource (vm { system_identity; add_identity (createUserAssignedIdentity "test"); add_identity (createUserAssignedIdentity "test2"); username "isaac" }) } |> findAzureResources<VirtualMachine> client.SerializationSettings |> Seq.head
```
ninjarobot, line 149: I'm not sure this usage will really work in practice because the user assigned identity must also be added to the template. It works for a test, but I usually need to create them separately so I can add them to both the template and the resource.

ninjarobot (approved):
This unifies a lot of the code around managed identities to keep the implementation consistent as its added to more resources, looks great!

src/Tests/VirtualMachine.fs
ninjarobot: Maybe we need add something like `link_to_user_assigned_identity` for those scenarios where we want to reference an existing identity.

isaacabraham: Linking to an existing UAI - good point, I don't think we have support for that. Can add that as a separate thing though?

> I'm not sure this usage will really work in practice because the user assigned identity must also be added to the template. It works for a test, but I usually need to create them separately so I can add them to both the template and the resource.

What line are you referring to here - it's not clear from the code sample?

ninjarobot: Linking to an existing identity can be in a later PR, of course. It was really the only thing I could think of in a general framework for Identity.

ninjarobot: I was just referring to using `createUserIdentity` directly in the vm builder, and probably creating is separately is more common so you can assign it to multiple resources and include the resource in the template.

isaacabraham: ah. this is just for the purposes of the unit test. happy to move it out separately if that improves readability.

isaacabraham: Linking to #672

isaacabraham: Referring to your earlier comment - I actually think that if you simply create a managed user identity and *don't* add it to Farmer, it effectively acts as exactly what you're referring to (I think).

Ryan-Palmer-CIT:
Looks good to me!

PR #686, opened 2021-07-09
VirtualHub/HubRouteTables add support for labels.

ericsoler1:
This PR closes #

* add VirtualHubRouteTable builder support for add_default_label and add_labels


src/Farmer/Builders/Builders.VirtualHub.fs
```diff
@@ -73,7 +73,8 @@ type VirtualHubBuilder() =
 type HubRouteTableConfig =
     { Name : ResourceName
       Vhub : LinkedResource
-      Routes : HubRoute list }
+      Routes : HubRoute list
+      Labels : string list }
```
ninjarobot, line 77: Shouldn't this be a `Set<string>` so it can't have duplicates?

src/Farmer/Builders/Builders.VirtualHub.fs
```diff
@@ -133,6 +135,12 @@ type HubRouteTableBuilder() =
     /// Adds the routes to the HubRouteTable
     member _.AddRoutes(state:HubRouteTableConfig, routes) =
         { state with Routes = state.Routes @ routes}
+    [<CustomOperation "add_default_label">]
```
ninjarobot, line 138: Given the default route table is a bit special (only one per hub, routes in the default table are propagated to all hubs in a vWAN), it may need to be represented a little differently in order to ensure a user won't apply the default label to different tables in the same hub. It might be better to only add this label when setting the default route table for a vHub.

```fsharp
vhub {
    // new operation on the builder that accepts a 'HubRouteTableConfig list' and sets
   // the 'Vhub' property on them when added
    add_route_tables [
        hubRouteTable {
            // define the route table here. 
        }
    ]
    // new operation on the builder that accepts a single 'HubRouteTableConfig' and sets
    // the 'Vhub' property and adds the 'default' label.
    default_route_table (
        hubRouteTable {
            // this defines the one route table that will be emitted with a "default" label.
        }
    )
}
```

ninjarobot (changes requested):
The way this is currently defined means the route tables are always built in a separate builder from the vHub. With the default route table, it's really important there is only one specified on the hub. Rather than just a label, the default route table should be defined in the context of the hub itself.

src/Farmer/Builders/Builders.VirtualHub.fs
ericsoler1: I considered the same but Azure allows it to have duplicates.

src/Farmer/Builders/Builders.VirtualHub.fs
ericsoler1: took the add_default_label off

src/Farmer/Builders/Builders.VirtualHub.fs
ninjarobot: Ok, maybe we can fix that up later. If it allows that, but just ignores them, we should do better here.

src/Farmer/Builders/Builders.VirtualHub.fs
ninjarobot: Ok. Hopefully we can figure out some way to represent the hub <-> route table relationship so that it prevents adding more than one default.

ninjarobot (approved):

PR #701, opened 2021-07-17
Refactored WebApp and FunctionsConfig to contain a `CommonWebConfig`

### LINKED ISSUE #696: CommonWebConfig Composition-over-copying
Currently we use a `CommonWebConfig` to represent configuration which is common to WebApps and Functions. To enable this we have methods which copy properties from `WebConfig` and `FunctionsConfig` to `CommonWebConfig` and back.

I was wondering if it might be worth considering using composition here to reduce the overhead involved?
By defining WebAppConfig as follows and a similar change to Functions would allow adding new common properties with less ceremony
```fsharp
{ CommonWebConfig: CommonWebConfig
  HTTPSOnly : bool
  HTTP20Enabled : bool option
  ClientAffinityEnabled : bool option
  WebSocketsEnabled: bool option
  ... }
```

TheRSP:
This PR closes #696

* WebAppConfig no longer duplicates properties of CommonWebConfig but contains an instance of it
* FunctionsConfig no longer duplicates properties of CommonWebConfig but contains an instance of it

- There is no change to visible API surface or output as such, existing tests and release notes are assumed to cover all functionality

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
+* Refactored Web App and Functions builders to simplify adding new common properties
 
 ## 1.6.7
 * Container Groups: Reference Azure container registry credentials.
--- a/src/Farmer/Builders/Builders.Functions.fs
+++ b/src/Farmer/Builders/Builders.Functions.fs
@@ -24,59 +24,16 @@ type PublishAs =
     | DockerContainer of DockerInfo
 type FunctionsExtensionVersion = V1 | V2 | V3
 
-module private FunctionsConfig =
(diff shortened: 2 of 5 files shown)


isaacabraham:
@TheRSP I seem to recall when I first started the "shared" config between Functions and WebApps, I tried this - I think at the time there were either fewer properties and it didn't make as much sense, or it could be because of the cost of updates via nested fields (where F# doesn't have support for things like lenses) made it harder. However, I can't see any obvious issues here - this looks good to me.

I think as long as you can just update the release notes, that's good enough for me.

Thanks!

TheRSP:
Release notes updated as requested

Issue #689, opened 2021-07-09
DNS-zone target_resource wrong parameter type?


Thorium:
```fsharp
    let dns = dnsZone {
        name "myZone"
        zone_type Dns.Public
        add_records [
            aRecord {
                ttl 60
                target_resource trafficMgr.Name
            }
            yield!
                [ "de"; "es";"fr"; "it" ]
                |> List.map (fun nm ->
                    cnameRecord {
                        name nm
                        ttl 3600
                        target_resource trafficMgr.Name
                    }
                )
        ]
    }
```

Fails for:
```JSON
  "error": {
    "code": "LinkedInvalidPropertyId",
    "message": "Property id 'myTrafficManager' at path 'properties.targetResource.id' is invalid. Expect fully qualified resource Id that start with '/subscriptions/{subscriptionId}' or '/providers/{resourceProviderNamespace}/'."
  }
```
...so I expect the `target_resource` should take ResourceId instead of the current ResourceName

```JSON
  "properties": {
      "TTL": 123,
      "targetResource": {
          "id": "/subscriptions/.../resourcegroups/.../providers/microsoft.network/trafficmanagerprofiles/myTrafficManager"
      }
  }
```

Also the traffic manager itself doesn't have a simple Id property, so I expect it has to be created with something like `(ResourceId.create(Farmer.Arm.TrafficManager.profiles, trafficMgr.Name))`

Thorium:
There is no `depends_on` in `dnsZone`.
If I try to hack it via:
`target_resource (ResourceId.create(Farmer.Arm.TrafficManager.profiles, trafficMgr.Name).Eval() |> ResourceName)`
then, from the list of routings, I get the first as success (!), but then:

```JSON
{
  "code": "Conflict",

  "message": "Resource /subscriptions/.../resourceGroups/.../providers/Microsoft.Network/trafficManagerProfiles/myTrafficManager/$fqdn/mytrafficmanager.trafficmanager.net is locked by record set /buckets/6/com/mydomain/mycname. Retry last operation on record set to mitigate."

}

```

ninjarobot:
I was testing this out in some other DNS work today, and I think the `TargetResource` needs to be a `ResourceId` rather than a `ResourceName` [here](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Common.fs#L1602). It probably works as is for some scenarios when the same resource is being deployed in the same deployment and same resource group, but even then, it often will need a full resourceId. 

[...]

PR #703, opened 2021-07-19
Adding DNS records to existing zones

ninjarobot:

* DNS Zone: Support for adding records to existing zones.

### DIFF
--- a/src/Farmer/Builders/Builders.Dns.fs
+++ b/src/Farmer/Builders/Builders.Dns.fs
@@ -11,8 +11,9 @@ open DnsRecords
 type DnsZoneRecordConfig =
     { Name : ResourceName
       Type : DnsRecordType
-      TTL : int }
-    static member Create(name, ttl, recordType) =
+      TTL : int
+      Zone : LinkedResource option }
+    static member Create(name, ttl, zone, recordType) =
         { Name =
               if name = ResourceName.Empty then failwith "You must set a DNS zone name"
               name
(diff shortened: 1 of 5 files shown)


Ryan-Palmer-CIT (commented):
Heya, I had a look over it and it seems fine but I would really get someone else who understands this stuff a bit more than me to check it before merging.

src/Farmer/Builders/Builders.Dns.fs
```diff
@@ -11,25 +11,39 @@ open DnsRecords
...
-type CNameRecordProperties =  { Name: ResourceName; CName : string option; TTL: int option; TargetResource: ResourceName option }
-type ARecordProperties =  { Name: ResourceName; Ipv4Addresses : string list; TTL: int option; TargetResource: ResourceName option  }
-type AaaaRecordProperties =  { Name: ResourceName; Ipv6Addresses : string list; TTL: int option; TargetResource: ResourceName option }
-type NsRecordProperties =  { Name: ResourceName; NsdNames : string list; TTL: int option; }
-type PtrRecordProperties =  { Name: ResourceName; PtrdNames : string list; TTL: int option; }
-type TxtRecordProperties =  { Name: ResourceName; TxtValues : string list; TTL: int option; }
-type MxRecordProperties =  { Name: ResourceName; MxValues : {| Preference : int; Exchange : string |} list; TTL: int option; }
-type SrvRecordProperties =  { Name: ResourceName; SrvValues : SrvRecord list; TTL: int option; }
+    interface IBuilder with
+        member this.ResourceId = zones.resourceId this.Name
```
ericsoler1, line 27: Doesnt this need the parent zone when generated the resourceId?

ericsoler1: Also doesnt this depend on the record type which ResourceId function should be used?

src/Farmer/Builders/Builders.Dns.fs
```diff
@@ -83,9 +102,13 @@ type DnsARecordBuilder() =
     [<CustomOperation "target_resource">]
     member _.RecordTargetResource(state:ARecordProperties, targetResource) = { state with TargetResource = Some targetResource }
 
+    /// Builds a record for an existing DNS zone.
+    [<CustomOperation "link_to_dns_zone">]
```
ericsoler1, line 106: Nit: Is there a convention for the name of this operation? I think usually the link_to_dns_zone would imply that its linking to a managed resource and it should take in a IBuilder while link_to_unmanaged_dns_zone would imply that its linking to an externally deployed dns zone.

ericsoler1 (changes requested):
Looks good other than I think I found a bug on the IBuilder.ResourceId for Records

src/Farmer/Builders/Builders.Dns.fs
ninjarobot: Really good catch, fixing this now.

src/Farmer/Builders/Builders.Dns.fs
ninjarobot: I think the convention has been that when using `link_to_{resource}` it's always unmanaged, linking to some existing resource that's not in the same deployment. If the resource is managed by farmer, it would be in the same template and you'd use an `add_{resource}` operation on the resource, like this.

```fsharp
dnsZone {
  add_records [
    cnameRecord { ... }
  ]
} 
```

ninjarobot: @isaacabraham that's a bit of a design question, though. Should `link_to_{some_resource}` be used for unmanaged linked resources, or should that be used for managed linked resources and have a separate operation `link_to_unmanaged_{some_resource}` for unmanaged linked resources?

Thorium:
This is interesting, as maybe I could get around the current locking problem https://github.com/CompositionalIT/farmer/issues/689#issuecomment-877341371 by deploying the CNAME records manually one-by-one afterwards.

src/Farmer/Builders/Builders.Dns.fs
ninjarobot: I looked at a few others that are doing this same sort of linking and see what you mean @ericsoler1. I changed it to follow that convention.

ericsoler1 (approved):

Issue #260, opened 2020-06-30
Static IP arm-template


Thorium:
I don't know how far Farmer is with this, so I'll try to help. Ignore all you already know or do better. As far as I know, you need first a "Network Interface" as a parent node.

Then you can bind resources like a 
- "Static IP address"
- "Network Security Group" (basic firewall settings)
- and "Virtual Machine" 

...under that "Network Interface". But for easier management, different "Network Interface"s can share one "Network Security Group", so you don't have to configure multiple times the same firewall-settings.
Let's start with the simplest:

#### Static IP
This is a template creating a static IP:

```json
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "publicIPAddresses_Myserver_ip_name": {
            "defaultValue": "Myserver-ip",
            "type": "String"
        }
    },
    "variables": {},
    "resources": [
        {
            "type": "Microsoft.Network/publicIPAddresses",
            "apiVersion": "2020-04-01",
            "name": "[parameters('publicIPAddresses_Myserver_ip_name')]",
            "location": "uksouth",
            "sku": {
                "name": "Basic"
            },
            "properties": {
                "ipAddress": "123.123.123.123",
                "publicIPAddressVersion": "IPv4",
                "publicIPAllocationMethod": "Static",
                "idleTimeoutInMinutes": 4,
                "dnsSettings": {
                    "domainNameLabel": "mydomain",
                    "fqdn": "mydomain.uksouth.cloudapp.azure.com"
                },
                "ipTags": []
            }
        }
    ]
}
```
#### Network Interface

This is how to define a "Network Interface" and bind the IP to it:
(The parameters/publicIPAddresses_Myserver_ip_externalid, I don't know if that could be done any easier way).

```JSON
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
        "networkInterfaces_mynetworkinterface_name": {
            "defaultValue": "mynetworkinterface",
            "type": "String"
        },
        "publicIPAddresses_Myserver_ip_externalid": {
            "defaultValue": "/subscriptions/<subscription-Guid>/resourceGroups/MyGroup/providers/Microsoft.Network/publicIPAddresses/Myserver-ip",
            "type": "String"
        },
[...]

PR #706, opened 2021-07-23
Added support to attach network security groups to VMs.

### LINKED ISSUE #542: Add support for setting the network security group on a VM network interface
There is a builder for network security groups, but no means of assigning them to a VM NIC. They can be assigned by referencing the ID of the NSG, but that needs to be done [here](https://github.com/CompositionalIT/farmer/blob/17ddc02ca68337938aa3de436c6a33a25a4be8f9/src/Farmer/Arm/Network.fs#L223). The NSG resource will also need to be added to the VM's dependencies.

Thorium:
This PR closes #542 and #260

This allows Network Security Group (NSG) to be attached to Virtual Machine (VM).
So you can e.g. manage multiple VMs by shared network security rules.
This does auto-add the NSG dependency to created VM Network interfaces. 

Example:
```fsharp
let myNsg = nsg { name "test-Nsg"; add_rules [ (* ... deny all and whitelist some ips here *) ] }
let spinVm (vmName:string) = vm { name vmName; username "foo"; network_security_group myNsg }
let deployment = arm { add_resources [ myNsg; spinVm "myVm1"; spinVm "myVm2" ] }
```
The above code will not share VNet between the items. 


Thorium:
This works in my tests, but I have a few questions here, regarding to this old piece of code:

https://github.com/CompositionalIT/farmer/blob/439d800fd512b9c7e82df6cc78dc9ca77d77f50e/src/Farmer/Builders/Builders.Vm.fs#L104-L112

- What is a `DeployableResource` of `ResourceRef<'TConfig>`, should NSG be one, or is the current `NsgConfig option` ok?
- Should the NSG be attached to VNET Subnet also, now it's None. I think yes?

> The above code will not share VNet between the items.

- Should the VNET and Subnet be also shared? I don't know that being currently possible with Farmer. What would be the benefits?
   1) Will there be more running-costs of not shared resources?
   2) When sharing, you can get nice topology maps in the Azure portal, like this:
topology
   3) Anything else?

Thorium:
NSG would be easy to add to Subnet by replacing L112 with `NetworkSecurityGroup = nsgId |> Option.map(fun x -> Managed x)` 
But the next question would be: Shouldn't Subnets, and basically every type in farmer that has `LinkedResource option` in it, implement depends_on?

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -301,6 +304,9 @@ type VirtualMachineBuilder() =
     member _.PrivateIpAllocation(state: VmConfig, ref: PrivateIpAddress.AllocationMethod Option) = { state with PrivateIpAllocation = ref}
     member _.PrivateIpAllocation(state: VmConfig, ref: PrivateIpAddress.AllocationMethod) = { state with PrivateIpAllocation = Some ref}
 
-
+    [<CustomOperation "network_security_group">]
+    /// IP allocation method
+    member _.Nsg(state: VmConfig, ref: NsgConfig Option) = { state with NetworkSecurityGroup = ref}
```
ninjarobot, line 309: Is there a good use case for the overload with an option here? It would be inconsistent with other operations, as usually the operation creates the `Some` instance when it is called.

src/Tests/VirtualMachine.fs
```diff
@@ -189,4 +189,13 @@ let tests = testList "Virtual Machine" [
...
         Expect.equal (ipToken.ToString()) (expectedIpToken) "Static IP is wrong or missing"
     }
+
+    test "Can attach to NSG" {
+        let myNsg = nsg { name "testNsg" }
+        let myVm = vm { name "fooVm"; username "foo"; network_security_group myNsg }
+        let deployment = arm { add_resources [ myNsg; myVm ] }
+        let json = deployment.Template |> Writer.toJson
+        Expect.isNonEmpty json "NSG found"
+        //System.IO.File.WriteAllText(@"c:\temp\nsg.json", json)
```
ninjarobot, line 199: Please remove the commented code.

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -98,6 +99,7 @@ type VmConfig =
                         |> Option.map (fun x -> x.toLinkedResource this) |} ]
               VirtualNetwork = vnetName
               PrivateIpAllocation = this.PrivateIpAllocation
+              NetworkSecurityGroup = nsgId
```
ninjarobot, line 102: What if we need to link it to an existing NSG and not one deployed with the VM? In that case, we can't include the NSG an a dependency because it's not in the same template.

src/Farmer/Arm/Network.fs
```diff
@@ -255,6 +255,7 @@ type NetworkInterface =
         {| SubnetName : ResourceName
            PublicIpAddress : LinkedResource option |} list
       VirtualNetwork : ResourceName
+      NetworkSecurityGroup: ResourceId option
```
ninjarobot, line 258: This may need to be a `LInkedResource option` instead. Then scenarios where it should not have a `dependsOn` can be supported by using `Unmanaged`.

src/Tests/VirtualMachine.fs
```diff
@@ -189,4 +189,13 @@ let tests = testList "Virtual Machine" [
         let expectedIpToken = "143.24.20.36"
         Expect.equal (ipToken.ToString()) (expectedIpToken) "Static IP is wrong or missing"
     }
+
+    test "Can attach to NSG" {
+        let myNsg = nsg { name "testNsg" }
+        let myVm = vm { name "fooVm"; username "foo"; network_security_group myNsg }
+        let deployment = arm { add_resources [ myNsg; myVm ] }
+        let json = deployment.Template |> Writer.toJson
+        Expect.isNonEmpty json "NSG found"
```
ninjarobot, line 198: What should we check here in the JSON content?

src/Farmer/Builders/Builders.Vm.fs
Thorium: This pattern I copied from other builder methods, I'd expect the overload is to reset the earlier selected value.
I don't know why that pattern is generally used, Maybe some people want to generate templates and then remove some items?

src/Tests/VirtualMachine.fs
Thorium: I have no idea, that's why I left the write-json commented there. Maybe something like
```JSON
{
  "type": "Microsoft.Network/networkInterfaces",
  "properties": {
	"networkSecurityGroup": {
	  "id": "[resourceId('Microsoft.Network/networkSecurityGroups', 'testNsg')]"
	}
  }
}
```

src/Farmer/Arm/Network.fs
Thorium: That was one of my questions above. As F# developer I immediately know what option type is, but I have no idea what `ResourceRef<'TConfig>` is.

src/Farmer/Builders/Builders.Vm.fs
Thorium: Actually I think it's irrelevant scenario. Why would you create NSGs in-advance, just for fun?
You create NSGs to manage Azure network settings when you create resources like VMs, and as Farmer already supports creating NSGs and you've chosen it to create the VM here, there is no NSG coming outside Farmer.

And, if you want to use "existing settings", just create a new Farmer NSG with those settings.

src/Farmer/Arm/Network.fs
isaacabraham: The docs should describe what a ResourceRef is, if not we can add it. It's more than just an Option.

Thorium: This is linked to the comment above, "What if we need to link it to an existing NSG and not one deployed with the VM?", I can't see a scenario where NSG wouldn't be managed with Farmer, when you manage VMs with Farmer.

src/Farmer/Builders/Builders.Vm.fs
Thorium: Overload removed.

src/Tests/VirtualMachine.fs
Thorium: Removed.

src/Tests/VirtualMachine.fs
Thorium: Check added.

src/Farmer/Builders/Builders.Vm.fs
ninjarobot: One scenario would be that the NSG is deployed with all the other network infrastructure, with compute resources being added later as needed or when scaling up. Also, NSG's are often deployed automatically by a subscription level-policy, and then you would want to use the existing one.

Thorium:
As per code-reviews, I changed the NsgConfig to be LinkedResorce as in SubnetSpecBuilder.
I did run the test and actual deployment, and everything was still working.

ninjarobot (approved):
Looks great, thanks for making the adjustments to also support linking to an existing NSG.

PR #707, opened 2021-07-25
Adding a list of firewall rules to database

Thorium:
This is a small PR.
It only adds `add_firewall_rules` to add list of rules to database firewalls, instead of a single rule.

I wanted to call `add_firewall_rule` dynamically not-yet-defined list of around 15 rules.
So first I tried to call `yield! myList |> List.map add_firewall_rule` but the compiler is not happy, __.For is not defined for this builder or whatever.

add_firewall_rule takes a 3 arguments, so it cannot be easily overloaded with 1 argument.
And it wouldn't be grammatically correct anyway. So that's why the new add_firewall_rules.


ninjarobot (approved):
Looks great, thanks again @Thorium

PR #713, opened 2021-07-29
Bugfix for the ARM deployment quickstart example

g3rm0ney:
Deploy.execute returns Map<string,string>, which has to be ignored


g3rm0ney:
I gues this is the wrong repo, also I have used not my real account (fbehrens) :)

g3rm0ney:
I have realised that it is the right repo,

ninjarobot (commented):
Maybe instead the docs should clarify that it returns a `Map<string,string>` of the output parameters.

fbehrens:
@ninjarobot  Thanks. I was running into this, when I know little about Farmer. It happened that I had to build a full `fsproj` to get the autocomplete working. The compiler complained about the output, while - at the time- I did not know what the output is for. Thanks again, Frank

PR #714, opened 2021-07-30
Merge identities for slots

TheRSP:
This PR closes #

* Slots inherit userAssignedIdentities from their owning webApp

* [ ] ~~**Updated the documentation** in the docs folder for the affected changes.~~

* No existsing documentation covers this level of detail

### DIFF
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -76,7 +76,8 @@ type SlotConfig =
       AutoSwapSlotName: string option
       AppSettings: Map<string,Setting>
       ConnectionStrings: Map<string,(Setting * ConnectionStringKind)>
-      Identity: ManagedIdentity 
+      Identity: ManagedIdentity
+      KeyVaultReferenceIdentity: UserAssignedIdentity option
       Tags: Map<string,string>
       Dependencies: ResourceId Set}
     member this.ToArm (owner: Arm.Web.Site) = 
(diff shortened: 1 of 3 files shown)


src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -109,13 +112,21 @@ type SlotBuilder() =
...
     member this.SystemIdentity (state: SlotConfig) =
-        { state with Identity = { state.Identity with SystemAssigned = Enabled } }
+        { state with Identity = {state.Identity with SystemAssigned = Enabled } }
+        
+    [<CustomOperation "keyvault_identity">]
+    member this.AddKeyVaultIdentity (state: SlotConfig, identity:UserAssignedIdentity) =
+        { state with
+            Identity = state.Identity + identity
+            KeyVaultReferenceIdentity = Some identity
+            AppSettings = state.AppSettings.Add("AZURE_CLIENT_ID", Setting.ExpressionSetting identity.ClientId) }
```
ninjarobot, line 128: Won't this overwrite any other user assigned identity's `AZURE_CLIENT_ID`?

ninjarobot (commented):
Given that setting the keyvault_identity can overwrite the `AZURE_CLIENT_ID` from `add_identity`, it needs a little documentation, at the very least. I'm not sure if there is a way to prevent it.

src/Farmer/Builders/Builders.WebApp.fs
TheRSP: For this, I replicated the existing behaviour in WebApps and functions to ensure that we have consistent behaviour. This does seem to be a problem in all three situations now that the last identity added (including keyvault identity) will set the azure client id

https://github.com/CompositionalIT/farmer/blob/7d400b1e6c744bac7ccc91b602767a46020eb781/src/Farmer/Builders/Builders.WebApp.fs#L703-L708

ninjarobot: Yes, I guess this is not new or unique to slots. We can go ahead and get this in, but I also think we need to come up with a way to address this for all three scenarios and prevent overwriting `AZURE_CLIENT_ID` unexpectedly depending on the order of operations. Maybe we can ensure only one of these is called or include an option to `add_identity` for setting the key vault reference identity at the same time so it's clear to the user when they would have set this setting.

CC: @isaacabraham

Issue #718, opened 2021-08-03
Unlike you'd expect, only one custom script with VMs


Thorium:
I was debugging why the VM custom_script is not executing.

```fsharp
    let devVirtualMachine = vm {
        name machineName
        username vmUsername
        custom_script "md c:\git"
        custom_script "dir > c:\git\tmp.txt" // what directory we are in?
        custom_script "dir %ProgramFiles% >> c:\git\tmp.txt" //Is there any programs?
        custom_script "echo hello3 >> c:\git\tmp.txt" // hello!
    }
```

But actually, the real issue here is that `custom_script` is `Some x` not a list of commands. So it overwrote my trials hiding my errors:
```fsharp
    [<CustomOperation "custom_script">]
    member _.CustomScript(state:VmConfig, script:string) = { state with CustomScript = Some script }
```

Thorium:
Now that I know it, and I know the operating system command separator, I can probably hack around it via something like `custom_script (String.Join(" && ", myScript1, myScript2, ...))` but this should have happened in Farmer, as the API clearly indicates you can add multiple times custom_script without any compilation errors.

Thorium:
@Ryan-Palmer-CIT found good alternative to connect later the VM via Powershell with `Invoke-AzVMRunCommand` to perform more complex installation scenarios.

Basically something like:
```fsharp
// Have Azure connection first: powershell -ExecutionPolicy Unrestricted -Command "Connect-AzAccount -Force"
let install = $"Set-ExecutionPolicy -ExecutionPolicy Unrestricted -Scope Process; Install-Module -Name Az.Compute -Scope CurrentUser -Force -AllowClobber; Import-Module Az.Compute;Invoke-AzVMRunCommand -ResourceGroupName '{deployName}' -Name '{vm.Name.Value}' -CommandId 'RunPowerShellScript' -ScriptPath './localscript.ps1' -Parameter @{{azSubscriptionId='{subId}';user='{vmUsername}'}}"
```

PR #719, opened 2021-08-03
Custom script concat

Thorium:
This PR improves #718

I don't know if this is your perfect solution, but it's a lot better than the current behaviour of removing silently the previous scripts.

### DIFF
--- a/src/Farmer/Builders/Builders.Vm.fs
+++ b/src/Farmer/Builders/Builders.Vm.fs
@@ -272,7 +272,11 @@ type VirtualMachineBuilder() =
     member _.CustomScript(state:VmConfig, script:string) = 
         match state.CustomScript with
         | None -> { state with CustomScript = Some script }
-        | Some previousScript -> { state with CustomScript = Some $"{previousScript} && {script}" }
+        | Some previousScript -> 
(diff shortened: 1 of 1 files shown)


Thorium:
Sorry, branched initially from wrong branch, fixed now.

Thorium:
After this PR, this works:
```fsharp
custom_script "dir > c:\git\tmp1.txt"
custom_script "dir > c:\git\tmp2.txt"
```
...by creating `"commandToExecute": "dir > c:\git\tmp1.txt && dir > c:\git\tmp2.txt"`

But the next question is, what if my commands are ARM-expressions that have parameters:

```fsharp
custom_script "[concat('powershell -ExecutionPolicy Unrestricted -File install-ssl.ps1 ' , subscription().subscriptionId)]"
custom_script $"[concat('powershell -ExecutionPolicy Unrestricted -File install-ssl.ps2 ', '{resourceGroup}')]"
```
...is the && still enough, or should we do something else?

Thorium:
Changed the behaviour to raise exception.
This works for me, I already merged my scripts.

More convenient PRs can be referenced to #718.
Or it can be closed after this as non-urgent thing, however.

ninjarobot (approved):
I like the error message, thanks again @Thorium.

PR #721, opened 2021-08-05
Nested deployments create resource groups

ninjarobot:

* Nested deployments create their resource groups if needed.

work in progress.

### DIFF
--- a/src/Farmer/Builders/Builders.ResourceGroup.fs
+++ b/src/Farmer/Builders/Builders.ResourceGroup.fs
@@ -13,7 +13,7 @@ type ResourceGroupConfig =
       Resources : IArmResource list 
       Mode: DeploymentMode
       Tags: Map<string,string> }
-    member this.ResourceId = resourceGroupDeployment.resourceId (this.Name |> Option.defaultValue "farmer-deploy")
+    member this.ResourceId = resourceGroupDeployment.resourceId (this.Name |> Option.map (sprintf "%s-deploy") |> Option.defaultValue "farmer-deploy")
     member private this.ContentDeployment = 
         if this.Parameters.IsEmpty && this.Outputs.IsEmpty && this.Resources.IsEmpty then
             None // this resource group has no content so there's nothing to deploy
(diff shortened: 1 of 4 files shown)


TheRSP:
As you've noted, creating resource groups inside ARM are only possible when doing a subscription-scoped deployment. As such, I don't think Farmer can support this at the moment as it always does a resource-group-scoped deployment.

When I first tried to add subscription-scoped deployments, I created a parallel interface to the IBuilder (something like `ISubscriptionBuilder`) which then allowed us to have two methods in the ResourceGroupBuilder - one which produced the deployment only and one which produced the resourceGroup itself and, if there were any nested resources/outputs, the deployment as well. This was the only way I could think of to achieve the result of having both produced when possible but only the deployment locally.

However, if we wanted to keep the single interface, it might be good to pass the deployment scope through to the BuildResources method to allow resourceGroupBuilder to make the decision of which resources to emit. (I think there have been a few scenarios where replacing the location argument with some sort of context object has been proposed)

src/Farmer/Builders/Builders.ResourceGroup.fs
```diff
@@ -13,7 +13,7 @@ type ResourceGroupConfig =
       Resources : IArmResource list 
       Mode: DeploymentMode
       Tags: Map<string,string> }
-    member this.ResourceId = resourceGroupDeployment.resourceId (this.Name |> Option.defaultValue "farmer-deploy")
+    member this.ResourceId = resourceGroupDeployment.resourceId (this.Name |> Option.map (sprintf "%s-deploy") |> Option.defaultValue "farmer-deploy")
```
ninjarobot, line 16: @TheRSP I had to change the name here, otherwise if I were to add a `resourceGroups` resource to the same template, it would have the same name and they won't both get emitted.

However, I also have noticed that if you try to do multiple nested deployments targeting the same group, then only one is emitted, for the same reason. Maybe we should either generate a name for the nested deployments based on their order in the sequence (e.g. 'ResourceGroupName-deploy-1', 'ResourceGroupName-deploy-2') or let end users specify a name for the nested deployment itself. What do you recommend?

TheRSP: I like the idea of auto generating the names. It seems to me that this name is a detail of ARM and doesn't seem to have much relevance to the user?

TheRSP:
Nope, that limitation of one nested deployment per resource group was definitely a mistake on my part.

src/Farmer/Builders/Builders.ResourceGroup.fs
ninjarobot: For the most part, yes, we just need to keep from conflicting. I don't find it very useful to keep old ones around so I often reuse the same name (replacing the existing deployment of the same name). Other strategies might be to append a GUID so they are always unique and you keep a history.

Issue #710, opened 2021-07-28
Microsoft Application Insights - Availability Test with VS WebTest


Thorium:
I wanted to share you a Farmer script that creates a WebTest recording, that will reload your site front page every 15 minutes from 5 different locations in the world. Helps you monitor your website uptime and SLA.
If you want to write more complex WebTests, they are just XML.
Feel free to use and embed "natively" to Farmer if you want.

```fsharp
let appInsightsName = "myAppInsights"
let deployLocation = Location.UKSouth

let ai = appInsights {
    name appInsightsName
}

let availabilityTest (testName:string) (urlToVisit:System.Uri) =
    let timeOutSec = 120
    let visitFrequencySec = 900
    let appInsightResource = $"[concat('hidden-link:', resourceId('{Farmer.Arm.Insights.components.Type}', '{appInsightsName}'))]"
    {|  ``type`` = "microsoft.insights/webtests"
        apiVersion = "2015-05-01"
        name = $"{testName.ToLower()}-applicationinsights"
        location = deployLocation.ArmValue
        dependsOn = [
            ResourceId.create(Farmer.Arm.Insights.components, ResourceName appInsightsName).Eval()
        ]
        tags = Farmer.Serialization.ofJson ("{\"" + appInsightResource + "\": \"Resource\"}")
        properties =
        {|  SyntheticMonitorId = $"{testName.ToLower()}-applicationinsights"
            Name = testName
            Enabled = true
            Frequency = visitFrequencySec
            Timeout = timeOutSec
            Kind ="ping"
            RetryEnabled = true
            Locations = [
                {| Id = "emea-nl-ams-azr"  |} // West Europe
                {| Id = "emea-se-sto-edge" |} // North Europe
                {| Id = "emea-gb-db3-azr"  |} // UK West
                {| Id = "emea-fr-pra-edge" |} // France Central
                {| Id = "us-fl-mia-edge"   |} // UK West
            ]
            Configuration =
            {|  WebTest = $"""<WebTest xmlns="http://microsoft.com/schemas/VisualStudio/TeamTest/2010"
                                Name="{testName}"
                                Id="{Guid.NewGuid()}"
                                Enabled="True"
                                CssProjectStructure=""
                                CssIteration=""
                                Timeout="{timeOutSec}"
                                WorkItemIds=""
                                Description=""
                                CredentialUserName=""
                                CredentialPassword=""
                                PreAuthenticate="True"
                                Proxy="default"
                                StopOnError="False"
[...]

PR #723, opened 2021-08-06
Availability tests

Thorium:
This PR brings Microsoft.Insights/webtest Availability Test support to farmer,
as briefly implemented from #710 but now as native Farmer builder resource, to be more convenient for the end-user.

* [x] **Written unit test** against the modified code that I have made.

### DIFF
--- a/src/Farmer/Arm/AvailabilityTest.fs
+++ b/src/Farmer/Arm/AvailabilityTest.fs
@@ -1,0 +1,84 @@
+[<AutoOpen>]
+module Farmer.Arm.InsightsAvailabilityTest
+
+open Farmer
+
+let availabilitytest = ResourceType("Microsoft.Insights/webtests", "2015-05-01")
+
+type AvailabilityTest =
+    { Name : ResourceName
+      AppInsightsName : ResourceName
+      Timeout : int<Seconds>
+      VisitFrequency : int<Seconds>
+      Location : Location
+      Locations : AvailabilityTest.TestSiteLocation list
+      WebTest : AvailabilityTest.WebTestType option }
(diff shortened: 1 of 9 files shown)


src/Farmer/Arm/AvailabilityTest.fs
```diff
@@ -0,0 +1,84 @@
+[<AutoOpen>]
+module Farmer.Arm.InsightsAvailabilityTest
+
+open Farmer
+
+let availabilitytest = ResourceType("Microsoft.Insights/webtests", "2015-05-01")
+
+type AvailabilityTest =
+    { Name : ResourceName
+      AppInsightsName : ResourceName
```
ninjarobot, line 10: It would be good to make this a `LinkedResource option` so we could support adding these in a separate template from defining the AppInsights workspace itself, but that can come later if it turns out to be a common case.

src/Farmer/Common.fs
```diff
@@ -1725,6 +1725,34 @@ module VirtualHub =
...
+    /// Availability test types: WebsiteUrl or CustomWebtestXml
+    type WebTestType =
+    /// Raw Visual Stuido WebTest XML
+    | CustomWebtestXml of string
+    /// URL of website that the test will ping
+    | WebsiteUrl of System.Uri
+
+    /// Availability test sites, from where the webtest is run
+    type TestSiteLocation =
+    | AvailabilityTestSite of Farmer.Location
```
ninjarobot, line 1738: Nice - these made it quite easy to use. Where does the list come from, in case we need to add more?

ninjarobot (approved):

PR #729, opened 2021-08-10
Web app validation

isaacabraham:
This PR closes #

* Add validation to the name of a Site
* Cascade validation to Functions and WebApp builders
* Refactored Site "type" for Slot vs Site.

* No requirement to change documentation

### DIFF
--- a/src/Farmer/Arm/Web.fs
+++ b/src/Farmer/Arm/Web.fs
@@ -156,7 +156,7 @@ type SiteType =
         | Site _ -> sites
 
 type Site =
-    { Name : SiteType
+    { SiteType : SiteType
       Location : Location
       ServicePlan : ResourceId
       AppSettings : Map<string, Setting>
@@ -184,7 +184,10 @@ type Site =
       Metadata : List<string * string>
       AutoSwapSlotName: string option
       ZipDeployPath : (string * ZipDeploy.ZipDeployTarget * ZipDeploy.ZipDeploySlot) option }
-    member this.ResourceType = this.Name.ResourceType
+    /// Shorthand for SiteType.ResourceType
+    member this.ResourceType = this.SiteType.ResourceType
+    /// Shorthand for SiteType.ResourceName
+    member this.Name = this.SiteType.ResourceName
     interface IParameters with
         member this.SecureParameters =
             Map.toList this.AppSettings
@@ -196,7 +199,7 @@ type Site =
     interface IPostDeploy with
         member this.Run resourceGroupName =
             match this with
-            | { ZipDeployPath = Some (path, target, slot); Name = name } ->
+            | { ZipDeployPath = Some (path, target, slot); SiteType = siteType } ->
                 let path =
                     ZipDeploy.ZipDeployKind.TryParse path
                     |> Option.defaultWith (fun () ->
@@ -204,12 +207,12 @@ type Site =
                 let slotName = slot.ToOption
                 printfn "Running ZIP deploy to %s for %s" (slotName |> Option.defaultValue "WebApp") path.Value
                 Some (match target with
-                      | ZipDeploy.WebApp -> Deploy.Az.zipDeployWebApp name.ResourceName.Value path.GetZipPath resourceGroupName slotName
-                      | ZipDeploy.FunctionApp -> Deploy.Az.zipDeployFunctionApp name.ResourceName.Value path.GetZipPath resourceGroupName slotName)
+                      | ZipDeploy.WebApp -> Deploy.Az.zipDeployWebApp siteType.ResourceName.Value path.GetZipPath resourceGroupName slotName
+                      | ZipDeploy.FunctionApp -> Deploy.Az.zipDeployFunctionApp siteType.ResourceName.Value path.GetZipPath resourceGroupName slotName)
             | _ ->
                 None
     interface IArmResource with
-        member this.ResourceId = sites.resourceId this.Name.ResourceName
+        member this.ResourceId = sites.resourceId this.Name
         member this.JsonModel =
             let dependencies = this.Dependencies + (Set this.Identity.Dependencies)
             let keyvaultId = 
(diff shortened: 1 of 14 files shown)


src/Farmer/Arm/Web.fs
```diff
@@ -143,9 +143,20 @@ module ZipDeploy =
...
+        | Site r -> r.ResourceName
+    member this.ResourceType =
+        match this with
+        | Slot _ -> slots
+        | Site _ -> sites
+
 type Site =
-    { Type: ResourceType
-      Name : ResourceName
+    { Name : SiteType
```
ninjarobot, line 159: I understand why you did this, but from a consistency point of view, this would be the only (I think) resource where the `Name` field is not a `ResourceName` so I think it would be helpful to name the field differently (maybe `SiteType`) and have a convenience property like `member this.Name = this.SiteType.Name` like you did with `this.ResourceType` to keep the API relatively consistent. It would also cut out most of the code changes.

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -466,23 +464,23 @@ type WebAppConfig =
...
+                    slot.ToSite site
 
             yield! (PrivateEndpoint.create location this.ResourceId ["sites"] this.PrivateEndpoints)
         ]
 
 type WebAppBuilder() =
     member __.Yield _ =
         { CommonWebConfig = 
-            { Name = ResourceName.Empty
+            { Name = WebAppName.Create("defaultvalue").OkValue
```
ninjarobot, line 483: Is this a sane default or would this be something people would never really want to use? If no one would want this default, could we use `ResourcName.Empty` so we can validate for that and throw an error if they didn't specify a name in the builder?

src/Farmer/Common.fs
```diff
@@ -594,6 +592,20 @@ module WebApp =
     module Extensions =
         /// The Microsoft.AspNetCore.AzureAppServices logging extension.
         let Logging = ExtensionName "Microsoft.AspNetCore.AzureAppServices.SiteExtension"
+    open Validation
+    type WebAppName =
```
ninjarobot, line 596: If supporting `ResourceName.Empty` for validation purposes, it would be helpful to have a member like 
`static member Default = WebAppName (ResourceName.Empty)` so that the builders could use this in `_.Yield` for the default value before anyone sets the name.

ninjarobot (commented):
Just a few suggestions to help with resource type consistency.

src/Farmer/Builders/Builders.WebApp.fs
isaacabraham: I was thinking about this. The problem is the the `WebAppName` of course doesn't allow an empty name ! :-) I copied the same pattern that I used with (I think) SQL Server.

What I could do it make `defaultvalue` a magic string and throw an error (or warning?) on the `Run` method of the builder?

src/Farmer/Common.fs
isaacabraham: See above - `WebAppName` doesn't allow empty resources name e.g. you can't have a web app name with no value. I suppose we could permit it and then just check on the Run method of the builder but that also feels wrong to me.

The real solution would be to have two different types - the Web App Config used during the builder, which used the based ResourceName (no validation), and another type which was created after Run(), which would look identical except use the WebAppName type instead. However, until now I've held off doing this for the simple reason of fear of death by a thousand types.

src/Farmer/Arm/Web.fs
isaacabraham: Fixed this now - I've promoted Name as a member to give back ResourceName and renamed the original Name as SiteType. Looks much better now.

src/Farmer/Builders/Builders.WebApp.fs
ninjarobot: Either can work, just I thought this was the sort of scenario where we often use `ResourceName.Empty`.

ninjarobot (approved):
Approving, as this is a good improvement either way.

src/Farmer/Builders/Builders.WebApp.fs
isaacabraham: Only when we're using `ResourceName`. Storage and SQL both have validation and suffer from this same issue. We could also make the field optional, but this would then bleed downstream which I really wouldn't want.

Let me try a couple of options out and see what they look like and get back to you.

isaacabraham: I've tried an Option variant out - it's too painful to let it bleed out - it's used by Functions, EventGrid, Traffic Manager etc. - by the time we get to them, the optionality should not be there. So that's one possibility we can rule out.

src/Farmer/Common.fs
```diff
@@ -594,6 +592,20 @@ module WebApp =
...
+        private | WebAppName of ResourceName
+        static member Create name =
```
ninjarobot, line 598: Maybe there could be another member for creating a `WebAppName` with `ResourceName.Empty` like this:
```fsharp
static internal member Empty () =
    ResourceName.Empty |> WebAppName |> Result.Ok
```
and the builder uses this in `Yield` to create a `WebAppName` with an empty `ResourceName` instead of one with a magic string. There's a chance people could call this to create a `WebAppName` with `ResourceName.Empty`, but it would be intentional compared to using `"defaultvalue"` as the name.

src/Farmer/Builders/Builders.WebApp.fs
isaacabraham: Ok, I've tried a generic WebAppConfig which could be specialised by what the Name type was e.g. ResourceName or WebAppConfig. This doesn't work either because members can't be bound to generic specializations.

isaacabraham: @ninjarobot if you're cool with this "as-is", please feel free to merge whenever you like - I'd rather not accept my own PR ;-)

src/Farmer/Common.fs
isaacabraham: Hmmm, good idea, let me try that out

src/Farmer/Builders/Builders.WebApp.fs
ninjarobot: I’m good with it and approved it. I’m happy to merge it, but will wait if you’re trying something.

ninjarobot: And you just committed :)

src/Farmer/Common.fs
isaacabraham: Done! There's now a function that looks as follows:

```fsharp
static member internal Empty = WebAppName ResourceName.Empty
```

Notice that it's internal so only Farmer itself can use this. External callers can only call the standard `Create` method. I like this solution - it feels a good compromise between type safety and death by types.

Now Functions and WebApps both use `WebApp.Empty` instead of the "defaultvalue" stuff. In the `Run` method, we then do a quick check and fail there.

ninjarobot (approved):
Oh, I see…the validation meant all the tests needed an update. I like this approach.

isaacabraham:
I'll backport it to the other ones.

PR #732, opened 2021-08-11
Minor addition: Dynamic list of outputs

Thorium:
If you create dynamic amount of resources, you'd probably want to add a dynamic list of `outputs`, for example:
```fsharp
let deployment = arm {
    location deployLocation
    add_resources virtualMachines
    outputs [
        virtualMachines
        |> List.filter(fun v -> v.PublicIp.IsSome)
        |> List.map(fun v -> $"ip-{v.Name.Value}", v.PublicIpId.Value.ArmExpression)
    ]
}
```

This minor addition adds the outputs-keyword.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,7 +1,8 @@
 Release Notes
 =============
-## vNext
+## 1.6.11
 * DNS Zone: Added configuration member of NameServers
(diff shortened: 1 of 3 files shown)


ninjarobot (approved):
Seems like anything broken in a refactor would get caught by the type system, so I can't really think of anything to test here.

PR #736, opened 2021-08-12
PublicIpAddress ARM for VM

### LINKED ISSUE #663: How can I feed a VM's IP address to another resource?
I am setting up a VM with Farmer. 

We also need a TrafficManager which has just been added to Farmer. The TrafficManager contains a list of endpoints, which themselves require an IP address (string).

I would like to pass the IP address of the VM to the TrafficManager at deployment time, a bit like how you can access connection strings of a db resource and add them to settings etc.

I can't see a way to do this. @ninjarobot , @isaacabraham  said you may be able to assist here?

Cheers! :)

Thorium:
This PR closes #663

* Adds VM configration member PublicIpAddress which returns an ARM expression to retrieve public IP address of the virtual machine.

- This adds a minor change, there were are no tests for the other properties either.


ninjarobot (approved):
Thanks again!

PR #742, opened 2021-08-13
Validation cleanup

isaacabraham:

* Create a dedicated `FarmerException` and a factory helper function and apply across the whole application.
* Apply the "Empty" naming pattern for validated resources across SQL and Storage accounts.
* Update the format of validation errors to be a little clearer.


src/Farmer/Common.fs
```diff
@@ -9,7 +9,7 @@ type NonEmptyList<'T> =
 module NonEmptyList =
     let create list =
         match list with
-        | [] -> failwith "This list must always have at least one item in it."
+        | [] -> raiseFarmer "This list must always have at least one item in it."
```
ninjarobot, line 12: I like this over the raw `failwith` since it gives a place to customize the behavior in the future.

src/Farmer/Common.fs
```diff
@@ -305,8 +305,11 @@ module Vm =
     type DiskInfo = { Size : int; DiskType : DiskType }
 
 module internal Validation =
+    // ANDs two validation rules
```
ninjarobot, line 308: It would be helpful to have some tests dedicated to these custom operators to help show their usage.

ninjarobot (approved):
This gives Farmer a nicer validation framework, thank you for adding this. My only suggestion is to add some tests on the custom operators to illustrate their usage, but that can be done in a later PR.

isaacabraham:
Re: 1.7 - probably worth coming up with some roadmap for what 1.6 / 1.7 will contain; then we can close one of them off. etc.

PR #743, opened 2021-08-13
Removed anonymous records from public API

### LINKED ISSUE #741: Do not use anonymous records in public APIs
Just a quick question (I'm playing catch up with all the changes over the last few weeks): We seem to be using anonymous records in the public surface API e.g. for `GeoReplicaServer` on `SQL`.

I'm really very, very reluctant to do this - it exposes another syntax to users of the library for a start and may cause issues from the point of view of C# consumption (believe it or not you can actually use Farmer from there).

How many places are we doing this, and how difficult will it be to revert those to classic F# records? @ninjarobot @Thorium ?

Thorium:
This PR closes #741 

* Moved anonymous records from public API as strongly typed records


ninjarobot (approved):
LGTM. @isaacabraham are you OK with the minor breaking change on these two?

isaacabraham:
Yep, no problem here.

PR #745, opened 2021-08-16
add_resource option

Thorium:
ResourceGroup:
`add_resource` was missing option-overloads. There are already ones in `output`, so just copy&pasted those.

This is to make conditional resources simple to deploy.

For example:
```fsharp
let myProductionResource =
    match env with
    | "production" -> Some resource
    | "test" -> None

let deployment = arm {
    location deployLocation
    add_resource myProductionResource
}
```


ninjarobot (approved):
Great idea - yes this is helpful.

PR #746, opened 2021-08-17
Allow an IArmResource list to be added to a resource group.

Ryan-Palmer-CIT:

* Added an IArmResource list accepting function to the Resource Group builder

The motivation for the change is that I wanted to add a list of role assignments to a resource group.


Ryan-Palmer-CIT:
I've just realised this broke all the other builders as they pick the wrong overload, bear with me...

Ryan-Palmer-CIT:
Ok - after talking to @isaacabraham  I have added a separate function for lists of ARM resources, `add_arm_resources`. Overloading the existing `add_resources` caused all builders to require explicit casting to IBuilder.

isaacabraham:
Just needs docs adding for the builder.

Ryan-Palmer-CIT:
Ok added the docs bit

isaacabraham:
Thanks @Ryan-Palmer-CIT !

PR #748, opened 2021-08-19
Add support for FTPS-state and IpSecurityRestrictions

### LINKED ISSUE #603: Add ftpsState and ipSecurityRestrictions to WebApp
We are using the settings for *ftpsState* and *ipSecurityRestrictions* for our webapps and have had to make custom changes to Farmer for this.

We'd be happy to make a pull-request to add this functionality to the library. Can we push a draft PR and discuss it?

viktorvan:
This PR closes #603

* Add support for FtpsState for webapp and functions
* Add support for IpSecurityRestrictions for webapp and functions

### DIFF
--- a/src/Farmer/Arm/Web.fs
+++ b/src/Farmer/Arm/Web.fs
@@ -184,7 +184,9 @@ type Site =
       Metadata : List<string * string>
       AutoSwapSlotName: string option
       ZipDeployPath : (string * ZipDeploy.ZipDeployTarget * ZipDeploy.ZipDeploySlot) option
-      HealthCheckPath : string option }
+      HealthCheckPath : string option 
+      FtpsState : FtpsState  option
+      IpSecurityRestrictions : IpSecurityRestriction list }
     /// Shorthand for SiteType.ResourceType
     member this.ResourceType = this.SiteType.ResourceType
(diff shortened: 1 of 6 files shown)


viktorvan:
Let's discuss the proposed API for FtpsState and IpRestrictions, and when we have settled on something I will update the docs with the new settings.

viktorvan:
Ok, so we can skip anything related to FTPS state in this PR. 

For the api for IPRestrictions we can try to be consistent with storage accounts. 
When I just now had a quick glance at the arm schema for storage accounts it seems to me they only support they only support “Allow” action in IPRule. For webapp we also want to support “Deny”. 

The name used in the storage account builder gets a little ambiguous when we need to support allow/deny. 

E.g. what does `restrict_to_ip` mean? Would the opposite be `restrict_for_ip`? Not very clear to me which is which?
IMO A better api would be `allow_for_ip` and `deny_for_ip`.

But there is another issue with IP restrictions, and that is that they are not swapped during a slot swap, so for this PR we need to verify that adding this setting does not break anything related to web app slots. And probably also be clear in the documentation on this behavior.

viktorvan:
I am closing this PR now, since it has drifted away from master. 
When I have the time to revisit IpRestrictions I will open a new PR with only that change.

PR #752, opened 2021-08-20
Support for Alerts.

Thorium:

* Initial support for Alerts
* Changed some AppInsight common values to Farmer.Insights.

### DIFF
--- a/src/Farmer/Arm/Alert.fs
+++ b/src/Farmer/Arm/Alert.fs
@@ -1,0 +1,137 @@
+[<AutoOpen>]
+module Farmer.Arm.InsightsAlerts
+
+open Farmer
+open Farmer.Insights
+
+let metricAlert = Farmer.ResourceType("microsoft.insights/metricAlerts", "2018-03-01")
+
+type AlertSeverity =
+/// 0
+| Alert_Critical
+/// 1
+| Alert_Error
+/// 2
+| Alert_Warning
+/// 3
+| Alert_Informational
+/// 4
+| Alert_Verbose
+
+type MetricComparison =
+| GreaterThan
+| LessThan
+
+type MetricAggregation =
+| Average
+| Count
+| Maximum
+| Minimum
+| Total
+
+/// See the MetricNames and their Aggregations:
+/// https://docs.microsoft.com/en-us/azure/azure-monitor/essentials/metrics-supported
+type ResourceCriteria = {
+    /// Resource type name
+    MetricNamespace : ResourceType
+    /// Name of the metric
+    MetricName : MetricsName
+    /// Threshold to exceed to hit the alert
+    Threshold : int
+    /// GreaterThan or LessThan
+    Comparison : MetricComparison
+    /// Average, Count, Total, Maximum, Minimum
+    Aggregation : MetricAggregation
+}
+
+/// Metric criterias
+/// https://docs.microsoft.com/en-us/azure/templates/microsoft.insights/metricalerts?tabs=json#metricalertcriteria
+type MetricAlertCriteria =
+| MultipleResourceMultipleMetricCriteria of MultiCriterias : obj list
+/// If avg of metric x is going over threshold for selected windowSize time. E.g. if average of VM CPU is going over 80% for 15 minutes -> alert
+| SingleResourceMultipleMetricCriteria of Criterias : ResourceCriteria list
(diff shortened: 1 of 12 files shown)


src/Farmer/Common.fs
```diff
@@ -399,6 +400,29 @@ module ServiceBusValidation =
 
         member this.ResourceName = match this with ServiceBusName name -> name
 
+module Insights =
+
+    type DurationInterval =
+    | ISO8601DurationFormat of string
```
ninjarobot, line 406: Instead of hardcoding these strings, they can be created using [this](https://github.com/CompositionalIT/farmer/blob/aea530a502e5402b6a6d140850abc35387a54769/src/Farmer/Common.fs#L87) (although maybe we should rename that to `IsoDuration`).

src/Tests/Alerts.fs
```diff
@@ -0,0 +1,99 @@
...
+open Farmer.Builders
+open Farmer.Arm.InsightsAlerts
+
+let tests = testList "Alerts" [
+    test "Create a VM alert" {
+        let vm = vm { name "foo"; username "foo" }
+        let vmAlert = alert { 
+            name "myVmAlert2"
+            description "Alert if VM CPU goes over 80% for 15 minutes"
+            frequency (DurationInterval.FiveMinutes)
```
ninjarobot, line 15: Are the parentheses needed here?

src/Farmer/Arm/Alert.fs
```diff
@@ -0,0 +1,137 @@
+[<AutoOpen>]
+module Farmer.Arm.InsightsAlerts
+
+open Farmer
+open Farmer.Insights
+
+let metricAlert = Farmer.ResourceType("microsoft.insights/metricAlerts", "2018-03-01")
+
+type AlertSeverity =
```
ninjarobot, line 9: It might be better usage with `[<RequireQualifiedAccess>]` and then the DU cases could be named `Critical`, `Error`, etc (no underscore).

src/Tests/Alerts.fs
```diff
@@ -0,0 +1,99 @@
...
+            window (DurationInterval.FifteenMinutes)
+            add_linked_resource vm
+            severity AlertSeverity.Alert_Warning
```
ninjarobot, line 18: Given the comment above, this could be `Severity = AlertSeverity.Error`.

src/Tests/Alerts.fs
```diff
@@ -0,0 +1,99 @@
...
+            criteria 
+                (SingleResourceMultipleMetricCriteria [
```
ninjarobot, line 20: Could these be combined in the builder operation so usage is something like this for adding multiple a list of criteria (so usage isn't mixing parens and arrays)?
```
single_resource_multiple_metric_criteria [
  {
     MetricNamespace = vm.ResourceId.Type
     MetricName = MetricsName.PercentageCPU
     Threshold = 80
     Comparison = GreaterThan
     Aggregation = Average
  }
]
```

ninjarobot (commented):
It will be great to have these alerts, thanks again @Thorium !

I have a few questions and suggestions around the operations and usage to see if we can simplify it a bit for end users.

src/Tests/Alerts.fs
Thorium: Removed

src/Farmer/Arm/Alert.fs
Thorium: Good idea, I had forgotten that attribute.

src/Tests/Alerts.fs
Thorium: This is good idea, thanks. Implemented.

src/Farmer/Common.fs
Thorium: I ran out of imagination here. I originally thought frequencies should use F# units of measure, one divided by time, like `let hertz = 1m/1m<Seconds>`. But that's not very intuitive. 
How would you easily say "once per 15 mins"?

Thorium: Changed to IsoDateTime as that is the same functionality. I do agree IsoDuration would be better name, and TimeSpans are familiar to .NET developers.

ninjarobot (approved):
Thanks @Thorium - this is clearer to use. Merging now.

PR #756, opened 2021-08-23
Dashboads: Fixes for complex dashboards: custom parts and monitor parts.

Thorium:
This PR continues adding more features after we now already have the initial support of simple dashboards.

* Dashboards, add_monitor_chart was not working, mainly because of F# Option type JSON serialization. Option-type had to be removed from the class serialization
* This also adds an unit test of complex dashboard with monitor-parts and custom-part which is tested to be working.
* Some documentation updates.


Thorium:
Basically the serialization issue was that, in the following code:

```fsharp
type MyObj = { x: string; y: string }
let itm = {x="hi"; y=null}
```

as expected, itm serializes to 
```JSON
{"x":"hi"}
```

...so I expected that in Farmer serialization the following F# would serialize the same JSON:
```fsharp
type MyObj = { x: string option; y: string option }
let itm = {x=Some "hi"; y=None}
```
But instead it created extra Value object, so the None case works but the Some case doesn't:
```JSON
{"x":{"Value":"hi"}}
```

This PR removes the extra Value-wrapper that was breaking the ARM template.

src/Farmer/Arm/Dashboard.fs
```diff
@@ -46,32 +46,32 @@ let generateVideoPart (videoProperties:VideoPartParameters) = {
     ``type`` = "Extension[azure]/HubsExtension/PartType/VideoPart"
     inputs = List.empty
     settings = {| content = {| settings = {| title = videoProperties.title; subtitle = videoProperties.subtitle; src = videoProperties.url; autoplay= false |} |} |} :> obj
-    filters = None
-    asset = None 
-    isAdapter = None
-    defaultMenuItemId = None
+    filters = null
```
ninjarobot, line 49: This has bitten me quite a few times as well. I think the next release of System.Text.Json will have more awareness of DU's and serialize options more like you'd expect.

ninjarobot (approved):
Thanks for fixing the DU serialization and adding improved test coverage. Merging.

PR #757, opened 2021-08-24
Allow deploying app slots without deploying the production slot

### LINKED ISSUE #750: WebApp with slot deletes all appSettings and connectionStrings from production slot during ARM deployment.
When deploying a webapp that has slots enabled the currently generated arm-template will remove all appSettings and connectionStrings from the production slot.

These are the offending lines in Builders.WebApp.fs L476:
```fsharp
            if Map.isEmpty this.CommonWebConfig.Slots then
                site
            else
                { site with AppSettings = Map.empty; ConnectionStrings = Map.empty }
                for (_,slot) in this.CommonWebConfig.Slots |> Map.toSeq do
                    slot.ToSite site
```

together with the template generation in Web.fs L239:
```fsharp
appSettings = this.AppSettings |> Map.toList |> List.map(fun (k,v) -> {| name = k; value = v.Value |})
connectionStrings = this.ConnectionStrings |> Map.toList |> List.map(fun (k,(v, t)) -> {| name = k; connectionString = v.Value; ``type`` = t.ToString() |})
```

This results in a generated arm template where the webApp (production slot) gets the following in properties::siteConfig:
```json
    {
          "appSettings": [],
          "connectionStrings": [],
    }
```

And when this arm template is deployed it removes any appsettings and connectionstrings from the production slot.
This will of course cause the webapp in the production slot to crash and not work until the slots are swapped again.

For me, the expected behaviour is that appSettings and connectionStrings will be set to null which means that they will be ignored and no changes are made to the production slot.

Updating Web.fs to set appSettings and connectionStrings to null if they are Map.empty solves the issue for me at least.

TheRSP:
This PR closes #750

* The production app slot can now take settings
* Added `deploy_production_slot` operator to webApp & functions to enabled disabling production slot deployment

On of the main purposes of slots in webapps and functions is to enable zero-downtime deployments. This is achieved by making all your changes to the staging slot and then swapping that slot with the production slot in a way which does not cause any downtime in the production slot. However, this acting of swapping effectively changes the production infrastructure in a way which no long matches the ARM template it was generated from. This PR adds the option to exclude the production slot from the ARM template and, in so doing, prevents the production slot being affected directly. This in turn permits the zero-downtime deployment model which is desirable.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,6 +1,10 @@
 Release Notes
 =============
 
+## 1.6.14
+* WebApp: allow disabling deployments to the production slot of multi-slot apps
+* Functions: allow disabling deployments to the production slot of multi-slot apps
+
 ## 1.6.13
 * Container Groups: Fix to generate parameters for secure environment variables on `initContainers`.
 * Alerts: Initial support for Alerts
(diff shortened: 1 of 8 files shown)


TheRSP:
Putting this PR on hold as there may be a much better solution

PR #759, opened 2021-08-24
Custom domain added

### LINKED ISSUE #213: Support for Custom domains for Web App
It would be nice to be able to add mydomain.com, and even better the free ssl certificate part as well. (Luckily it doesn't get removed if one redeploy.)

Be aware: Custom domains are not available for the free tier.

I guess this is some documentation on the matter: 
https://github.com/Azure/azure-quickstart-templates/blob/master/201-web-app-custom-domain-and-ssl/azuredeploy.json

AbdulCodat:
This PR closes #213 

* Added Custom domain property to Web
* ResourceGroup name type changed to ArmExpression
* Added extra step in builder to check for custom domain


src/Farmer/Builders/Builders.ResourceGroup.fs
```diff
@@ -5,32 +5,31 @@ open Farmer
 open Farmer.Arm.ResourceGroup
 
 type ResourceGroupConfig =
-    { Name: string Option
+    { Name: ArmExpression
```
ninjarobot, line 8: Why does the name need to be an `ArmExpression` for a custom domain name on the web app?

src/Farmer/Arm/ResourceGroup.fs
```diff
@@ -11,14 +11,15 @@ type DeploymentMode = Incremental|Complete
 
 /// Represents all configuration information to generate an ARM template.
 type ResourceGroupDeployment =
-    { Name: ResourceName
+    { Name: ArmExpression
```
ninjarobot, line 14: What is it about customizing the domain that would lead to needing an `ArmExpression` to calculate the resource group name at deployment time?

src/Farmer/Arm/ResourceGroup.fs
```diff
@@ -32,19 +33,23 @@ type ResourceGroupDeployment =
                 (function 
                 | :? ResourceGroupDeployment as rg -> rg.RequiredResourceGroups 
                 | _ ->  [])
-        List.distinct (this.Name.Value :: nestedRgs)        
+        if this.ResourceName.[0] = '[' then
```
ninjarobot, line 36: Is there a way to check if the type is an `ArmExpression` rather than comparing on the string that's output from the `Eval()`?

src/Farmer/Arm/ResourceGroup.fs
```diff
@@ -32,19 +33,23 @@ type ResourceGroupDeployment =
...
           Outputs = this.Outputs |> Map.toList
           Resources = this.Resources }
     interface IParameters with 
         member this.SecureParameters = this.Parameters
     interface IArmResource with
-        member this.ResourceId = resourceGroupDeployment.resourceId this.Name
+        member this.ResourceId = resourceGroupDeployment.resourceId this.ResourceName
         member this.JsonModel = 
-            {| resourceGroupDeployment.Create(this.Name, this.Location, dependsOn = this.Dependencies, tags = this.Tags ) with
+            {| resourceGroupDeployment.Create(ResourceName this.ResourceName, this.Location, dependsOn = this.Dependencies, tags = this.Tags ) with
```
ninjarobot, line 50: Does this just need to customize the deployment name rather than the name of the resource group it's targeting?

ninjarobot (commented):
Thanks for adding domain customization to web apps. Is it possible to make this change without a change to the deployment pipeline as well with the `ArmExpression` for the resource group and deployment name? It has such a broad impact that it would be safer to do changes to the resource group and nested deployments in a separate PR. @isaacabraham may also have some suggestions on this.

src/Farmer/Arm/ResourceGroup.fs
TheRSP: This is primarily because setting the domain requires a nested deployment and that nested deployment needs to know the name of the outer resource group. We considered changing the BuildResource function to take a context object including the resource group name but this seemed like the smaller change

src/Farmer/Builders/Builders.ResourceGroup.fs
ninjarobot: I understand there are scenarios where we might want the resourceGroup name to be calculated in an expression, but what I don't understand is how this correlates to a custom domain on a web app. Could customizations to the target resource group name be implemented independently of this PR to customize the domain name?

TheRSP: Ah I understand. The problem we were faced with here is a circular depenedency between hostname binding and certificate.
The certificate depends on the hostNameBinding but the hostNameBinding's SSL state depends on the certficate. As such, it is impossible to deploy a webapp with certfiicate and custom domain all in one deployment. The solution we had for this was to use the nested resource group deployment to provide a second phase in which we update the hostName binding to enable TLS with the certificate.

This can be seen in action in Builders.WebApp.fs Lines 480-512.

It is possible that we could pull out the ResourceGroup name change into a separate PR. However, the rest of this custom domain PR wouldn't be possible until that had been approved.

src/Farmer/Arm/ResourceGroup.fs
AbdulCodat: There probably is but not sure what it could be (beginner), I know that when we don't do this it appends the resource group name that we added in the webapp builder (`"[resourceGroup.Name()]"`) and we end up with a console log llike this:

```
Creating resource group my-resource-group-name-2 (1/2)...
Creating resource group [resourceGroup().name] (2/2)...
```

AbdulCodat:
@ninjarobot It makes sense to not include this change in this PR so I've reverted the change to the resourceGroup name type

AbdulCodat:
Bump @ninjarobot , @isaacabraham 🙂 what's your thoughts on the latest changes so far?

src/Farmer/Arm/ResourceGroup.fs
```diff
@@ -32,7 +32,10 @@ type ResourceGroupDeployment =
                 (function 
                 | :? ResourceGroupDeployment as rg -> rg.RequiredResourceGroups 
                 | _ ->  [])
-        List.distinct (this.Name.Value :: nestedRgs)        
+        if this.Name.Value.[0] = '[' then
```
ninjarobot, line 35: I'm still a bit confused on what this is trying to do. It looks as though detecting it's using an ARM expression for the resource group name would mean it wouldn't be included in the list of resource groups. This is still making the assumption that a nested deployment is named for and 1:1 with a resource group. Ultimately, this seems to just be used when deploying the resource groups to try to infer which ones are needed. Can we simplify this and actually emit `resourceGroups` resources when they are added by the builder?

src/Farmer/Arm/Web.fs
```diff
@@ -313,6 +315,58 @@ type StaticSite =
         member this.SecureParameters = [
             this.RepositoryToken
         ]
+type SslState = 
+    | SslDisabled
+    | SniBased of thumbprint: ArmExpression
```
ninjarobot, line 320: Why is an expression needed here? In the tests, it's just using a literal anyway, but is there a better illustrative example of why it would be dynamically defined in the template?

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -476,6 +477,46 @@ type WebAppConfig =
...
+                    { Location = location
+                      SiteId = this.ResourceId
+                      ServicePlanId = this.ServicePlanId
+                      DomainName = customDomain }
+                hostNameBinding
+                cert
+                let resourceLocation = location
+
+                // nested deployment to update hostname binding with specified SSL options
+                yield! (resourceGroup {
```
ninjarobot, line 497: Why does a custom domain emit an entirely separate deployment? Are they always going to be in separate resource groups and have to use separate deployments or is this more specific to how you happen to be deploying these web apps?

ninjarobot (commented):
I'm still not sure if there is something about using a custom domain on a web app that needs to result in these nested deployments and dynamic resource group names. Is that something more specific to the way you happen to be deploying them, and if so, are we forcing that same topology on anyone that needs to deploy a webapp with a custom domain?

TheRSP:
@ninjarobot this is simply the only way we could figure out to get around the circular dependency:
The cert depends on the hostnameBinding and the hostnameBinding's SslState depends on the cert.

To solve this we deploy the app with a host name and cert but don't enable TLS - this can be done in a single template. We then use a second deployment _to the same resource group_ in order to update the hostnameBinding to turn on TLS. As the cert has already been deployed by the time this second deployment runs, this is now possible.

It's not possible in a single deployment to deploy a webapp with a custom domain and app service managed certificate. As such we require two deployments. This could either be done by having two completely independent deployments which the user has to run in order, or, we can use this trick with a nested resource group deployment to create two deployments in one. This lets us perform a create-and-update operation to get around the circular dependency.

src/Farmer/Arm/Web.fs
ninjarobot: I see - if it's SNI based, then it will _either_ come from a referencing an existing certificate resource's thumbprint or from a literal thumbprint. Could this be changed to represent this in the type system?

```fsharp
type CertificateThumbprint =
    | Thumbprint of string
    | Certificate of ResourceId

type SslState =
    | SniDisabled
    | SniBased of CertificateThumbprint
```
and then when building the `JsonModel` for the `HostNameBinding`, you create an ArmExpression only when referencing a certificate (SslState.SniBased(Certificate certificateResourceId)?

src/Farmer/Builders/Builders.WebApp.fs
ninjarobot: @TheRSP as you explained, this needs to be a separate deployment for the host name binding to be modified to enable TLS after the certificate is issued. My one concern with this is that the nested deployments are always given the same name as the resource group, so if there are other nested deployments for the same resource group in this same template, whichever is defined last will be present in the emitted template.

To prevent any collisions, can you address this issue in the nested deployments so they can have a different name than the resource group they're being deployed to?

PR #767, opened 2021-08-30
Allow traffic manager endpoints without weight or priority

ninjarobot:

* Traffic Manager: allow priority and weight to be optional for endpoints.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 Release Notes
 =============
+
+## 1.6.16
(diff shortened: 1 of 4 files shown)


Thorium:
I don't know if this makes the routing more clear or not. I'm using `routing_method` priority based, so I expect then it's not optional for me, but if you'd use some kind of traditional round-robin, then maybe.

I don't have strong opinion here, but I think the effort should be making `routing_method` modes clear: what parameters are for what routing and what they'll do.

Then again, you may want to pre-setup your weights just if you consider changing the routing_method later.

Thorium:
Fair enough. :-)

PR #772, opened 2021-09-03
Replicated database itself added to geo-replica dependency

Thorium:
Sorry to not notice this earlier but I was replicating existing databases, not new ones.

* Geo-replicated database dependency added to the geo-replica deployment.

Yes, tested: Without this it failed on a new database, and with it, the deployment went through. This is a minor fix, shouldn't need anything else.


ninjarobot (approved):
Thanks for the fix, @Thorium

PR #779, opened 2021-09-22
Dashboard add_monitor_chart minor fix

Thorium:
Dashboard add_monitor_chart minor fix:
Instead of `obj list list` use just `obj list` when generating json.

The `add_monitor_chart` did work already when deployed, but it crashed the deployment if you do `Deploy.whatIf`.

Annoyingly the deployment crashed to weird error:
`Farmer.FarmerException: ERROR: object of type 'NoneType' has no len()`
...which is not well known by Google.

### DIFF
--- a/src/Farmer/Arm/Dashboard.fs
+++ b/src/Farmer/Arm/Dashboard.fs
@@ -98,7 +98,7 @@ let generateMonitorChartPart (chartProperties : MonitorChartParameters) = {
     inputs = [ box <| {| name = "sharedTimeRange"; isOptional = true |};
                box <| {| name = "options"
                          value = {| v2charts = true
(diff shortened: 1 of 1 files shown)


Thorium:
When you download a dashboard ARM, it in fact has a double array `charts = [ [ ... ] ]`, but I tested this change and it still works.

[ x ] Tested my code end-to-end against a live Azure subscription.

Thorium:
I still think this is bug in either Azure Portal Dashboards ARM template with MonitorChartPart, or this is bug in Azure Cli what-if command not handling double-arrays. This is definitely not Farmer bug, and for that reason I don't know should it even be fixed here.

ninjarobot (approved):
Thanks for the fix - or workaround the Azure CLI bug, really.

PR #786, opened 2021-10-06
Allow multiple nested deployments

TheRSP:
This PR closes #

* Resource Group deployments use an automtiaclly-generated deployment name


TheRSP:
When used in tests, I've chosen to hard-code the deployment name to ensure consistency when adding additional tests. There are better solution to this I can think of but they are much wider-ranging changes (e.g. using a context value passed to BuildResources and having the resourceGroup store IBuilders rather than IArmResources thus delaying the call to BuildResources until much later)

PR #805, opened 2021-10-27
Fix for functions referencing external resources

### LINKED ISSUE #802: Function Link to Unmanaged AppInsights ResourceId ignores Resource Group
When creating an ARM template for a function, setting the **link_to_unmanaged_app_insights** accepts a **ResourceId**, but if the **ResourceId** has an externally managed Resource Group this is ignored when the **APPINSIGHTS_INSTRUMENTATIONKEY** value is set.

As an example, I can define the function and external AppInsights like this: 

```
let myAppInsightsName = ResourceName.ResourceName "my-existing-ai"
let myAppInsights = ResourceId.create(Arm.Insights.components, myAppInsightsName, "my-existing-rg")
let myAzureFunction = functions {
        name "my-important-function"
        link_to_unmanaged_app_insights myAppInsights
}
let deployment = arm {
        location Location.WestEurope
        add_resource myAzureFunction
}
```

but when I generate the ARM template, I get:

```
            {
                "name": "APPINSIGHTS_INSTRUMENTATIONKEY",
                "value": "[reference(resourceId('Microsoft.Insights/components', 'my-existing-ai'), '2014-04-01').InstrumentationKey]"
            },

```
I was expecting the **ResourceId** to look like this:

`resourceId('my-existing-rg', 'Microsoft.Insights/components', 'my-existing-ai')`

Have I missed something?

TheRSP:
This PR closes #802

* Reinstate resource group in AppInsights and storage account references
* Add tests to prevent regression

- Bug fix with not documentation change required


src/Farmer/Builders/Builders.Functions.fs
```diff
@@ -53,10 +53,10 @@ type FunctionsConfig =
     member this.ServicePlanId = this.CommonWebConfig.ServicePlan.resourceId this.Name.ResourceName
     /// Gets the Service Plan name for this web app.
     member this.ServicePlanName = this.ServicePlanId.Name
-    /// Gets the App Insights name for this functions app, if it exists.
-    member this.AppInsightsName : ResourceName option = this.CommonWebConfig.AppInsights |> Option.map (fun ai -> ai.resourceId(this.Name.ResourceName).Name)
-    /// Gets the Storage Account name for this functions app.
-    member this.StorageAccountName : Storage.StorageAccountName = this.StorageAccount.resourceId(this).Name |> Storage.StorageAccountName.Create |> Result.get
```
ninjarobot, line 59: Removing these would be a breaking change if anyone were using these properties, wouldn't it? Is there any harm in leaving them in, maybe marked as Obsolete and removed from docs and samples so people will be less likely to use them?

samples/scripts/service-bus.fsx
```diff
@@ -1,34 +1,36 @@
...
-                    name "thesub"
-                    forward_to "queuenumberone"
-                }
-            ]
+            name "existing-topic"
         }
     ]
 }
 
+let followUpDeployment = resourceGroup {
```
ninjarobot, line 17: It looks like the changes in this file were for experimentation and not really meant to be included.

ninjarobot (changes requested):
If possible, I'd like to avoid a breaking change to the builders on a minor or patch release. Could we leave those in and guide users away from them for now?

TheRSP:
@ninjarobot I have updated this as requested

ninjarobot (approved):
Thanks again @TheRSP!

PR #806, opened 2021-10-28
Custom Domain added

TheRSP:
This PR supersedes #759 
The only change I have made is to resolve the single merge conflict in Arm/ResourceGroup.fs


TheRSP:
I have fixed the build. However, I'm not really sure how best to squash commits given the mixture of rebasing and merging in the history. I could try to rebase squash the whole PR?
(I've never really had to squash before so don't really know what I'm doing here)

src/Farmer/Arm/ResourceGroup.fs
```diff
@@ -60,7 +60,11 @@ type ResourceGroupDeployment =
                 (function 
                 | :? ResourceGroupDeployment as rg -> rg.RequiredResourceGroups 
                 | _ ->  [])
-        List.distinct (this.TargetResourceGroup.Value :: nestedRgs)        
+        // Ignore this target resourcegroup if it is an ARM expression caluclated at deploy-time
+        if this.TargetResourceGroup.Value.[0] = '[' then
```
ninjarobot, line 64: Since this filtering is only for the `Deploy.prepareForDeployment` function, I suggest moving it there rather than doing this in a property on the deployment itself. Then this property could still be used if including the resource groups as resources in the ARM template.

It also would be better to use something like `this.TargetResourceGroup.Value.StartsWith('[')` so it wouldn't end up with an exception should this be an empty string.

ninjarobot (changes requested):
Just some minor cleanup and this should be ready to merge.

src/Farmer/Arm/ResourceGroup.fs
ninjarobot: @TheRSP do you mind if I merge this and then make this change? I'd like to get this in for this release.

TheRSP: Yes, please do. Sorry I haven't got around to this. I've been a bit swamped recently :(

PR #813, opened 2021-11-09
Update to build and test with dotnet 6.0 SDK

ninjarobot:
Testing build pipeline with .NET 6.0.


forki:
please wait with this until the task {} compiler fix landed in .NET SDK

mattgallagher92:
The `task` computation expression bug that Steffen is referring to is https://github.com/dotnet/fsharp/issues/12359

isaacabraham:
@ninjarobot any objections to us moving back to targeting a single TFM (netstandard2) rather than both NS2 and Net5? It adds around 50% to the build time.

isaacabraham:
@ninjarobot at the time it was to remove the dependency on System.Text.Json (which is included in .NET 5) but in reality I'm not sure it's needed. Perhaps at some point in the future we'll move off of NS2 but I don't see that happening anytime soon.

isaacabraham:
@ninjarobot could you merge from master into this branch? another merge conflict and I don't know enough about the YAML to confidently do the merge myself - sorry :-(

isaacabraham:
I'm actually not sure we really need to wait on the `task { }` fix as we don't use any asynchronicity anywhere in the codebase.

isaacabraham:
Thank you!

PR #814, opened 2021-11-11
Adds GPU support for Container Instances

### LINKED ISSUE #794: GPU support for container instance
For azure container instances it is possible to specify gpu for containers:

https://docs.microsoft.com/en-us/azure/container-instances/container-instances-gpu

```json
{
    "$schema": "https://schema.management.azure.com/schemas/2015-01-01/deploymentTemplate.json#",
    "contentVersion": "1.0.0.0",
    "parameters": {
      "containerGroupName": {
        "type": "string",
        "defaultValue": "gpucontainergrouprm",
        "metadata": {
          "description": "Container Group name."
        }
      }
    },
    "variables": {
      "containername": "gpucontainer",
      "containerimage": "mcr.microsoft.com/azuredocs/samples-tf-mnist-demo:gpu"
    },
    "resources": [
      {
        "name": "[parameters('containerGroupName')]",
        "type": "Microsoft.ContainerInstance/containerGroups",
        "apiVersion": "2019-12-01",
        "location": "[resourceGroup().location]",
        "properties": {
            "containers": [
            {
              "name": "[variables('containername')]",
              "properties": {
                "image": "[variables('containerimage')]",
                "resources": {
                  "requests": {
                    "cpu": 4.0,
                    "memoryInGb": 12.0,
                    "gpu": {
                        "count": 1,
                        "sku": "V100"
                  }
                }
              }
            }
          }
        ],
        "osType": "Linux",
        "restartPolicy": "OnFailure"
        }
      }
    ]
}
```

It would be nice to be able to create this with farmer aswell or do I missed something and it is already possible?

brase:
This PR closes #794 

* adds gpu to resources of container instance

### DIFF
--- a/src/Farmer/Arm/ContainerInstance.fs
+++ b/src/Farmer/Arm/ContainerInstance.fs
@@ -26,6 +26,15 @@ type ImageRegistryAuthentication =
 /// Credentials for the container registry will be listed by ARM expression.
 | ListCredentials of ResourceId
 
+type Gpu =
+    { Count: int
+      Sku: ContainerInstanceGpu.Sku } 
+    member internal this.JsonModel =
+        {|
+            count = this.Count
+            sku = string this.Sku
+        |}
+
 /// Defines a command or HTTP request to get the status of a container.
 type ContainerProbe =
     { Exec : string list
@@ -68,6 +77,7 @@ type ContainerGroup =
            Ports : uint16 Set
            Cpu : float
            Memory : float<Gb>
+           Gpu : Gpu option
            EnvironmentVariables: Map<string, EnvVar>
            VolumeMounts : Map<string,string>
            LivenessProbe : ContainerProbe option
@@ -168,7 +178,9 @@ type ContainerGroup =
                                       resources =
                                        {| requests =
                                            {| cpu = container.Cpu
-                                              memoryInGB = container.Memory |}
+                                              memoryInGB = container.Memory
+                                              gpu = container.Gpu |> Option.map (fun g -> g.JsonModel |> box) |> Option.defaultValue null
+                                           |}
                                        |}
                                       volumeMounts =
                                           container.VolumeMounts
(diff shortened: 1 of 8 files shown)


src/Farmer/Arm/ContainerInstance.fs
```diff
@@ -26,6 +26,15 @@ type ImageRegistryAuthentication =
 /// Credentials for the container registry will be listed by ARM expression.
 | ListCredentials of ResourceId
 
+type Gpu =
+    { Count: int
```
ninjarobot, line 30: Are floating point values supported here? With CPU's, this allows you to balance the CPU priority between many container instances in the same container group by giving different containers a fractional core. Do GPUs use that same model for sharing and prioritizing between containers in the container group?

src/Farmer/Common.fs
```diff
@@ -684,6 +684,12 @@ module BingSearch =
         | S8
         | S9
 
+module ContainerInstanceGpu =
```
ninjarobot, line 687: It would be good to move this under the existing `ContainerGroup` module to keep from having to open additional modules. Then it could just be a module named `Gpu` for any types that are specific to container group GPU's. Then usage remains simple:

```fsharp
open Farmer.ContainerGroup

containerInstanceGpu {
    sku Gpu.V100
}
```

ninjarobot (changes requested):
Thanks for implementing ACI GPU support. The API specs and docs are a bit lacking on both CPU and GPU, but I know from usage that CPU accepts partial cores in order to prioritize CPU between containers, and I wonder if GPU does the same, or if you just are only allowed one GPU container instance at a time. If it is possible for you to verify whether this value must be an integer?

My only other request is that the GPU types go in a module under Farmer.ContainerGroup so users won't have to open multiple modules to use it.

src/Farmer/Arm/ContainerInstance.fs
brase: I found out that there is only 1, 2 or 4 allowed. Should we model this differently?

image

ninjarobot: If that's the case, modeling it as `int` is good.

ninjarobot (approved):
Thanks for checking on the SKU's and making the namespace adjustment. This looks great!

PR #818, opened 2021-11-14
Add CustomerId to LogAnalytics

forki:
This seems to be needed for azure container apps. See https://github.com/jeffhollan/container-apps-dotnet-eventing/blob/main/deploy/environment.bicep#L29
I have read the [contributing guidelines](CONTRIBUTING.md) and have completed the following:

* [x] **Tested my code** end-to-end against a live Azure subscription.
* [x] **Updated the documentation** in the docs folder for the affected changes.
* [ ] **Written unit tests** against the modified code that I have made.
* [ ] **Updated the [release notes](RELEASE_NOTES.md)** with a new entry for this PR.
* [x] **Checked the coding standards** outlined in the [contributions guide](CONTRIBUTING.md) and ensured my code adheres to them.

If I haven't completed any of the tasks above, I include the reasons why here:

Issue #809, opened 2021-11-03
Support Azure Container Apps


forki:
Example ARM: https://docs.microsoft.com/de-de/azure/container-apps/microservices-dapr-azure-resource-manager?tabs=bash#create-azure-resource-manager-arm-templates
Example for Scaling: https://docs.microsoft.com/de-de/azure/container-apps/scale-app#event-driven

forki:
API spec: https://docs.microsoft.com/en-us/azure/container-apps/azure-resource-manager-api-spec

isaacabraham:
Will probably need a full list of the KEDA scaling rules e.g. it looks like queue-based rules can be many implementations of which service bus is only one.

isaacabraham:
But we could start simple just HTTP / Mem / CPU rules to start.

Would you have any interest in providing the building block for this - even the "anonymous record / function" stage of building a supported resource; we could then convert this into a full thing.

forki:
>even the "anonymous record / function" stage of building a supported resource

working on it

forki:
Sneak Peek
![image](https://user-images.githubusercontent.com/57396/141787178-afad0b71-57c4-4977-9482-979add822aaa.png)

isaacabraham:
Coming along nicely! I think that you might want to consider closer integration of the different components (in the same way that we do for k8s containers or app service) e.g.:

1. Ability to auto-create an environment without supplying one, and / or ability to create an environment with a collection of container apps e.g.

```fsharp
let x =
    containerAppEnvironment {
        add_apps [
            containerApp {
                ....
            }
        ]
    }
````

2. Helper keywords to connect other azure resources e.g. `connect_servicebus_queue` which does the boilerplate in a consistent, well-defined way using a strongly-typed service bus queue config value.
3. Consider changing scaling to `add_scale_rule` which takes in a DU of the scale type.
4. Parameters should never normally be required to be explicitly supplied e.g. in webapp, we have `add_secret_setting` (or something like that) in which you just specify the name of the setting and we automatically create a parameter based on a naming convention (basically what you've done here).
5. Consider min/max replicas as a tuple rather than two keywords? I can't imagine many people will only want to set only one.

forki:
how can I get the outer containerAppEnvironment ID?

isaacabraham:
[...]

PR #824, opened 2021-11-15
Containerapps

forki:
This PR references #809 and subsumes #818

### DIFF
--- a/src/Farmer/Arm/Web.fs
+++ b/src/Farmer/Arm/Web.fs
@@ -459,10 +459,10 @@ module ContainerApp =
                                            for cred in this.ImageRegistryCredentials do
                                                match cred with
                                                | ImageRegistryAuthentication.Credential cred ->
-                                                   {| name = cred.Server
+                                                   {| name = cred.Username
                                                       value = cred.Password.ArmExpression.Eval() |}
                                                | ImageRegistryAuthentication.ListCredentials resourceId ->
-                                                   {| name = ArmExpression.create($"reference({resourceId.ArmExpression.Value}, '2019-05-01').loginServer").Eval()
+                                                   {| name = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').username").Eval()
                                                       value = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').passwords[0].value").Eval() |}
                                            for setting in this.Secrets do
                                                {| name = setting.Key.Value
@@ -478,11 +478,11 @@ module ContainerApp =
                                                | ImageRegistryAuthentication.Credential cred ->
                                                    {| server = cred.Server
                                                       username = cred.Username
-                                                      passwordSecretRef = cred.Password.ArmExpression.Eval() |}
+                                                      passwordSecretRef = cred.Username |}
                                                | ImageRegistryAuthentication.ListCredentials resourceId ->
                                                    {| server = ArmExpression.create($"reference({resourceId.ArmExpression.Value}, '2019-05-01').loginServer").Eval()
                                                       username = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').username").Eval()
-                                                      passwordSecretRef = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').passwords[0].value").Eval() |}
+                                                      passwordSecretRef = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').username").Eval() |}
                                        |]
                                        ingress =
                                            match this.IngressConfig with
--- a/src/Farmer/Arm/Web.fs
+++ b/src/Farmer/Arm/Web.fs
@@ -4,18 +4,19 @@ module Farmer.Arm.Web
 open Farmer
 open Farmer.Identity
 open Farmer.WebApp
-open Farmer.Identity
 open System
 
 let serverFarms = ResourceType ("Microsoft.Web/serverfarms", "2018-02-01")
 let sites = ResourceType ("Microsoft.Web/sites", "2020-06-01")
 let config = ResourceType ("Microsoft.Web/sites/config", "2016-08-01")
 let sourceControls = ResourceType ("Microsoft.Web/sites/sourcecontrols", "2019-08-01")
-let staticSites = ResourceType("Microsoft.Web/staticSites", "2019-12-01-preview")
-let siteExtensions = ResourceType("Microsoft.Web/sites/siteextensions", "2020-06-01")
+let staticSites = ResourceType ("Microsoft.Web/staticSites", "2019-12-01-preview")
+let siteExtensions = ResourceType ("Microsoft.Web/sites/siteextensions", "2020-06-01")
 let slots = ResourceType ("Microsoft.Web/sites/slots", "2020-09-01")
-let certificates = ResourceType("Microsoft.Web/certificates", "2019-08-01")
-let hostNameBindings = ResourceType("Microsoft.Web/sites/hostNameBindings", "2020-12-01")
+let certificates = ResourceType ("Microsoft.Web/certificates", "2019-08-01")
+let hostNameBindings = ResourceType ("Microsoft.Web/sites/hostNameBindings", "2020-12-01")
+let containerApps = ResourceType ("Microsoft.Web/containerApps", "2021-03-01")
+let kubeEnvironments = ResourceType ("Microsoft.Web/kubeEnvironments", "2021-02-01")
 
 let private mapOrNull f = Option.map (Map.toList >> List.map f) >> Option.defaultValue Unchecked.defaultof<_>
 
@@ -339,7 +340,7 @@ type StaticSite =
         member this.SecureParameters = [
             this.RepositoryToken
         ]
-type SslState = 
+type SslState =
     | SslDisabled
     | SniBased of thumbprint: ArmExpression
 
@@ -348,25 +349,25 @@ type HostNameBinding =
       SiteId: LinkedResource
       DomainName: string
       SslState: SslState }
-        member this.SiteResourceId = 
-            match this.SiteId with 
+        member this.SiteResourceId =
+            match this.SiteId with
             | Managed id -> id.Name
             | Unmanaged id -> id.Name
         member this.ResourceName =
             this.SiteResourceId / this.DomainName
-        member this.Dependencies = 
+        member this.Dependencies =
             [ match this.SiteId with
               | Managed resid -> resid
               | _ -> () ]
-        member this.ResourceId = 
+        member this.ResourceId =
             hostNameBindings.resourceId (this.SiteResourceId, ResourceName this.DomainName)
         interface IArmResource with
             member this.ResourceId = hostNameBindings.resourceId this.ResourceName
             member this.JsonModel =
                 {| hostNameBindings.Create(this.ResourceName, this.Location, this.Dependencies) with
                     properties =
-                        match this.SslState with 
-                        | SniBased thumbprint -> 
+                        match this.SslState with
+                        | SniBased thumbprint ->
                             {| sslState = "SniEnabled"
                                thumbprint = thumbprint.Eval() |} :> obj
                         | SslDisabled -> {| |} :> obj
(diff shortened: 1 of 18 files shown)


forki:
Looks like this right now:

    let serviceBusContainerApp containerName serviceBusQueueName =
        containerApp {
            name containerName
            activeRevisionsMode ActiveRevisionsMode.Single
            docker_image containerRegistryDomain containerRegistry containerName version
            replicas 0 1
            setting "ServiceBusQueueName" serviceBusQueueName
            secret_setting "StorageConnectionKey"
            secret_setting "EventsHubConnectionKey"
            secret_setting "ServiceBusConnectionKey"
            add_scale_rule "sb-keda-scale" (ScaleRuleType.ServiceBus {| QueueName = serviceBusQueueName; MessageCount = 5; SecretRef = "servicebusconnectionkey" |})
        }

    let httpContainerApp containerName =
        containerApp {
            name containerName
            activeRevisionsMode ActiveRevisionsMode.Single
            docker_image containerRegistryDomain containerRegistry containerName version
            replicas 1 5
            secret_setting "StorageConnectionKey"
            secret_setting "EventsHubConnectionKey"
            secret_setting "ServiceBusConnectionKey"
            add_secretref_variable "SecurityToken" "security-token"
            ingress { External = true; TargetPort = 80; Transport = "auto" }
            dapr { AppId = containerName }
            add_secrets [
                {
                    Name = "security-token"
                    Value = $"[parameters('SecurityToken-{containerName}')]"
                }
            ]
            add_scale_rule "http-rule" (ScaleRuleType.Http {| ConcurrentRequests = 100 |})
        }


    let ecarEnv =
        containerEnvironment {
            name  "ecarenv"
            logAnalytics ecarLogs
            add_containers [
                serviceBusContainerApp "reports" "salesreport"
                serviceBusContainerApp "ocpiimports" "ocpiimport"
                httpContainerApp "transactions"
                httpContainerApp "dimoco"
            ]
        }

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,428 @@
...
+      Secrets : Secret list
+      ActiveRevisionsMode : ActiveRevisionsMode
+      Resources : Resources option
+      Ingress : Ingress option
+      ScaleRules : ScaleRule list
+      MinReplicas : int
+      MaxReplicas : int
+      Settings : Map<string, Setting>
+      DaprSettings : DaprSettings option
+      EnvironmentVariables : EnvironmentVariable list
```
ninjarobot, line 63: It would be good to use the `EnvVar` from [Common](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Common.fs#L93). Here's how container instances use them [for IParameters](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Arm/ContainerInstance.fs#L127-L132) and [in the JsonModel](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Arm/ContainerInstance.fs#L167-L174).

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,428 @@
...
+                location = this.Location.ArmValue
+                properties =
+                    {|
+                        kubeEnvironmentId = this.ContainerEnvironment.Value.Eval()
+                        configuration =
+                            {|
+                                secrets = [|
+                                    yield
+                                        {|
+                                            name = $"container-registry-password-for-{containerRegistry}"
```
ninjarobot, line 95: This `containerRegistry` is referring to the `ContainerRegistryBuilder` instance instead of something in this `IArmResource`. I think it will need some collection of one or more registries to include along with builder syntax to add them - maybe this can also [match with what's in container instances](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Builders/Builders.ContainerGroups.fs#L235-L242).

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,428 @@
...
+                                activeRevisionsMode =
+                                    match this.ActiveRevisionsMode with
+                                    | ActiveRevisionsMode.Single -> "Single"
+                                    | ActiveRevisionsMode.Multiple -> "Multiple"
+                                registries =
+                                    [|
+                                        {|
+                                            server = containerSettings.RegistryDomain
+                                            username = containerSettings.RegistryName
+                                            passwordSecretRef = $"container-registry-password-for-{containerSettings.RegistryName}"
```
ninjarobot, line 109: We might want to do [similar to ContainerInstances here](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Arm/ContainerInstance.fs#L219-L230), too, because you can add secure parameters with a registry credential or reference a registry and it will generate an expression to get the registry's admin credential.

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,428 @@
...
+        { state with Ingress = Some ingress }
+
+    /// Sets the dapr settings of the Azure Container App.
+    [<CustomOperation "dapr">]
+    member _.SetDapr(state: ContainerAppConfig, dapr:DaprSettings) =
+        { state with DaprSettings = Some dapr }
+
+    /// Sets the replicas settings of the Azure Container App.
+    [<CustomOperation "replicas">]
+    member _.SetReplicas(state: ContainerAppConfig, minReplicas:int, maxReplicas: int) =
```
ninjarobot, line 298: Would it make sense to have two operations here - `min_replicas` and `max_replicas`? It looks like these aren't required, and there are probably scenarios where people would only want to set one of those two values.

ninjarobot (commented):
Thanks for adding all of this. I have a few suggestions to see if we can keep the usage consistent with other resources, especially for things like secrets that would be passed as parameters.

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,428 @@
...
+                | ParameterSetting s -> Some s
+                | ExpressionSetting _
+                | LiteralSetting _ ->
+                    None)
+
+    interface IArmResource with
+        member this.ResourceId = containerAppResourceType.resourceId this.Name
+        member this.JsonModel =
+            let containerSettings = this.DockerImage.Value
+            {|  name = this.Name.Value
```
ninjarobot, line 82: This can be simplified a little by using the type to generate those top level properties in the model:

```fsharp
{| containerAppResourceType.Create(this.Name, this.Location) with
   properties = ...
```

[Concise example](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Arm/Bastion.fs#L23).

src/Farmer/Builders/Builders.ContainerApps.fs
forki: I had that. But @isaacabraham proposed to do it this way

src/Farmer/Builders/Builders.ContainerApps.fs
forki: well the problem is that EnvVar does not support secretRefs

src/Farmer/Builders/Builders.ContainerApps.fs
forki: done

src/Farmer/Builders/Builders.ContainerApps.fs
ninjarobot: I see...they made a new thing that refers to [Kubernetes secrets](https://kubernetes.io/docs/concepts/configuration/secret/) instead of secrets coming in from ARM parameters. Sorry for my confusion while I learn about these :)

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,425 @@
...
+    member _.SetDockerImage(state:ContainerAppConfig, registryDomain, registryName, containerName, version) =
+        { state with
+            DockerImage =
+                Some {| RegistryDomain = registryDomain
+                        RegistryName = registryName
+                        ContainerName = containerName
+                        Version = version |} }
+
+    /// Sets the environment of the Azure Container App.
+    [<CustomOperation "activeRevisionsMode">]
```
ninjarobot, line 312: By convention everything in Farmer uses `snake_case` for custom operations.

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,425 @@
...
+        | None -> raiseFarmer $"The LogAnalytics connections was not set. Please use the logAnalytics function of the containerEnvironment builder."
+        | _ ->
+            state
+
+    /// Sets the name of the Azure Container App Environment.
+    [<CustomOperation "name">]
+    member _.Name(state: ContainerEnvironmentConfig, name:string) = { state with Name = ResourceName name }
+
+    /// Sets the environment of the Azure Container App.
+    [<CustomOperation "logAnalytics">]
```
ninjarobot, line 402: This should also be `snake_case`.

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -0,0 +1,425 @@
...
+    member _.SetLogAnalytics(state: ContainerEnvironmentConfig, logAnalytics:WorkspaceConfig) =
+        { state with LogAnalytics = Some logAnalytics }
+
+    /// Sets the InternalLoadBalancerEnabled property of the Azure Container App Environment.
+    [<CustomOperation "internalLoadBalancerEnabled">]
```
ninjarobot, line 407: This should also be `snake_case`.

isaacabraham:
@forki thanks for doing this. I've made some (fairly significant) changes, some of which it would be good to get your feedback on.

1. I've separated out the ARM and Builder configuration types per the rest of Farmer - this eliminates the need for some of the atypical patterns in this builder like optional types and such.
2. I've moved some of the strings into proper types (like Ingress).
3. I've separated out some elements into specific keywords instead of requiring a record for all of them at once.
4. Support for creating a LA Workspace automatically.
5. Support for public docker images.
6. Simplified the code around secrets, environment vars and settings (although I'm not convinced I've gotten it exactly correct).

forki:
not really sure how to test this. Is there a waay to add a custom build to my project or something?

src/Farmer/Arm/Web.fs
```diff
@@ -402,3 +403,241 @@ module SiteExtensions =
...
+                               configuration =
+                                   {|
+                                       secrets = [|
+                                           match this.DockerImage with
+                                           | PublicImage _ ->
+                                               ()
+                                           | PrivateImage image ->
+                                               // TODO: This looks suspicious
+                                               {| name = ContainerApp.ContainerRegistryPassword image.RegistryName
+                                                  value = ContainerApp.DockerPassword(image.RegistryName).ArmExpression.Eval() |}
```
ninjarobot, line 460: It might be useful to get the password for the container registry by ARM expression as long as they have the admin credential enabled on it.

```
"[listCredentials(resourceId('Microsoft.ContainerRegistry/registries','mycontainerregistry'),'2019-05-01').passwords[0].value]
```

This is a configuration member on container registries, so you an get this expression with `myContainerRegistry.Password`.

src/Farmer/Common.fs
```diff
@@ -2067,6 +2084,32 @@ module AvailabilityTest =
...
+        | Http of HttpScaleRule
+        | CPU of CpuScaleRule
+        | Memory of MemoryScaleRule
+        | StorageQueue of StorageQueueScaleRule
+        | Custom of obj
+    type Transport = HTTP1 | HTTP2 | Auto
+    type Visibility = External | Internal
+    type ActiveRevisionsMode = Single | Multiple
+    type DockerImageKind =
+        | PrivateImage of {| RegistryDomain : string; RegistryName : string; ContainerName : string; Version:string |}
```
ninjarobot, line 2108: The `ContainerName` should be a `string option`. You may have either of these.

myregistry.azurecr.io/somerepo/somecontainer:1.0.0
myregistry.azurecr.io/somerepo:1.0.0

isaacabraham: How should we expose this in the CE?

isaacabraham:
@ninjarobot I think we're pretty close here. I have some ideas on further enhancements but I think that they can be added as a second pass:

1. Add more "Farmer native" integrations for scale rules e.g. Service Bus and Event Hub scale rules.
2. Add more KEDA scale rules.
3. Make it easier to create "simple" containers - currently to create a single container from e.g. a public container, you need to make an environment, a container app and then a container. I would like to see some helper methods on e.g. the container app  that would create a container in a single line - obviously without all the config of the full builder e.g. `create_public_container` or something.

Is there a separate resource type for Containers themselves? We have the `kubeEnvironment` and `containerApps` but currently the individual containers are bundled with the container app. No problem with that if that's how it is, but if we can create separate ARM resources for each, that would follow the pattern we've used for other multi-level resources e.g. server bus etc.

PR #826, opened 2021-11-16
Add dotnet-isolated function support for linux

### LINKED ISSUE #825: Set linuxFxVersion for isolated function apps
When creating an isolated function to run on Linux, `linuxFxVersion` needs to be set to `DOTNET-ISOLATED|5.0`

Currently I am setting this manually after the deployment via command line using 
```az functionapp config set --name $appName --resource-group $resourceGroup --linux-fx-version "`"DOTNET-ISOLATED|5.0`""```

### PR DESCRIPTION
**Pull request by TheRSP**
This PR closes #825

* Update FunctionRuntime DU to include DotnetCore and version details for all .NET variations [BREAKING]
* Set WebApp.LinuxFxVersion for function apps

This is a bug fix with no change to the public API surface so no documentation change seems necessary.

Note that I have only added support for .NET isolated functions on linux as this what I have the means to test. I have left the situation unchanged for non-.NET functions

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
+* Functions: fix for .Net isolated functions hosted on linux
 
 ## 1.6.21
 * CDN: Adds new SKU types for Azure Front Door Standard/Premium
--- a/src/Farmer/Builders/Builders.Functions.fs
+++ b/src/Farmer/Builders/Builders.Functions.fs
@@ -252,7 +252,7 @@ type FunctionsConfig =
                       match this.VersionedRuntime with
                       | DotNet, Some version -> 
                         match Double.TryParse(version) with
-                        | true, versionNo when versionNo < 5 -> Some $"DOTNETCORE|{version}"
+                        | true, versionNo when versionNo < 4.0 -> Some $"DOTNETCORE|{version}"
                         | _ -> Some $"DOTNET|{version}"
                       | DotNetIsolated, Some version -> Some $"DOTNET-ISOLATED|{version}"
                       | _, Some version -> Some $"{functionsRuntime.ToUpper()}|{version}"
--- a/src/Farmer/Builders/Builders.Functions.fs
+++ b/src/Farmer/Builders/Builders.Functions.fs
@@ -12,18 +12,16 @@ open Farmer.Arm.KeyVault
 open Farmer.Arm.KeyVault.Vaults
 open System
 
-type FunctionsRuntime = 
-    | DotNetCore of version: string 
-    | DotNet of version: string 
-    | DotNetIsolated of version: string
-    | Node 
-    | Java 
-    | Python
-    static member DotNetCore31 = DotNetCore "3.1"
-    static member DotNet50 = DotNet "5.0"
-    static member DotNet50Isolated = DotNetIsolated "5.0"
-    static member DotNet60 = DotNet "6.0"
-    static member DotNet60Isolated = DotNetIsolated "6.0"
+type FunctionsRuntime = DotNetCore | DotNet | DotNetIsolated | Node | Java | Python
+type VersionedFunctionsRuntime =  FunctionsRuntime * string option
+type FunctionsRuntime with
+    // These values are defined on FunctionsRuntime to reduce the need for users to be aware of the distinction 
+    // between FunctionsRuntime and VersionedFunctionsRuntime as well as to provide parity with WebApp runtime
+    static member DotNetCore31 = DotNetCore, Some "3.1"
+    static member DotNet50 = DotNet, Some "5.0"
+    static member DotNet50Isolated = DotNetIsolated, Some "5.0"
+    static member DotNet60 = DotNet,Some "6.0"
+    static member DotNet60Isolated = DotNetIsolated, Some"6.0"
 
 type DockerInfo = {
     User: string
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: src/Farmer/Builders/Builders.Functions.fs [hunk 1]
```diff
@@ -12,7 +12,19 @@ open Farmer.Arm.KeyVault
 open Farmer.Arm.KeyVault.Vaults
 open System
 
-type FunctionsRuntime = DotNet | DotNetIsolated | Node | Java | Python
+type FunctionsRuntime = 
+    | DotNetCore of version: string
```
ninjarobot, line 16: Adding the `version` here is a breaking change for anyone who currently specifies their runtime settings (anywhere `use_runtime` is used). This has been a problematic area since Microsoft has added more .NET hosting options over time, so maybe there should be a new DSL operation like `use_dotnet_runtime` for specifying any of the .NET runtime hosting options.

ninjarobot (changes requested):
Thanks for adding these options for improved linux support. The proposed changes would be breaking, so can you think of a way to specify these options without breaking existing users? Also, please keep in mind that V3 doesn't work with .NET 6.0, so we we need to be wary of those compatibility options.

src/Farmer/Builders/Builders.Functions.fs
TheRSP: Thanks for the prompt. I think I've found a reasonable solution to the breaking changes.

src/Farmer/Builders/Builders.Functions.fs
```diff
@@ -12,7 +12,17 @@ open Farmer.Arm.KeyVault
 open Farmer.Arm.KeyVault.Vaults
 open System
 
-type FunctionsRuntime = DotNet | DotNetIsolated | Node | Java | Python
+type FunctionsRuntime = DotNetCore | DotNet | DotNetIsolated | Node | Java | Python
```
ninjarobot, line 15: We shouldn't add `DotNetCore` here, as it's not a different runtime like Node, Java, and Python. This is only needed for the `LinuxFxVersion` so it could be simply if the version is < 4, emit `DOTNETCORE`.

src/Farmer/Builders/Builders.Functions.fs
```diff
@@ -223,7 +235,15 @@ type FunctionsConfig =
                   HTTP20Enabled = None
                   ClientAffinityEnabled = None
                   WebSocketsEnabled = None
-                  LinuxFxVersion = None
+                  LinuxFxVersion = 
+                    match this.CommonWebConfig.OperatingSystem with
+                    | Windows -> None
+                    | Linux ->
+                      match this.VersionedRuntime with
+                      | DotNetCore, Some version  -> Some $"DOTNETCORE|{version}"
```
ninjarobot, line 243: This really is the only place we need to worry with `DOTNETCORE`, so this is where the logic could go to emit `DOTNETCORE` if using a lower version of the runtime.

src/Tests/Functions.fs
```diff
@@ -108,11 +108,13 @@ let tests = testList "Functions tests" [
         Expect.equal secrets.[1].Value (ExpressionSecret sa.Key) "Incorrect secret value"
         Expect.sequenceEqual secrets.[1].Dependencies [ vaults.resourceId "testfuncvault"; storageAccounts.resourceId "teststorage" ] "Incorrect secret dependencies"
     }
-
+    
     test "Supports dotnet-isolated runtime" {
-        let f = functions { name "func"; use_runtime (FunctionsRuntime.DotNetIsolated) }
+        let f = functions { name "func"; use_runtime (FunctionsRuntime.DotNet50Isolated); operating_system Linux }
```
ninjarobot, line 113: Please add a new test for the Linux hosting scenario. The existing test should still pass.

It also would be good to have a test to ensure the correct value is emitted for LinuxFxVersion for DotNet 3.1 and DotNet 6.

src/Farmer/Builders/Builders.Functions.fs
```diff
@@ -223,7 +235,15 @@ type FunctionsConfig =
...
+                      | DotNet, Some version -> Some $"DOTNET|{version}"
+                      | DotNetIsolated, Some version -> Some $"DOTNET-ISOLATED|{version}"
+                      | _ -> None
```
ninjarobot, line 246: Maybe this should support all the runtimes instead of emitting None for any of the others.

```fsharp
| runtime, Some version -> Some $"{(functionsRuntime runtime).ToUpper()}|{version}"
```

TheRSP:
I have added all the runtimes returned by `az webapp list-runtimes --linux` for me

src/Tests/Functions.fs
```diff
@@ -108,14 +108,31 @@ let tests = testList "Functions tests" [
         Expect.equal secrets.[1].Value (ExpressionSecret sa.Key) "Incorrect secret value"
         Expect.sequenceEqual secrets.[1].Dependencies [ vaults.resourceId "testfuncvault"; storageAccounts.resourceId "teststorage" ] "Incorrect secret dependencies"
     }
-
+    
     test "Supports dotnet-isolated runtime" {
-        let f = functions { name "func"; use_runtime (FunctionsRuntime.DotNetIsolated) }
+        let f = functions { name "func"; use_runtime (FunctionsRuntime.DotNet50Isolated); }
```
ninjarobot, line 113: How can we avoid this breaking change for people?

RELEASE_NOTES.md
```diff
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
```
ninjarobot, line 3: This should be in the 1.6.21 release.

ninjarobot (changes requested):
Thanks for making the adjustments, but it's still a breaking change for anyone is using `use_runtime`. Maybe an overload is possible that could handle the mapping to a sane default.

src/Tests/Functions.fs
TheRSP: Sorry, I forgot to revert this change to the test but I have already made it an overload to prevent breaking existing code .

ninjarobot: I see...ok, I missed that, great!

ninjarobot (approved):

PR #837, opened 2021-12-05
Fix missing AutoSwapSlotName in json output.

### LINKED ISSUE #836: Bug: WebApp slots: AutoSwapSlotName is not added to json output
There is a setting for AutoSwapSlotName in the SlotBuilder but it is never used when writing the template json output.

My expectation is that is should be written to the json output so AutoSwapSlotName can be configured.

I have made a fix locally and it works fine to set the AutoSwapSlotName and the deployed slot will correctly autoswap in azure. The only caveat seems to be that if you then remove AutoSwapSlotName from the template the setting will not switch back to AutoSwap enabled = Off, i.e. to disable the feature I had to manually turn it off in the azure portal. 
But this seems like a bug or limitation in ARM for this specific setting.

I'll submit a PR with my fix.

viktorvan:
This PR closes #836

* Add autoSwapSlotName to JsonModel for WebApp.

* The SlotBuilder does not seem to be documented.

### DIFF
--- a/src/Farmer/Arm/Web.fs
+++ b/src/Farmer/Arm/Web.fs
@@ -287,10 +287,7 @@ type Site =
                                            supportCredentials = credentials |> Option.toNullable |})
                             |> Option.toObj
                            healthCheckPath = this.HealthCheckPath |> Option.toObj
-                           autoSwapSlotName =
-                               match this.AutoSwapSlotName with
-                               | None -> null
-                               | Some name -> name
+                           autoSwapSlotName = this.AutoSwapSlotName |> Option.toObj
                         |}
                     |}
             |} :> _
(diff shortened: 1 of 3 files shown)


src/Farmer/Arm/Web.fs
```diff
@@ -287,6 +287,10 @@ type Site =
                                            supportCredentials = credentials |> Option.toNullable |})
                             |> Option.toObj
                            healthCheckPath = this.HealthCheckPath |> Option.toObj
+                           autoSwapSlotName =
+                               match this.AutoSwapSlotName with
```
isaacabraham, line 291: Replace this match with `Option.toObj`

isaacabraham (commented):
Tiny fix and this can then go in. Interesting that there's no docs on this apparently.

PR #850, opened 2021-12-15
Checkout submodules and include test coverage reports in CI builds

ninjarobot:
CI changes only:

* Checkout submodules in CI builds so SourceLink no longer warns it cannot find some of the source.
* Include test coverage reports.


isaacabraham:
That is awesome!! Great stuff

PR #852, opened 2021-12-19
Target netstandard2 only.

isaacabraham:
This PR closes #

* Revert to targetting netstandard2 only. Improves build time by ~30%.

No changes aside from the project file. Unit tests still all pass.

### DIFF
--- a/src/Farmer/Farmer.fsproj
+++ b/src/Farmer/Farmer.fsproj
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
   <PropertyGroup>
     <!-- General -->
     <AssemblyName>Farmer</AssemblyName>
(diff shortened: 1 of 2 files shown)


src/Farmer/Farmer.fsproj
```diff
@@ -41,8 +41,6 @@
   <ItemGroup>
     <PackageReference Update="FSharp.Core" Version="5.0.0" />
     <PackageReference Include="Microsoft.SourceLink.GitHub" Version="1.0.0" PrivateAssets="All" />
-  </ItemGroup>
-  <ItemGroup Condition=" '$(TargetFramework)' == 'netstandard2.0' ">
     <PackageReference Include="System.Text.Json" Version="5.0.1" />
```
ninjarobot, line 44: Should this just be `Version="5"`? Then (I think) when used in net5.0 and net6.0 projects or scripts it will use the one included in with the runtime.

ninjarobot (approved):
I agree, as a library, netstandard2.0 should cover all use cases.

src/Farmer/Farmer.fsproj
isaacabraham: OK - if that's the case, let's do that - that would be even better, yes.

isaacabraham: Done.

PR #853, opened 2021-12-19
Lots of changes to Container Apps API + docs.

isaacabraham:

* Clean up Container Apps docs
* Separate keywords for each Scale Rule (=> cleaner syntax)
* Sample script for container apps.
* Simplified Ingress mode.
* Ability to create a simple container on a container app without requiring a full container builder.

I've left the end-to-end tests out for now - getting feedback on the API changes would be welcome first. Same for the release notes (especially if this goes into the next release).

### DIFF
--- a/docs/content/api-overview/resources/application-gateway.md
+++ b/docs/content/api-overview/resources/application-gateway.md
@@ -2,7 +2,7 @@
 title: "Application Gateway"
 date: 2021-10-23T10:47:02-05:00
 chapter: false
-weight: 5
+weight: 1
 ---
 
 #### Overview
@@ -140,7 +140,7 @@ let myAppGateway =
             backend_address_pool backendPool
             backend_http_settings backendSettings
         }
-    
+
     appGateway {
         name "app-gw"
         sku_capacity 2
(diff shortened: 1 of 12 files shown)


ninjarobot (approved):
All the changes seem fine. They change the DSL so I think we should ship this right away before anyone uses the prior version.

isaacabraham:
OK, great. I'll do a couple of end-to-ends and feed back once that's done.

isaacabraham:
I was also thinking about releasing "experimental" / "incubation" builders in a different namespace e.g. `Farmer.Builders.Experimental`? That way we could release new Builders with people able to opt in to use them and give feedback, whilst they are still subject to change.

isaacabraham:
Thank you! Let's do it.

isaacabraham:
I've found one tiny issue - I set CPU and Memory resources to optional, thinking that ARM would set minimum defaults. It doesn't. Instead it gives an error that you set CPU to 0. I'm going to make another fix that changes this so it defaults to minimums (0.25 VCores, 0.5 Gb)

PR #862, opened 2022-01-18
Comp it main

### LINKED ISSUE #838: Support deployment slot settings
It is possible to set slot settings to be sticky which will keep it in the deployment slot after swapping deployment/prod. 
image

In ARM templates it is done using `slotConfigNames` of `Microsoft.Web/sites/config` resource. [Documentation](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/2020-06-01/sites/config-slotconfignames?tabs=json)

Proposed solution:

```
let myWebApp = webApp {
    name "test"
    add_slot deploymentSlot

    settings ["my_setting", "test value"
                  "my_slot_setting", "sticky value"]

    slot_setting_names ["my_slot_setting" ]
}
```

nargiz:
This PR closes #838

* Allows you to specify slot settings for web app/functions using slot_setting and slot_settings

```fsharp
```

### DIFF
--- a/docs/content/api-overview/resources/functions.md
+++ b/docs/content/api-overview/resources/functions.md
@@ -45,6 +45,8 @@ The Functions builder is used to create Azure Functions accounts. It abstracts t
 | publish_as | Specifies whether to publish function as code or as a docker container. |
 | add_slot | Adds a deployment slot to the app |
 | add_slots | Adds multiple deployment slots to the app |
+| slot_setting | Sets a deployment slot setting of the function in the form "key" "value". |
+| slot_settings | Sets a list of deployment slot setting of the function as tuples in the form of ("key", "value"). |
 | health_check_path | Sets the path to your functions health check endpoint, which Azure load balancers will ping to determine which instances are healthy.|
 
 #### Post-deployment Builder Keywords
--- a/src/Farmer/Builders/Builders.Functions.fs
+++ b/src/Farmer/Builders/Builders.Functions.fs
@@ -323,6 +323,12 @@ type FunctionsConfig =
                 { site with AppSettings = None; ConnectionStrings = None }
                 for (_, slot) in this.CommonWebConfig.Slots |> Map.toSeq do
                     slot.ToSite site
+                    
+            if this.CommonWebConfig.SlotSettingNames <> List.Empty then
+                {
+                    SiteName = this.Name.ResourceName;
+                    SlotSettingNames = this.CommonWebConfig.SlotSettingNames;
+                }
         ]
 
 type FunctionsBuilder() =
@@ -343,7 +349,8 @@ type FunctionsBuilder() =
               Slots = Map.empty
               WorkerProcess = None
               ZipDeployPath = None
-              HealthCheckPath = None }
+              HealthCheckPath = None
+              SlotSettingNames = List.Empty }
           StorageAccount = derived (fun config ->
             let storage = config.Name.ResourceName.Map (sprintf "%sstorage") |> sanitiseStorage |> ResourceName
             storageAccounts.resourceId storage)
--- a/src/Farmer/Builders/Builders.Functions.fs
+++ b/src/Farmer/Builders/Builders.Functions.fs
@@ -324,7 +324,7 @@ type FunctionsConfig =
                 for (_, slot) in this.CommonWebConfig.Slots |> Map.toSeq do
                     slot.ToSite site
                     
-            if this.CommonWebConfig.SlotSettingNames <> List.Empty then
+            if this.CommonWebConfig.SlotSettingNames <> Set.empty then
                 {
                     SiteName = this.Name.ResourceName;
                     SlotSettingNames = this.CommonWebConfig.SlotSettingNames;
@@ -350,7 +350,7 @@ type FunctionsBuilder() =
               WorkerProcess = None
               ZipDeployPath = None
               HealthCheckPath = None
-              SlotSettingNames = List.Empty }
+              SlotSettingNames = Set.empty }
           StorageAccount = derived (fun config ->
             let storage = config.Name.ResourceName.Map (sprintf "%sstorage") |> sanitiseStorage |> ResourceName
             storageAccounts.resourceId storage)
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -179,7 +179,7 @@ type CommonWebConfig =
       WorkerProcess : Bitness option
       ZipDeployPath : (string*ZipDeploy.ZipDeploySlot) option
       HealthCheckPath: string option
-      SlotSettingNames: List<string> }
+      SlotSettingNames: string Set }
 
 type WebAppConfig =
     { CommonWebConfig: CommonWebConfig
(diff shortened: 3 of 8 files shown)


src/Farmer/Builders/Builders.Functions.fs
```diff
@@ -323,6 +323,12 @@ type FunctionsConfig =
                 { site with AppSettings = None; ConnectionStrings = None }
                 for (_, slot) in this.CommonWebConfig.Slots |> Map.toSeq do
                     slot.ToSite site
+                    
+            if this.CommonWebConfig.SlotSettingNames <> Set.empty then
+                {
```
isaacabraham, line 328: Formatting - I actually do prefer this style but the rest of Farmer uses the more common record creation style, could you follow that? Also `;` unnecessary and can be removed.

docs/content/api-overview/resources/functions.md
```diff
@@ -45,6 +45,8 @@ The Functions builder is used to create Azure Functions accounts. It abstracts t
 | publish_as | Specifies whether to publish function as code or as a docker container. |
 | add_slot | Adds a deployment slot to the app |
 | add_slots | Adds multiple deployment slots to the app |
+| slot_setting | Sets a deployment slot setting of the function in the form "key" "value". |
```
isaacabraham, line 48: I wonder whether it's possible to piggy back on the existing `setting` keywords? Also, there's also the `secret_setting` keyword for adding secure settings - this should probably be migrated as well if we keep with separate keywords.

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -540,6 +541,12 @@ type WebAppConfig =
                   DomainName = customDomain
                   SslState = SslDisabled }
             | NoDomain -> ()
+            
+            if this.CommonWebConfig.SlotSettingNames <> Set.empty then
+                {
```
isaacabraham, line 546: As above re: record style.

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -888,3 +896,16 @@ module Extensions =
         [<CustomOperation "health_check_path">]
         /// Specifies the path Azure load balancers will ping to check for unhealthy instances.
         member this.HealthCheckPath(state:'T, healthCheckPath:string) = this.Map state (fun x -> {x with HealthCheckPath = Some(healthCheckPath)})
+        
+        /// Adds slot  settings
+        [<CustomOperation "slot_setting">]
+        member this.AddSlotSetting (state:'T, key, value) =
+            let current = this.Get state
+            { current with Settings = current.Settings.Add(key, LiteralSetting value); SlotSettingNames =current.SlotSettingNames.Add(key) }
```
isaacabraham, line 904: Newline for each updated field please.

isaacabraham: And space before and after `=`

isaacabraham (commented):
Thank you very much! I think the core of this looks good, but I'm not sure adding a new keyword is the right way to go (see review comments).

I know that there were some other PRs and issues about slots - @ninjarobot @TheRSP do they relate to this at all?

docs/content/api-overview/resources/functions.md
ninjarobot: How does this really differ from including config when adding a slot with `add_slot` and using the SlotBuilder with `addSlot`? I see this PR adds `slotconfignames` when you add these - is the purpose of this _just_ for sticky config?

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -888,3 +896,16 @@ module Extensions =
...
+            |> this.Wrap state
+        [<CustomOperation "slot_settings">]
+        member this.AddSlotSettings(state:'T, settings: (string*string) list) =
```
ninjarobot, line 907: If this is needed independently of adding slots, it at least should support the same capabilities as the settings in slots by using the `Setting` type instead of just strings. Then it can be a literal or an ARM expression.

src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -888,3 +896,16 @@ module Extensions =
...
         /// Specifies the path Azure load balancers will ping to check for unhealthy instances.
         member this.HealthCheckPath(state:'T, healthCheckPath:string) = this.Map state (fun x -> {x with HealthCheckPath = Some(healthCheckPath)})
+        
+        /// Adds slot  settings
+        [<CustomOperation "slot_setting">]
+        member this.AddSlotSetting (state:'T, key, value) =
+            let current = this.Get state
+            { current with Settings = current.Settings.Add(key, LiteralSetting value); SlotSettingNames =current.SlotSettingNames.Add(key) }
+            |> this.Wrap state
+        [<CustomOperation "slot_settings">]
```
ninjarobot, line 906: The convention is to use `add_` when adding items to the config, so this should be `add_slot_settings`.

src/Tests/Functions.fs
```diff
@@ -364,4 +364,63 @@ let tests = testList "Functions tests" [
             |> Seq.map (fun x-> x.ToObject<{|name:string;value:string|}> ())
         Expect.contains appSettings {|name="APPINSIGHTS_INSTRUMENTATIONKEY"; value="[reference(resourceId('shared-group', 'Microsoft.Insights/components', 'theName'), '2014-04-01').InstrumentationKey]"|} "Invalid value for APPINSIGHTS_INSTRUMENTATIONKEY"
     }
+    
+    test "Supports slot settings" {
+        let functionsApp = functions { name "test"; slot_settings [ "sticky_config", "sticky_config_value"; "another_sticky_config", "another_sticky_config_value" ]}
```
ninjarobot, line 369: Could any of these tests be enhanced to make it a practical example that would show usage along with the existing DSL to `add_slots`?

docs/content/api-overview/resources/functions.md
nargiz: To make settings sticky, their names should be specified in webapp(`sites`) child resource, `slotconfignames`. We usually add settings in webapp, so the idea was to have `slot_setting `which adds the setting and creates child resource `slotconfignames`. 

The alternative solution was something similar to this:

```
let myWebApp = webApp {
    name "test"
    add_slot deploymentSlot

    settings ["my_setting", "test value"
                  "my_slot_setting", "sticky value"]

    slot_setting_names ["my_slot_setting" ]
}
```
Where `slot_setting_names` was only marking the settings as sticky by adding `slotconfignames`, But we thought the current solution with `slot_setting` will make things clearer.

isaacabraham:
@nargiz I'm really sorry this hasn't been looked at for such a long time. We need to get this reviewed and merged in before it drifts too far from the main code base. @TheRSP @ninjarobot what do you both think about this - are there outstanding review issues that need to be resolved?

src/Farmer/Builders/Builders.WebApp.fs
TheRSP: Given that `setting` doesn't follow that convention, would it be best if operator aligned with `setting` or with other operators?

nargiz:
Sorry, I got a bit confused with the DSL requirements, how slot_settings should be named/represented and distracted with other things. Thanks for reminding me. I will check it by the end of the week.

martinbryant:
Hi @nargiz - can we look at getting the conflicts dealt with so it can be reviewed? Thanks

PR #867, opened 2022-01-23
Only turn on "auto" extension for Windows

### LINKED ISSUE #782: Can't create a Web App on Linux
I was trying to deploy a SAFE Stack application with Linux OS and was getting the following error:
```
No route registered for '/api/siteextensions/Microsoft.AspNetCore.AzureAppServices.SiteExtension'
```

Current work around:

```diff
let web = webApp {
        name "test"
        operating_system Linux
++      automatic_logging_extension false
    }
```

This is being handled in `Builders.WebApp.fs` but needs to be updated to include the OS as well:
```fsharp
{ state with
            SiteExtensions =
                match state with
                // its important to only add this extension if we're not using Web App for Containers - if we are
                // then this will generate an error during deployment:
                // No route registered for '/api/siteextensions/Microsoft.AspNetCore.AzureAppServices.SiteExtension'
                | { Runtime = Runtime.DotNetCore _; AutomaticLoggingExtension = true ; DockerImage = None } ->
                    state.SiteExtensions.Add WebApp.Extensions.Logging
                | _ ->
                    state.SiteExtensions
```

isaacabraham:
This PR closes #782

* Stop automatically turning on the logging extension if OS is Linux.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
+* WebApps/Functions: Don't turn on Logging Extension for Linux App Service.
 
 ## 1.6.25
 * CosmosDb: Add support for serverless capacity mode.
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -497,16 +497,16 @@ type WebAppConfig =
                 hostNameBinding
 
                 // Get the resource group which contains the app service plan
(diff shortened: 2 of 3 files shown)


ninjarobot (approved):
Makes sense - hopefully we do t have too many more deviations for Linux webapps or we may want to add a separate builder.

isaacabraham:
@ninjarobot possibly. I think as long as the deviations occur in terms of ARM generation rather than usage, it's fine. This is one where that's not the case, but I think that this is the exception to the rule.

PR #878, opened 2022-02-14
Escape deploy parameter values

### LINKED ISSUE #877: Escape parameters passed to az deployment
I'm trying to call the execute function with some parameters. It seems like parameter escaping is not being handled within the Deploy module and instead I have to explicitly escape the parameters myself before calling execute. 

It would be nice if the Deploy module handled parameter escaping for us so we don't need to be aware of how the parameters are being passed to the Azure powershell command and what type of escaping we need.

amypc:
This PR closes #877 

* Escape parameters passed to az deployment command

- No unit tests for deployment module
- Documentation for parameters not updated as this seems like it should be the default behaviour

```fsharp
let deployment = arm {
    add_resource (deploymentScript {
        name "thing"
        env_vars [
            EnvVar.createSecure "foo" "secret-foo"
        ]
    })
}
        
Deploy.execute "RESOURCE_GROUP_NAME" [
        "secret-foo", "Unescaped secret with special charcters"
    ]
    deployment
```


isaacabraham:
@amypc any risk of breaking anyone with this change?

amypc:
Hi @isaacabraham, yes this will be breaking for anyone passing in escaped parameters into the `tryValidate`, `tryWhatIf`, `tryExecute`, `execute` or `whatIf` function. The parameters would end up being escaped twice after this change. 

I've documented this in the release notes

RELEASE_NOTES.md
```diff
@@ -4,6 +4,9 @@ Release Notes
 * WebApp: Support custom port for docker container with `docker_port`
 * WebApps/Functions: Don't turn on Logging Extension for Linux App Service.
 
+## 1.6.27
+* Deployment Scripts: Escape parameters passed to the az deployment command (breaking change). Any previously escaped parameters need to be unescaped before passing to the execute or whatIf functions.
```
ninjarobot, line 8: This isn't a change to deploymentScripts resources, but rather to the Farmer wrappers for the Az CLI.

src/Farmer/Deploy.fs
```diff
@@ -106,10 +106,13 @@ module Az =
             | Validate -> "validate"
 
     let private deployOrValidate (deploymentCommand:DeploymentCommand) resourceGroup deploymentName templateFilename parameters =
+        let escapeParameterArguments = (fun (a, (b: string)) -> 
+            let escapedValue = b.Replace("\"","\\\"")
+            $"\"%s{a}\"=\"%s{escapedValue}\"")
         let parametersArgument =
             match parameters with
             | [] -> ""
-            | parameters -> sprintf "--parameters %s" (parameters |> List.map(fun (a, b) -> $"%s{a}=%s{b}") |> String.concat " ")
```
ninjarobot, line 112: It is always challenging to safely build parameters that will go on a command line. Escaping is tough, especially xplat, and it still becomes an attack surface because these go straight into a command line.

What about generating a JSON parameters file in the temp directory and passing that instead? The JSON serializers already deal with any escaping that way, and the only thing passed on the command line is the path to that temporary file.

ninjarobot (changes requested):
Please clarify the release notes because this is a change to the deployment module, not the deploymentScripts ARM resource.

Also consider getting rid of the parameter generation in favor of generating a JSON parameters file.

src/Farmer/Deploy.fs
amypc: Hi @ninjarobot , sorry for the late response.

We're passing in securestring parameters. I'd prefer not to write these parameter values to a file and risk the file not being deleted properly after the deploy and exposing the secret

amypc:
Hi @ninjarobot , sorry for the late response.

We're passing in securestring parameters. I'd prefer not to write these parameter values to a file and risk the file not being deleted properly after the deploy and exposing the secret

ninjarobot (approved):
LGTM. @isaacabraham do you have any suggestions on communicating about the breakage?

RELEASE_NOTES.md
amypc: Updated, thanks!

amypc:
@ninjarobot @isaacabraham is this PR ok to merge? Would you like me to bump the package to a major version instead to indicate the breaking change or are there other comms to do to indicate the breaking change?

isaacabraham:
Let's do it, it's the right behaviour to have in there.

PR #879, opened 2022-02-15
Added possibility to set vnet rules for PostgreSQL

rq-abrahamsson:
This PR closes -

* Added possibility to set vnet rules for PostgreSQL

Could not figure out how to best test this as I mostly looked at the configuration for adding firewall rules. Can try to add them if this is something you want to merge.

```fsharp
let networkResourceId = $"/subscriptions/%s{subscriptionId}/resourceGroups/%s{resourceGroup}/providers/Microsoft.Network/virtualNetworks/%s{myVnet.ResourceId.Name.Value}/subnets/default"

postgreSQL {
  admin_username "Admin name"
  name "the-name"
  capacity 1<VCores>
  storage_size 1<Gb>
  ...  
  add_vnet_rule "TheNetworkName" networkResourceId
}
```

### DIFF
--- a/src/Farmer/Arm/DBforPostgreSQL.fs
+++ b/src/Farmer/Arm/DBforPostgreSQL.fs
@@ -7,6 +7,7 @@ open Farmer.PostgreSQL
 
 let databases = ResourceType ("Microsoft.DBforPostgreSQL/servers/databases", "2017-12-01")
 let firewallRules = ResourceType ("Microsoft.DBforPostgreSQL/servers/firewallrules",  "2017-12-01")
+let virtualNetworkRules = ResourceType ("Microsoft.DBforPostgreSQL/servers/virtualNetworkRules",  "2017-12-01")
 let servers = ResourceType ("Microsoft.DBforPostgreSQL/servers", "2017-12-01")
 
 [<RequireQualifiedAccess>]
@@ -44,6 +45,18 @@ module Servers =
             member this.JsonModel =
                 {| firewallRules.Create(this.Server/this.Name, this.Location, [ servers.resourceId this.Server ]) with
                     properties = {| startIpAddress = string this.Start; endIpAddress = string this.End; |}
+                |}
+
+    type VnetRule =
+        { Name : ResourceName
+          Server : ResourceName
+          VirtualNetworkSubnetId : string
+          Location : Location }
+        interface IArmResource with
+            member this.ResourceId = virtualNetworkRules.resourceId (this.Server/this.Name)
+            member this.JsonModel =
+                {| virtualNetworkRules.Create(this.Server/this.Name, this.Location, [ servers.resourceId this.Server ]) with
+                    properties = {| virtualNetworkSubnetId = this.VirtualNetworkSubnetId |}
                 |}
 
 type Server =
--- a/src/Farmer/Arm/DBforPostgreSQL.fs
+++ b/src/Farmer/Arm/DBforPostgreSQL.fs
@@ -50,13 +50,13 @@ module Servers =
     type VnetRule =
         { Name : ResourceName
           Server : ResourceName
-          VirtualNetworkSubnetId : string
+          VirtualNetworkSubnetId : ResourceId
           Location : Location }
         interface IArmResource with
             member this.ResourceId = virtualNetworkRules.resourceId (this.Server/this.Name)
             member this.JsonModel =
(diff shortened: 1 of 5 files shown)


ninjarobot (changes requested):
This looks great overall, the only thing to add would be a unit test so that in case there are changes, we can be sure they won't break the builder. There are some [similar tests for firewall rules](https://github.com/CompositionalIT/farmer/blob/master/src/Tests/PostgreSQL.fs#L134).

src/Farmer/Builders/Builders.PostgreSQL.fs
```diff
@@ -28,6 +28,7 @@ type PostgreSQLConfig =
       Tier : Sku
       Databases : PostgreSQLDbConfig list
       FirewallRules : {| Name : ResourceName; Start : IPAddress; End : IPAddress |} list
+      VirtualNetworkRules : {| Name : ResourceName; VirtualNetworkSubnetId : string |} list
```
ninjarobot, line 31: Is the `VirtualNetworkSubnetId` just the name of the subnet or is it expected to be the full ARM resource Id of a subnet?

rq-abrahamsson: It is the full ARM resource Id looking like this: `$"/subscriptions/%s{subscriptionId}/resourceGroups/%s{resourceGroup}/providers/Microsoft.Network/virtualNetworks/%s{myVnet.ResourceId.Name.Value}/subnets/default"`.

rq-abrahamsson:
Will try to add tests for this!

isaacabraham:
Consider adding an "add_vnet_rules" as well that takes in a list. This helps for programmatic scenarios.

rq-abrahamsson:
> Consider adding an "add_vnet_rules" as well that takes in a list. This helps for programmatic scenarios.

Ah, yes! It was added, can be found here: https://github.com/CompositionalIT/farmer/blob/352bde1d274d7f1a83cb31e5317475bd568ce2f3/src/Farmer/Builders/Builders.PostgreSQL.fs#L315
I should probably have added that too in the minimal example configuration, missed that part.

src/Farmer/Builders/Builders.PostgreSQL.fs
ninjarobot: It should be a `ResourceId` rather than just a string so nobody can just pass in a subnet name and then get a deployment error.

src/Farmer/Arm/DBforPostgreSQL.fs
```diff
@@ -46,6 +47,18 @@ module Servers =
                     properties = {| startIpAddress = string this.Start; endIpAddress = string this.End; |}
                 |}
 
+    type VnetRule =
+        { Name : ResourceName
+          Server : ResourceName
+          VirtualNetworkSubnetId : string
```
ninjarobot, line 53: Please make this a `ResourceId` so that it can't be populated with an arbitrary string.

ninjarobot (changes requested):
I only have a minor change requested to use a `ResourceId` instead of a `string` to ensure an ARM ID is used for the subnet. The ARM network resource provider doesn't validate for cases like that very well so errors will show up during the deployment if this isn't set correctly. It will be a breaking change to constrain this type later, so best to do it this way form the start.

rq-abrahamsson:
> I only have a minor change requested to use a `ResourceId` instead of a `string` to ensure an ARM ID is used for the subnet. The ARM network resource provider doesn't validate for cases like that very well so errors will show up during the deployment if this isn't set correctly. It will be a breaking change to constrain this type later, so best to do it this way form the start.

Yes, I completely agree. Just forgot to answer and have not had time to fix it yet, will hopefully have time to fix it sometime this week. 👍

rq-abrahamsson:
> I only have a minor change requested to use a `ResourceId` instead of a `string` to ensure an ARM ID is used for the subnet. The ARM network resource provider doesn't validate for cases like that very well so errors will show up during the deployment if this isn't set correctly. It will be a breaking change to constrain this type later, so best to do it this way form the start.

Now I have added a commit that should address this and use `ResourceId` instead of `string`. It was a bit trickier to get right than I thought it would be as I could not find any examples usages and for KeyVault it seems to be a string. Is this a new pattern used for this?

rq-abrahamsson:
@ninjarobot What is the status of this? Is it supposed to be merged anytime soon? Or is there still something that needs to be fixed?

ninjarobot (commented):
Thanks for adding type safety on the subnet ID.

PR #883, opened 2022-03-04
Allow adding an IP restriction with CIDR as a string to web apps and functions

theprash:
#835 added support for adding IP restrictions to web apps and functions. If adding an IP address directly as a string, CIDR notation was not allowed.

This PR allows using CIDR notation, while keeping the prefix as 32 if no CIDR is specified. This is a non-breaking change. 

---

```fsharp
let myWebApp = webApp {
    name "mysuperwebapp"
    add_allowed_ip_restriction "allow everything" "0.0.0.0/0"
    add_denied_ip_restriction "deny" "1.2.3.4/31"
}
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,11 +1,14 @@
 Release Notes
 =============
 
+## 1.6.30
+* WebApps/Functions: Allow adding IP restriction string with CIDR
+
 ## 1.6.29
 * CLI: include `--only-show-error` option when executing Azure CLI commands.
 
 ## 1.6.28
-* ServicePlan/WebApp: Support for enabling ZoneDedundant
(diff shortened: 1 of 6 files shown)


theprash:
@isaacabraham @ninjarobot That's all done now. I don't think docs changes are needed but let me know if you think otherwise.

isaacabraham:
@theprash if there are new keywords then they need to be added to the docs. I know that's a bit crap but the lack of easy discoverability of CE keywords is such that we've had to do it.

theprash:
@isaacabraham I didn't add any new keywords, just updated the behaviour of existing overloads in a non-breaking way. So the keywords are already in the docs but not with their parameters and overloads documented, which seems to be the case for almost all other keywords.

ninjarobot (approved):

isaacabraham:
Great thanks :-)

PR #887, opened 2022-03-07
Hint resource provider

tforkmann:
Had to learn it the hard way that you need to enable the resource providers first in your Azure Subscriptions before you can deploy a container app.


isaacabraham:
Thanks :-)

PR #892, opened 2022-03-11
Add --overwrite true to storage blob upload command

### LINKED ISSUE #890: Function App zip deploy fails with ErrorCode:BlobAlreadyExists on Azure CLI v2.34.1
```
Running ZIP deploy to WebApp for publish/app
Unhandled exception. System.Exception: ERROR: The specified blob already exists.
RequestId:b01219f5-d01e-004c-2fd0-34212a000000
Time:2022-03-10T22:47:02.3618883Z
ErrorCode:BlobAlreadyExists
If you want to overwrite the existing one, please add --overwrite in your command.
```

Looks to be related to https://github.com/Azure/azure-cli/issues/21477

The failing task in my case appears to be the zip deploy executed by the following:
```F#
    let functionApp = functions {
        name functionAppName
        add_tag "Environment" environmentTag
        service_plan_name $"{appName}-plan"
        link_to_storage_account functionAppStorage.Name.ResourceName
        link_to_app_insights insights
        use_runtime FunctionsRuntime.DotNet60Isolated
        use_extension_version V4
        worker_process Bits64
        system_identity
        https_only
        zip_deploy "publish/app"
    }
```

duncanworthy:
This PR closes #890

* Adds the parameter `--overwrite true` to the `az storage blob upload-batch` command used for WebApp zip deployment, which is required due to breaking changes introduced in [Azure CLI v2.34.0](https://github.com/MicrosoftDocs/azure-docs-cli/blob/main/docs-ref-conceptual/release-notes-azure-cli.md#march-03-2022)

* No feature changes - only a small change to something not covered by existing tests in order to maintain compatibility with the latest Azure CLI version

### DIFF
--- a/src/Farmer/Deploy.fs
+++ b/src/Farmer/Deploy.fs
@@ -79,7 +79,30 @@ module Az =
     let login() = az "login" |> Result.ignore
     /// Logs you into the Az CLI using the supplied service principal credentials.
     let loginWithCredentials appId secret tenantId = az $"login --service-principal --username %s{appId} --password %s{secret} --tenant %s{tenantId}"
+    /// Gets the version of Az CLI
     let version() = az "--version"
+    /// Checks that the version of the Azure CLI meets minimum version.
+    let checkVersion minimum = result {
+        let! versionOutput = version()
+        let! version =
+            versionOutput.Replace("\r\n","\n").Replace("\r","\n").Split([| "\n" |], StringSplitOptions.RemoveEmptyEntries)
+            |> Array.tryHead
+            |> Option.bind(fun text ->
+                match text.Split([| ' ' |], StringSplitOptions.RemoveEmptyEntries) with
+                | [| _; version |]
+                | [| _; version; _ |] ->
+                    Some version
+                | _ ->
+                    None)
+            |> Option.bind(fun versionText ->
+                try Some(Version versionText)
+                with _ -> None)
+            |> Result.ofOption $"Unable to determine Azure CLI version. You need to have at least {minimum} installed. Output was: %s{versionOutput}"
+        return!
+            if version < minimum then Error $"You have {version} of the Azure CLI installed, but the minimum version is {minimum}. Please upgrade."
+            else Ok version
+    }
+    
     /// Lists all subscriptions
     let listSubscriptions() = az "account list --all"
     let setSubscription subscriptionId = az $"account set --subscription %s{subscriptionId}"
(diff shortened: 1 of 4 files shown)


isaacabraham:
@duncanworthy Thank you for this! Just to double check, the extra argument was always there, but it just wasn't on by default, right - so older versions of the CLI will still continue to work?

duncanworthy:
@isaacabraham - good point! I need to do a bit more testing and code changes around that.
I made an incorrect assumption from the conversation in the az cli issue.
Having looked through the az cli repo history, it seems they may have added `--overwrite` in v2.34.0, and then made the breaking change to flip the default behaviour in v2.34.1.
We'll need to include or not that parameter based on azure CLI version.

duncanworthy:
@isaacabraham I've pushed a new commit to include the CLI version check and only add the `--overwrite` parameter on 2.34 and above.
I've tested locally to a live azure subscription with CLI versions:
- 2.33.1 doesn't accept the parameter and works fine overwriting files without it
- 2.34.0 and 2.34.1 accept the parameter and only overwrite if the parameter is specified

In order to use your existing checkVersion function I had to move it up into the Farmer.Deploy.Az module. Let me know if that's ok or if you want it done differently.

src/Farmer/Deploy.fs
```diff
@@ -133,7 +156,8 @@ module Az =
         |> String.concat " "
         |> az
     let batchUploadStaticWebsite name path =
-        az $"storage blob upload-batch --account-name %s{name} --destination $web --source %s{path}"
+        let overwriteParameter = checkVersion (Version "2.34.0") |> function | Ok _ -> "--overwrite true" | Error _ -> ""
```
isaacabraham, line 159: Perhaps bind `Version "2.34.0"` into a (private?) symbol in the module? Nothing wrong with leaving it here, but it might be good to put it somewhere with a specific comment about why 2.34.0 is "special".

isaacabraham: Also consider replacing the use of `function` with a (slightly more verbose but more common) pattern match.

isaacabraham (commented):
LGTM aside from a couple of aesthetic changes. Thank you!

src/Farmer/Deploy.fs
duncanworthy: Something like this?
```fsharp
    /// The overwrite parameter was introduced in Azure CLI v2.34.0 with a breaking change to the default behaviour
    let private cliVersionWithOverwriteParameter = Version "2.34.0"
    let batchUploadStaticWebsite name path =
        let additionalParameters =
            match checkVersion cliVersionWithOverwriteParameter with
            | Ok _ -> "--overwrite true"
            | _ -> ""
        az $"storage blob upload-batch --account-name %s{name} --destination $web --source %s{path} {additionalParameters}"
```

isaacabraham: Yep exactly

duncanworthy: changes pushed

PR #894, opened 2022-03-12
Vm spot instance

bigjonroberts:

* Adds Priority and Spot Instance settings for VMs

```fsharp
#r "nuget:Farmer"

open Farmer
open Farmer.Builders
open Farmer.Vm

let myVm = vm {
    name "isaacsVM"
    username "isaac"
    spot_instance Deallocate
    vm_size Standard_A2
    operating_system WindowsServer_2012Datacenter
    os_disk 128 StandardSSD_LRS
    add_ssd_disk 128
    add_slow_disk 512
    diagnostics_support
    system_identity
}

let deployment = arm {
    location Location.NorthEurope
    add_resource myVm
}

deployment
|> Deploy.execute "my-resource-group-name" Deploy.NoParameters
```

### DIFF
--- a/src/Farmer/Arm/Compute.fs
+++ b/src/Farmer/Arm/Compute.fs
@@ -7,7 +7,7 @@ open Farmer.Vm
 open System
 open System.Text
 
-let virtualMachines = ResourceType ("Microsoft.Compute/virtualMachines", "2018-10-01")
+let virtualMachines = ResourceType ("Microsoft.Compute/virtualMachines", "2019-03-01")
 let extensions = ResourceType ("Microsoft.Compute/virtualMachines/extensions", "2019-12-01")
 
 type CustomScriptExtension =
@@ -70,6 +70,7 @@ type VirtualMachine =
       Location : Location
       StorageAccount : ResourceName option
       Size : VMSize
+      Priority : Priority
       Credentials : {| Username : string; Password : SecureParameter |}
       CustomData : string option
       DisablePasswordAuthentication: bool option
@@ -93,12 +94,9 @@ type VirtualMachine =
                 networkInterfaces.resourceId this.NetworkInterfaceName
                 yield! this.StorageAccount |> Option.mapList storageAccounts.resourceId
             ]
-            {| virtualMachines.Create(this.Name, this.Location, dependsOn, this.Tags) with
-                identity =
-                    if this.Identity = ManagedIdentity.Empty then Unchecked.defaultof<_>
-                    else this.Identity.ToArmJson
-                properties =
-                 {| hardwareProfile = {| vmSize = this.Size.ArmValue |}
+            let properties =
+                 {| priority = this.Priority.ArmValue
+                    hardwareProfile = {| vmSize = this.Size.ArmValue |}
                     osProfile =
                      {| computerName = this.Name.Value
                         adminUsername = this.Credentials.Username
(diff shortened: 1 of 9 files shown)


bigjonroberts:
@isaacabraham @ninjarobot 
Do you have any suggestions for automated tests for this change?  Or does what I have here now look sufficient?

src/Farmer/Arm/Compute.fs
```diff
@@ -93,12 +94,9 @@ type VirtualMachine =
                 networkInterfaces.resourceId this.NetworkInterfaceName
                 yield! this.StorageAccount |> Option.mapList storageAccounts.resourceId
             ]
-            {| virtualMachines.Create(this.Name, this.Location, dependsOn, this.Tags) with
-                identity =
-                    if this.Identity = ManagedIdentity.Empty then Unchecked.defaultof<_>
-                    else this.Identity.ToArmJson
-                properties =
-                 {| hardwareProfile = {| vmSize = this.Size.ArmValue |}
+            let properties =
```
bigjonroberts, line 97: Implemented this way because ARM chokes on any priority setting than 'Spot1 if `evictionPolicy` or `billingProfile` are included.

src/Farmer/Arm/Compute.fs
```diff
@@ -7,7 +7,7 @@ open Farmer.Vm
 open System
 open System.Text
 
-let virtualMachines = ResourceType ("Microsoft.Compute/virtualMachines", "2018-10-01")
+let virtualMachines = ResourceType ("Microsoft.Compute/virtualMachines", "2019-03-01")
```
bigjonroberts, line 10: Version 2019-03-01 is the oldest version to support the spot instance properties.

src/Farmer/Builders/Builders.Vm.fs
```diff
@@ -258,6 +262,14 @@ type VirtualMachineBuilder() =
     /// Adds a data disk to the VM with a specific size and type.
     [<CustomOperation "add_disk">]
     member _.AddDisk(state:VmConfig, size, diskType) = { state with DataDisks = { Size = size; DiskType = diskType } :: state.DataDisks }
+    /// Sets priority of VMm. Overrides spot_instance.
+    [<CustomOperation "priority">]
+    member _.Priority(state:VmConfig, priority) = { state with Priority = priority }
+    /// Makes VM a spot instance. Overrides priority.
+    [<CustomOperation "spot_instance">]
+    member _.Spot(state:VmConfig, (evictionPolicy, maxPrice)) : VmConfig = { state with Priority = Spot (evictionPolicy, maxPrice) }
```
ninjarobot, line 270: If a user sets priority and then sets spot_instance, their priority will be set to Spot, which is fine. What about the other way around if they set spot_instance and then set priority? Is there some validation needed to ensure they can't end up with the billing profile set while using a priority other than Spot?

src/Farmer/Common.fs
```diff
@@ -309,6 +309,20 @@ module Vm =
 
     /// Represents a disk in a VM.
     type DiskInfo = { Size : int; DiskType : DiskType }
+    type EvictionPolicy =
+        | Deallocate
+        | Delete
+        member this.ArmValue = match this with x -> x.ToString()
```
ninjarobot, line 315: Prefer to use strings here just in case one of the DU cases is changed to where it holds some data since that would cause ToString to be more than just the name.

ninjarobot (changes requested):
Thanks for contributing this Jon! I have a couple small suggestions where I commented in the code. As for unit testing, it would be helpful to add a test that the logic around the priority and billing profile holds up even if both priority and spot_instance are set, regardless of order.

src/Farmer/Common.fs
bigjonroberts: Changed on this and the other new DU that was introduced.

bigjonroberts:
@ninjarobot I think my latest design change addresses this concern in the correct way.  It's covered with tests now, too.  Take a look at the last commit and see if you agree.

ninjarobot (approved):
Thanks again for the contribution!

PR #896, opened 2022-03-14
Fix multiple custom domains

JustinWilkinson:
This PR fixes conflicts seen when attempting to deploy multiple custom domains due to a race in the deploy. 

* Deployments of custom domains are forced to run sequentially using a dependency on the previous deployment.
* Fixes issue seen when assigning a dependency to a nested deployment which contains a template string in its name.

There aren't any changes to the API, the behaviour underneath is all that has changed.

```fsharp
 webApp {
	name webappName
	custom_domains ["secure1.io" ; "secure2.io" ; "secure3.io"]
}
```

### DIFF
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -514,26 +514,34 @@ type WebAppConfig =
                 { site with AppSettings = None; ConnectionStrings = None } // Don't deploy production slot settings as they could cause an app restart
                 for (_,slot) in this.CommonWebConfig.Slots |> Map.toSeq do
                     slot.ToSite site
-                
-            // Host Name Bindings must be deployed sequentially to avoid an error, as the site cannot be modified concurrently.
-            // To do so we add a dependency to the previous binding.
-            let mutable previousHostNameBinding = None
-            for customDomain in this.CustomDomains |> Map.toSeq |> Seq.map snd do
-                let dependsOn = 
-                    match previousHostNameBinding with 
-                    | Some previous -> Set.singleton previous
-                    | None -> Set.empty
+             
+            // Need to rename `location` binding to prevent conflict with `location` operator in resource group
+            let resourceLocation = location
 
+            // Host Name Bindings must be deployed sequentially to avoid an error, as the site cannot be modified concurrently.
+            // To do so we add a dependency to the previous binding deployment.
+            let mutable previousHostNameCertificateLinkingDeployment = None
+            for customDomain in this.CustomDomains |> Map.toSeq |> Seq.map snd do
                 let hostNameBinding =
                     { Location = location
                       SiteId =  Managed (Arm.Web.sites.resourceId this.Name.ResourceName)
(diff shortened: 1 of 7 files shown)


src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -514,74 +514,84 @@ type WebAppConfig =
...
-
                 let hostNameBinding =
                     { Location = location
                       SiteId =  Managed (Arm.Web.sites.resourceId this.Name.ResourceName)
                       DomainName = customDomain.DomainName
-                      SslState = SslDisabled // Initially create non-secure host name binding, we link the certificate in a nested deployment below if this is a secure domain.
-                      DependsOn = dependsOn }
+                      SslState = SslDisabled } // Initially create non-secure host name binding, we link the certificate in a nested deployment below
+
+                let dependsOn : ResourceId list =
```
ninjarobot, line 531: This should be the previous host name binding deployment(s), right? I realize it was named `dependsOn` before, but it would help readability to name it for what it is.

Also, this should be a `set` rather than a `list`. Behavior is similar, but it will ensure no duplicates. It will ultimately be converted to a set when building the ARM template resources, but if it can be a set right away, anything that processes it in the interim will also not have to worry with duplicates.

src/Farmer/Types.fs
```diff
@@ -138,7 +138,7 @@ type ResourceId with
               this.Type.Type
               this.Name.Value
               for segment in this.Segments do segment.Value ]
-            |> List.map (sprintf "'%s'")
+            |> List.map (fun x -> if x.StartsWith("[") then x.Trim('[' ,']') else sprintf $"'%s{x}'") // Fixes case where a template e.g. [resourceGroup().name] is used.
```
ninjarobot, line 141: An interpolated string doesn't also need `sprintf`.

ninjarobot (changes requested):
Thanks for the contribution to fix this race condition. I recommended a few minor adjustments, please, and then it should be good to merge.

src/Farmer/Builders/Builders.WebApp.fs
JustinWilkinson: The `depends_on` operator takes a `ResourceId list` as a parameter, so this has to be a `list` at that point. I can make it a `Set` but then convert it back if you'd like.

As for the naming, I was playing it safe and have two dependencies, one to the WebApp itself, and another to the previous host name binding deployment. Whilst the former is probably not strictly necessary leaving it in the generated template might make things more explicit?

src/Farmer/Types.fs
JustinWilkinson: Oh bother - merged the wrong master into this branch 😞

JustinWilkinson: Ah bother - seem to have merged the wrong master 😞 - will need to sort that...

JustinWilkinson: Right, sorry about that, have merged the correct `master` this time - doh.

PR #898, opened 2022-03-14
Support Virtual Applications

### LINKED ISSUE #891: Support Virtual Applications for Web Apps
Currently, Farmer does not support any method to add "Virtual Applications" to Azure Web Apps. These are useful for allowing you to easily host front ends and back ends together under one web app (e.g. in case you do not need separate APIs for a "simple" application).

[Virtual Application arm template reference](https://docs.microsoft.com/en-us/azure/templates/microsoft.web/sites?tabs=json#virtualapplication)

```json
{
  "type": "Microsoft.Web/sites",
  "apiVersion": "2021-03-01",
  "name": "string",
  "location": "string",
  //...
  "properties": {
    //...
    "siteConfig": {
      //...
      "virtualApplications": [
        {
          "physicalPath": "string",
          "preloadEnabled": "bool",
          "virtualDirectories": [
            {
              "physicalPath": "string",
              "virtualPath": "string"
            }
          ],
          "virtualPath": "string"
        }
      ],
      //...
    },
  }
}
```

Example of usage of proposed feature:
```fsharp
let wa = webApp {
    name "my-two-in-one-app"
    add_virtual_application "/" "site\\wwwroot" // Default virtual app definition - Required if any virtual application specified
    add_virtual_application_preloaded "/api" "site\\wwwroot\\api" // Checks "preloadEnabled" boolean
}
```

Additional consideration:
Azure _requires_ that the physical path start with `site`. Would it make sense to just omit that in Farmer? e.g. `add_virtual_application "/" "wwwroot"` instead. This would make the ARM template and Farmer not exactly 1-to-1, but you aren't allowed to specify anything outside of site anyway.
image
Otherwise, should the path be validated to prevent the erroneous ARM template from being generated?

Zshazz:
This PR closes #891

* Adds support for Virtual Applications nested under a Web App

```fsharp
let wa = webApp {
    name "my-two-in-one-app"

    // Adds a virtual application under the "api" sub-folder of the default app, checking the Preload option
    add_virtual_application_preloaded "/api" "wwwroot\\api"
}
```

### DIFF
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -187,6 +187,31 @@ type SlotBuilder() =
 
 let appSlot = SlotBuilder()
 
+type VirtualApplicationConfig =
+    { VirtualPath: string
+      PhysicalPath: string
+      PreloadEnabled: bool option }
+
+type VirtualApplicationBuilder() =
+    member this.Yield _ =
+        { VirtualPath = ""
+          PhysicalPath = ""
+          PreloadEnabled = None }
+    member _.Run (config: VirtualApplicationConfig) =
+        if String.IsNullOrWhiteSpace config.VirtualPath then
+            raiseFarmer "Missing Virtual Path on Virtual Application - specify 'virtual_path' on all virtual applications"
+        if String.IsNullOrWhiteSpace config.PhysicalPath then
+            raiseFarmer "Missing Physical Path on Virtual Application - specify 'physical_path' on all virtual applications"
+        config
+    [<CustomOperation "virtual_path">]
+    member _.VirtualPath (state, virtualPath) : VirtualApplicationConfig = { state with VirtualPath = virtualPath }
+    [<CustomOperation "physical_path">]
+    member _.PhysicalPath (state, physicalPath) : VirtualApplicationConfig = { state with PhysicalPath = physicalPath }
+    [<CustomOperation "preloaded">]
+    member _.Preloaded state : VirtualApplicationConfig = { state with PreloadEnabled = Some true }
+
+let virtualApplication = VirtualApplicationBuilder()
+
 /// Common fields between WebApp and Functions
 type CommonWebConfig =
     { Name : WebAppName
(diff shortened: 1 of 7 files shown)


src/Farmer/Builders/Builders.WebApp.fs
```diff
@@ -738,6 +740,23 @@ type WebAppBuilder() =
...
+    member _.AddVirtualApplication(state:WebAppConfig, physicalPath, virtualPath, preloadEnabled) = 
+        { state with VirtualApplications =
+                        if state.VirtualApplications.IsEmpty
+                            then Map [ ("/", VirtualApplication.Create "site\\wwwroot" None ) ]
+                            else state.VirtualApplications
+                        |> Map.add physicalPath (VirtualApplication.Create ("site\\" + virtualPath) preloadEnabled) }
+
+    /// Adds Virtual Application definition
+    [<CustomOperation "add_virtual_application">] 
+    member this.AddVirtualApplication(state:WebAppConfig, physicalPath, virtualPath) =
```
ninjarobot, line 752: Would a builder for the virtualApplication be useful here? Otherwise it's easy to mix up the physical and virtual path. Also, adding virtual directories will need another overload here whereas a builder won't. Do you think that will simplify it for end users?

ninjarobot (commented):
Thanks for contributing this! Please take a look at my comment about adding a builder instead of overloads.

src/Farmer/Builders/Builders.WebApp.fs
Zshazz: I agree. We're thinking something like this?

```
let wa = webApp {
    name "my-two-in-one-app"

    // Adds a virtual application under the "api" sub-folder of the default app, checking the Preload option
    add_virtual_application {
        physical_path "/api" // Optional: defaults to root "/"
        virtual_path "wwwroot\\api" // Optional? Could default to "wwwroot"
        preloaded
    }
}
```

I can get to work on that in a few days and make the changes in our deployment to test it out.

mattgallagher92:
@Zshazz I've merged from master and made a few tweaks

mattgallagher92:
I like the defaults you've chosen - adding "site\\" and, if not specified, the virtual app at "/".

src/Farmer/Builders/Builders.WebApp.fs
mattgallagher92: I support the idea of a builder 👍 

Would the defaults for `physical_path` and `virtual_path` be of very much use? They're easy to add manually and "/" -> "sites\\wwwroot" is already added if omitted anyway.

mattgallagher92:
Tested, confirm this works 👍 Thanks @Zshazz !

mattgallagher92 (commented):
Tested, confirm this works 👍 Thanks @Zshazz ! By the way, I merged from master and made a few tweaks.

I like the defaults you've chosen - adding "site\\" and, if not specified, the virtual app at "/".

Will hold off on approval until the builder thread is resolved.

src/Farmer/Builders/Builders.WebApp.fs
Zshazz: I should have an update to this PR with the builder pattern implemented shortly, but with blank "defaults" for those and using a `Run` method for validation (similar to usages elsewhere) to ensure they are set. I've also taken a little bit more time updating the documentation with an example.

```fsharp
let wa = webApp {
    name "my-site-with-api"
    always_on
    add_virtual_applications [
        virtualApplication {
            virtual_path "/"
            physical_path "frontend" 
        }
        virtualApplication {
            virtual_path "/api"
            physical_path "backend"
            preloaded
        }
    ]
}
```

I've modeled the approach based on existing similar builders (e.g. [vnet/subnets](https://compositionalit.github.io/farmer/api-overview/resources/vnet/#example---manual-subnets)).

ninjarobot (approved):
I like the builder - that will be better for ongoing compatibility, thanks for adding that @Zshazz.

PR #910, opened 2022-04-08
Implement changes in Container apps namespaces

forki:
See https://github.com/microsoft/azure-container-apps/issues/109


forki:
@ninjarobot @isaacabraham  I forgot to write: this is rather urgent since the old namespace doesn't work anymore

isaacabraham:
Thanks. I'll merge in now.

I think it should be moved to a separate file though to conform to the pattern we use in Farmer (i.e. Arm resources for Microsoft.App go into App.fs etc.).

PR #920, opened 2022-04-29
Operations management support

### LINKED ISSUE #918: Support for Operations Management
[Operations Management](https://docs.microsoft.com/en-us/azure/templates/microsoft.operationsmanagement/solutions?tabs=bicep) allows for creation solutions to use with a Log Analytics Workspace.

I specifically need this for Sentinel but it looks like it should be possible to make this generic enough to work for a variety of cases.

[Here](https://github.com/Azure/azure-quickstart-templates/blob/master/quickstarts/microsoft.operationsmanagement/azure-sentinel/main.bicep) is an example in Bicep.

PaigeM89:
I was able to get this to do what I need it to do on Azure, but I simply do not know enough about Azure in general to know how to test it further or how to make sure it's usable for a variety of use cases outside my specific one. I'm not even 100% sure I'm referencing the correct things in Azure when I say "Operations Management" or "Solutions".

This PR closes #918

* Add support for Operations Management, which allows the deployment of configurable Solutions.

* I have no idea how to meaningfully unit test this, as there's no way I can find to build the samples into an azure resource to verify against.

```fsharp
open Farmer
open Farmer.Builders

let sentinelWorkspace = logAnalytics {
    name "my-sentinel-workspace"
    retention_period 30<Days>
    enable_query
    daily_cap 5<Gb>
}

// I have no idea of the scope of possible solution names, so I left this as a user-configured string.
// If possible, it makes sense to make this more type-safe.
let solutionName = $"SecurityInsights({sentinelWorkspace.Name.Value})"

let sentinelSolution = solution {
    name solutionName
    plan (solutionPlan {
        name solutionName
        publisher "Microsoft"
        product "OMSGallery/SecurityInsights"
    })
    properties(solutionProperties {
        workspace sentinelWorkspace
    })
}

let deployment = arm {
  location Location.NorthCentralUS
  add_resource sentinelWorkspace
  add_resource sentinelSolution
}
```


PaigeM89:
I just realized this is also covered in [this](https://github.com/CompositionalIT/farmer/pull/632) PR. I can edit that one and delete this one if you want?

src/Farmer/Builders/Builders.OperationsManagement.fs
```diff
@@ -0,0 +1,120 @@
+[<AutoOpen>]
+module Farmer.Builders.OperationsManagement
+
+open Farmer
+open Farmer.Arm.OperationsManagement
+
+type SolutionProperties =
+    { Workspace: IBuilder option
+      // todo - should these be IBuilders to access ResourceIds?
+      ContainedResources: string list
```
PaigeM89, line 10: I can't quite tell from the docs if these need to be Resource Ids, nor do I feel like I understand enough about this feature to even test it fully.

src/Farmer/Arm/OperationsManagement.fs
```diff
@@ -0,0 +1,38 @@
+[<AutoOpen>]
+module Farmer.Arm.OperationsManagement
+
+open Farmer
+
+let solutions = ResourceType("Microsoft.OperationsManagement/solutions", "2015-11-01-preview")
+
+type Solution =
```
ninjarobot, line 8: Because `Solution` is such a well-known component in .NET and MSBuild, I suggest naming these `OperationsManagementSolution` or `OMS` (as people often refer to this) just to avoid conflating with the .NET Solution file.

src/Farmer/Builders/Builders.OperationsManagement.fs
ninjarobot: Yes, they are supposed to be resource Ids. For example, there may be views within the solution that contain various dashboards, so they'll need to be referenced here.

src/Farmer/Builders/Builders.OperationsManagement.fs
```diff
@@ -0,0 +1,120 @@
...
+        { Workspace = None
+          ContainedResources = []
+          ReferencedResources = [] }
+
+    /// Sets the workspace resource id of the Solution Properties
+    [<CustomOperation "workspace">]
+    member _.WorkspaceResourceId(state: SolutionProperties, workspace) = { state with Workspace = Some workspace }
+
+    /// Sets the contained resources of the Solution Properties
+    [<CustomOperation "contains">]
```
ninjarobot, line 69: Most of the builders here that have collections use `add_{thing}` to let you call the operation multiple times to add more items to the collection. [Here is an example](https://github.com/CompositionalIT/farmer/blob/master/src/Farmer/Builders/Builders.Alert.fs#L62-L65) just from another random resource. It might be good to cal this `add_contained_resources` and enable it to add more resources if called multiple times.

src/Farmer/Builders/Builders.OperationsManagement.fs
```diff
@@ -0,0 +1,120 @@
...
+    member _.ContainedResources(state: SolutionProperties, contained) = { state with ContainedResources = contained }
+
+    /// Sets the referenced resources of the Solution Properties
+    [<CustomOperation "references">]
+    member _.ReferencedResources(state: SolutionProperties, referenced) = { state with ReferencedResources = referenced }
```
ninjarobot, line 74: Same as for contained resources, this could be `add_referenced_resources` to follow the convention and allow more resources to be added by calling the operation multiple times.

ninjarobot (dismissed):
Thanks for the contribution, this looks very good overall. I have a few suggestions for the builder operation semantics to align with other resources. Also, the name "solution" often refers to the .NET Solution file, so to avoid confusing these two, I suggest a new name, either the wordy "OperationsManagementSolution" or how I often hear people refer to it, "oms".

As for unit tests, could you please add at least one that shows the builder usage and checks that it builds the expected resource?

Issue #921, opened 2022-04-29
Can't supply container registry credentials to Container App.


isaacabraham:
Something is up with the Container App / Container Registry integration. I could swear that this worked a month or so ago, but I tried creating a new container app service today. This is the offending line:

```fsharp
containerApp {
    reference_registry_credentials [ Arm.ContainerRegistry.registries.resourceId mySuperRegistry.Name ]
}
```

where `registry` is an Azure container registry that already exists.

When I try an deploy the container app, I get the following validation error:

```
{'code': 'ContainerAppInvalidPropertyValue', 'message': "Property 'secrets.name' has an invalid value '[listCredentials(resourceId('Microsoft.ContainerRegistry/registries', 'mySuperRegistry'), '2019-05-01').username]'. A value must consist of lower case alphanumeric characters, '-'
 or '.', and must start and end with an alphanumeric character."}
)
```

Here's an extract of the raw ARM template:

```json
          "registries": [
            {
              "passwordSecretRef": "[listCredentials(resourceId('Microsoft.ContainerRegistry/registries', 'mySuperRegistry'), '2019-05-01').username]",
              "server": "[reference(resourceId('Microsoft.ContainerRegistry/registries', 'mySuperRegistry'), '2019-05-01').loginServer]",
              "username": "[listCredentials(resourceId('Microsoft.ContainerRegistry/registries', 'mySuperRegistry'), '2019-05-01').username]"
            }
          ],
          "secrets": [
            {
              "name": "[listCredentials(resourceId('Microsoft.ContainerRegistry/registries', 'mySuperRegistry'), '2019-05-01').username]",
              "value": "[listCredentials(resourceId('Microsoft.ContainerRegistry/registries', 'mySuperRegistry'), '2019-05-01').passwords[0].value]"
            }
          ]
```

It's almost as though ARM hasn't seen that secrets.name is an expression, and has instead tried to parse the expression as a raw string.

@ninjarobot Have you any idea about this - or if this ever used to work?
@anthonychu has anything changed in the way that secret names are handled in terms of ARM template support - maybe something has changed since the namespace move?

anthonychu:
Sounds like it’s the same as this https://github.com/microsoft/azure-container-apps/issues/153

isaacabraham:
Ok, at least I'm not imagining things! Thanks

et1975:
[...]

PR #922, opened 2022-05-02
containerApp evn var collections

et1975:

* Adds support for passing a collection of env variables of a given type to `containerApp` 
Why is this useful? In the absence of broad support for `_.For` in this library this makes iterators possible and brings it to parity with some other CEs.

```fsharp
                containerApp {
                    name "multienv"
                    add_simple_container "mcr.microsoft.com/dotnet/samples" "aspnetapp"
                    ingress_target_port 80us
                    ingress_transport Auto
                    add_http_scale_rule "http-scaler" { ConcurrentRequests = 10 }
                    add_cpu_scale_rule "cpu-scaler" { Utilisation = 50 }
                    add_secret_parameters ["servicebusconnectionkey"]
                    add_env_variables ["ServiceBusQueueName","wishrequests"]
                    add_secret_expressions ["containerlogs", containerLogs.PrimarySharedKey]
                }
```

### DIFF
--- a/src/Farmer/Arm/App.fs
+++ b/src/Farmer/Arm/App.fs
@@ -55,6 +55,7 @@ type ContainerApp =
     interface IArmResource with
         member this.ResourceId = containerApps.resourceId this.Name
         member this.JsonModel =
+            let usernameSecretName (resourceId:ResourceId) = $"{resourceId.Name.Value}-username"
             {| containerApps.Create(this.Name, this.Location, this.dependencies) with
                    kind = "containerapp"
                    identity =
@@ -72,7 +73,7 @@ type ContainerApp =
                                                {| name = cred.Username
                                                   value = cred.Password.ArmExpression.Eval() |}
                                            | ImageRegistryAuthentication.ListCredentials resourceId ->
-                                               {| name = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').username").Eval()
+                                               {| name = usernameSecretName resourceId
                                                   value = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').passwords[0].value").Eval() |}
                                        for setting in this.Secrets do
                                            {| name = setting.Key.Value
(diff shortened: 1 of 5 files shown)


et1975:
Added a fix for #921

src/Farmer/Builders/Builders.ContainerApps.fs
```diff
@@ -34,7 +34,12 @@ type ContainerAppConfig =
       ImageRegistryCredentials : ImageRegistryAuthentication list
       Containers : ContainerConfig list
       Dependencies : Set<ResourceId> }
-
+    member this.ResourceId = containerApps.resourceId this.Name
```
et1975, line 37: I don't think [line 51](https://github.com/CompositionalIT/farmer/pull/922/files#diff-b82693f74fcd946fa59b9816bacfd8d576ea0920bbd9a9da3bb293d29044cac3R51) is correct actually, wondering if I should fix that too. I'm thinking it should be `managedEnvironments.resourceId`.

ninjarobot: Yes, you are right, it's the [managed environment](https://docs.microsoft.com/en-us/azure/templates/microsoft.app/managedenvironments?tabs=json), so this is a lurking bug.

src/Farmer/Arm/App.fs
```diff
@@ -90,9 +91,9 @@ type ContainerApp =
                                                   username = cred.Username
                                                   passwordSecretRef = cred.Username |}
                                            | ImageRegistryAuthentication.ListCredentials resourceId ->
-                                               {| server = ArmExpression.create($"reference({resourceId.ArmExpression.Value}, '2019-05-01').loginServer").Eval()
+                                               {| server = $"{resourceId.Name.Value}.azurecr.io"
                                                   username = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').username").Eval()
-                                                  passwordSecretRef = ArmExpression.create($"listCredentials({resourceId.ArmExpression.Value}, '2019-05-01').username").Eval() |}
+                                                  passwordSecretRef = usernameSecretName resourceId |}
```
ninjarobot, line 96: I take it we just need this for now until [this bug](https://github.com/microsoft/azure-container-apps/issues/153) is fixed, right?

ninjarobot (approved):
The new operations look good, thank you for adding those and the tests to use them.

Are the ACR changes just a temporary workaround?

src/Farmer/Arm/App.fs
isaacabraham: I think so, yes. Although the string concatenation approach should just work, once the underlying ARM provider is fixed I would prefer to revert to the original expression-based approach.

ninjarobot: Sounds good, and given we do t have to change the DSL, this will just keep working when they fix the bug. LGTM.

PR #924, opened 2022-05-03
Don't explicitly create default consumer group for event hubs

PaigeM89:
When creating an `EventHub` resource on the `Basic` tier, there is an error because the generated template attempts to create the Consumer Group `$Default`, but a `Basic` tier is limited to 1 consumer group. However, that `$Default` consumer group is created _automatically_ by Event Hub when it's created, and it doesn't need to be specified in the template.

* Don't create a `$Default` consumer group
* Clarify the default in the docs.

```fsharp
let basicEventHub = eventHub {
    name "test-event-hub"
    sku EventHubSku.Basic
    message_retention_days 1
    partitions 4
}

let deployment = arm {
    location Location.NorthEurope
    add_resource basicEventHub
}
```


isaacabraham:
Thank you very much!

PR #929, opened 2022-05-06
Add Connection Strings members

isaacabraham:

* Adds `ConnectionString` members for App Insights, SignalR and Communication Services
* Clean up Communication Services and fix a bug regarding Location.
* SignalR Key now actually returns the Key.

### DIFF
--- a/src/Farmer/Arm/CommunicationServices.fs
+++ b/src/Farmer/Arm/CommunicationServices.fs
@@ -1,18 +1,17 @@
 [<AutoOpen>]
-module Farmer.Arm.CommunicationServices
+module Farmer.Arm.Communication
 
 open Farmer
 
-let resource = ResourceType ("Microsoft.Communication/communicationServices", "2020-08-20-preview")
+let communicationServices = ResourceType ("Microsoft.Communication/communicationServices", "2020-08-20-preview")
 
-type Resource =
+type CommunicationService =
     { Name: ResourceName
-      Location: Location
       DataLocation: DataLocation
       Tags: Map<string,string> }
     interface IArmResource with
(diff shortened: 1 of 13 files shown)


isaacabraham:
There are a couple of minor breaks here - one is that Communication Service has had a minor rename. It appears that it was never working all this time though as the Location was being set incorrectly, so I don't expect that to harm. SignalR's Key member used to return the ConnectionString, but it now correctly returns the Key.

src/Farmer/Arm/CommunicationServices.fs
```diff
@@ -1,18 +1,17 @@
...
     { Name: ResourceName
-      Location: Location
       DataLocation: DataLocation
       Tags: Map<string,string> }
     interface IArmResource with
-        member this.ResourceId = resource.resourceId this.Name
+        member this.ResourceId = communicationServices.resourceId this.Name
         member this.JsonModel =
-            {| resource.Create(this.Name, this.Location, tags = this.Tags) with
+            {| communicationServices.Create(this.Name, Location.Global, tags = this.Tags) with
```
ninjarobot, line 15: I wonder if it would work before and the behavior changed. I feel we tested this when it was contributed. Anyway, thanks for fixing it.

ninjarobot (approved):
Thanks for adding the expressions to get the connection strings on these so we have a good standard pattern for it.

src/Farmer/Arm/CommunicationServices.fs
isaacabraham: Yes, I am surprised too. I only realised this because I deployed one yesterday and got an error saying Global is the only permitted location.

PR #945, opened 2022-06-16
Fix diagnostics storage account reference for existing storage account

### LINKED ISSUE #939: Farmer generates incorrect ARM expression while adding existing storage account to enable diagnostic logs during VM creation
The Farmer library has builder for Virtual machine creation, though it has support for parameter "diagnostics_support_external" which enables the boot diagnostics with existing created storage account. 
```
let storageAccount =  LinkedResource.Unmanaged resourceId
vm {
  diagnostics_support_external storageAccount
}
```
But it generates the incorrect arm template.
Currently it generates the arm template with following property attached to it.
```
"properties": {
        "diagnosticsProfile": {
          "bootDiagnostics": {
            "enabled": true,
            "storageUri": "[reference('sa0ce3c1e410a5ec454941').primaryEndpoints.blob]"
          }
        },
```
The above generated templates fail to create resources as Azure complains with following error
```
{
  "code": "InvalidTemplate",
  "message": "Deployment template validation failed: 'The template reference 'sa0ce3c1e410a5ec454941' is not valid: could not find template resource or resource copy with this name. Please see https://aka.ms/arm-template-expressions/#reference for usage details.'."
}
```

The correct template should look like this, as after this modification Azure is able to resolve the Storage Account resource.
````
"diagnosticsProfile": {
          "bootDiagnostics": {
            "enabled": true,
            "storageUri": "[reference(concat('Microsoft.Storage/storageAccounts/', 'sa0ce3c1e410a5ec454941'),'2015-06-15').primaryEndpoints.blob]"
          }
        },
```
Please check.

ninjarobot:
This PR closes #939

* Virtual Machines: Fix reference to an existing storage account in boot diagnostics.

Bugfix only, no doc to update. Uses existing unit test.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@ Release Notes
 
 ## 1.7.4
 * Container Apps: Support for mounted storage
+* Virtual Machines: Fix reference to an existing storage account in boot diagnostics.
 * Web App: add an overload for `link_to_service_plan` that accepts Web App
 * Added Basic Types documentation and examples for unmanaged resources.
 
--- a/src/Farmer/Arm/Compute.fs
+++ b/src/Farmer/Arm/Compute.fs
@@ -150,10 +150,17 @@ type VirtualMachine =
                     diagnosticsProfile =
                         match this.StorageAccount with
                         | Some storageAccount ->
+                            let resourceId = storageAccounts.resourceId storageAccount
+                            let storageUriExpr =
+                                ArmExpression
+                                    .reference(storageAccounts, resourceId)
+                                    .Map(fun r -> r + ".primaryEndpoints.blob")
+                                    .WithOwner(resourceId)
+                                    .Eval()
                             box
                                 {| bootDiagnostics =
                                     {| enabled = true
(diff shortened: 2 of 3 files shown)


Vishal7Sharma:
Hi @ninjarobot thanks for looking into this, yes, the above PR solves the issue and generates the correct arm template.

Vishal7Sharma:
Hi @ninjarobot, a minor issue though, now if we pass the parameter "diagnostics_support" without existing storage account, still it generates the arm expression like below. So, I think we have to handle default storage account scenario differently.
```
"diagnosticsProfile": {
          "bootDiagnostics": {
            "enabled": true,
            "storageUri": "[reference(resourceId('Microsoft.Storage/storageAccounts', 'abcdstorage'), '2019-06-01').primaryEndpoints.blob]"
          }
        },
```

Vishal7Sharma:
Hi @ninjarobot, my bad I thought arm expression has to be different for new and existing storage account. I just tested the above changes with new storage account as well and it worked.
So, the above changes are correct for both use cases (existing and new). Thanks

PR #972, opened 2022-09-08
Adding support for route tables and their child resource, routes

thinkdavid:
This PR closes #

* Adding support for route tables
* Adding support for routes

* [x ] **Tested my code** end-to-end against a live Azure subscription.
* [x ] **Updated the documentation** in the docs folder for the affected changes.
* [ x] **Written unit tests** against the modified code that I have made.
* [ x] **Updated the [release notes](RELEASE_NOTES.md)** with a new entry for this PR.
* [x ] **Checked the coding standards** outlined in the [contributions guide](CONTRIBUTING.md) and ensured my code adheres to them.

```fsharp
let deployment =
                    arm {
                        location Location.EastUS
                        add_resources
                            [
                                routeTable {
                                    name "myroutetable"
                                    add_routes [
                                        route {
                                            name "myroute"
                                            addressPrefix "10.10.90.0/24"
                                            nextHopType Route.HopType.VirtualAppliance
                                            nextHopIpAddress "10.10.67.5"
                                        }
                                        route {
                                            name "myroute2"
                                            addressPrefix "10.10.80.0/24"
                                        }
                                    ]
                                }
                            ]
                    }

```

### DIFF
--- a/src/Farmer/Common.fs
+++ b/src/Farmer/Common.fs
@@ -153,7 +153,7 @@ module Mb =
 
 module Route =
     type HopType =
-        | VirtualAppliance
+        | VirtualAppliance of System.Net.IPAddress option
         | Internet
         | Nothing
         | VirtualNetworkGateway
@@ -161,7 +161,7 @@ module Route =
 
         member x.ArmValue =
             match x with
-            | VirtualAppliance -> "VirtualAppliance"
+            | VirtualAppliance _ -> "VirtualAppliance"
             | Internet -> "Internet"
             | Nothing -> "None"
             | VirtualNetworkGateway -> "VirtualNetworkGateway"
--- a/src/Farmer/Common.fs
+++ b/src/Farmer/Common.fs
@@ -153,12 +153,12 @@ module Mb =
 
 module Route =
     type HopType =
-    | VirtualAppliance
-    | Internet
-    | Nothing
-    | VirtualNetworkGateway
-    | VnetLocal
-    with
+        | VirtualAppliance
+        | Internet
+        | Nothing
+        | VirtualNetworkGateway
+        | VnetLocal
+
         member x.ArmValue =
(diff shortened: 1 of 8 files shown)


src/Farmer/Common.fs
```diff
@@ -151,6 +151,22 @@ type EnvVar =
 module Mb =
     let toBytes (mb: int<Mb>) = int64 mb * 1024L * 1024L
 
+module Route =
+    type HopType =
+        | VirtualAppliance
```
ninjarobot, line 156: It might help to make this `VirtualAppliance of IpAddress` to enforce that people assign an IP address in this case (and this case only) since that seems to be how the resource is designed to work:

https://docs.microsoft.com/en-us/azure/templates/microsoft.network/routetables?pivots=deployment-language-arm-template#routepropertiesformat-1



nextHopIpAddress | The IP address packets should be forwarded to. Next hop values are only allowed in routes where the next hop type is VirtualAppliance.
-- | --

ninjarobot (changes requested):
Thanks for the contribution! My only suggestion is to encode the requirement for the nextHopIpAddress in the HopType DU so this can be enforced in the `route` builder and prevent user error.

src/Farmer/Common.fs
thinkdavid: Done!

src/Tests/Network.fs
```diff
@@ -679,13 +679,27 @@ let tests =
                                             route {
                                                 name "myroute"
                                                 addressPrefix "10.10.90.0/24"
-                                                nextHopType Route.HopType.VirtualAppliance
```
ninjarobot: I see...this becomes the default if they specify a next hop address, keeps it simple, thank you!

ninjarobot (approved):

PR #974, opened 2022-09-09
Initial work to move to GitHub Actions

isaacabraham:


.github/workflows/ci-build.yml
```diff
@@ -0,0 +1,33 @@
...
+        uses: actions/checkout@v3
+      - name: Restore .NET Tools
+        run: dotnet tool restore
+      - name: Fantomas Syntax Check
+        run: dotnet fantomas src -r --check
+      - name: Build Farmer
+        run: dotnet build -c release
+      - name: Run Tests
+        run: dotnet test -c release -l trx --collect:"XPlat Code Coverage"
+      - name: Create NuGet Package
```
ninjarobot, line 27: You might want to consider moving this to a different action so it's triggered when releases are created rather than on every push to master. Here's [an example from one of my other projects](https://github.com/ninjarobot/FsCloudInit/blob/main/.github/workflows/pack-and-publish.yml).

ninjarobot (commented):
This looks good - my one suggestion is to have a separate workflow for building and publishing the nuget package so it can be triggered on releases instead of on merges.

martinbryant:
@isaacabraham any chance you will have some time to re visit this one?

PR #981, opened 2022-09-29
Add ability to associate metadata with Storage Account Queues. Metada…

### LINKED ISSUE #916: Support Microsoft.Storage storageAccounts/queueServices/queues - metadata
**template**
resource symbolicname 'Microsoft.Storage/storageAccounts/queueServices/queues@2021-08-01' = {
  name: 'string'
  parent: resourceSymbolicName
  properties: {
    **metadata: {}**
  }
}

https://docs.microsoft.com/en-us/azure/templates/microsoft.storage/storageaccounts/queueservices/queues?tabs=bicep

cmarshall10450:
…ta can be added for a single queue, multiple queues with unique metadata or multiple queues with the same metadata

This PR closes #916 

* Adds ability to add metadata to Storage Account Queues

The unit tests don't pass but I could not work out why. I cannot see a difference in the actual and expected values. 

```fsharp
let storageAccount = storageAccount {
    name = "testaccount"
    add_queue_with_metadata "myqueue" (Map [
        "source", "imageStore"
    ])
}
```

### DIFF
--- a/samples/scripts/storage.fsx
+++ b/samples/scripts/storage.fsx
@@ -7,6 +7,12 @@ let myStorage = storageAccount {
     name "myfarmerstorage"
     sku Storage.Sku.Standard_LRS
     add_queues [ "queue1"; "queue2" ]
+    add_queues_with_same_metadata 
+        ["queue3"; "queue4"]
+        (Map [
+            ("environment", "dev")
+            ("source", "image")
+        ])
     add_private_container "container1"
     add_table "table1"
     add_tables [ "table2"; "table3" ]
--- a/samples/scripts/storage.fsx
+++ b/samples/scripts/storage.fsx
@@ -6,13 +6,22 @@ open Farmer.Builders
 let myStorage = storageAccount {
     name "myfarmerstorage"
     sku Storage.Sku.Standard_LRS
-    add_queues [ "queue1"; "queue2" ]
-    add_queues_with_same_metadata 
-        ["queue3"; "queue4"]
-        (Map [
-            ("environment", "dev")
-            ("source", "image")
-        ])
+    add_queues [
+        storageQueue {
+            name "queue1"
+            metadata [
+                "environment", "dev"
+                "project", "farmer"
+            ]
+        }
+        storageQueue {
+            name "queue2"
+            metadata [
+                "environment", "test"
+                "project", "barnyard"
+            ]
+        }
+    ]
     add_private_container "container1"
     add_table "table1"
     add_tables [ "table2"; "table3" ]
--- a/src/Farmer/Builders/Builders.Storage.fs
+++ b/src/Farmer/Builders/Builders.Storage.fs
@@ -33,6 +33,12 @@ type StoragePolicy =
         Filters: string list
     }
 
+type StorageQueueConfig = 
+    {
+        Name: StorageResourceName
+        Metadata: Metadata option
+    }
+
 type StorageAccountConfig =
     {
         /// The name of the storage account.
@@ -46,7 +52,7 @@ type StorageAccountConfig =
         /// File shares
         FileShares: (StorageResourceName * int<Gb> option) list
         /// Queues
-        Queues: (StorageResourceName * Metadata option) Set
+        Queues: StorageQueueConfig list
         /// Network Access Control Lists
         NetworkAcls: NetworkRuleSet option
         /// Tables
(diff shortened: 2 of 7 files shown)


src/Tests/Storage.fs
```diff
@@ -146,24 +146,127 @@ let tests =
...
+
+                            [
+                                for i in 1..2 do
+                                    account |> getResourceAtIndex client.SerializationSettings i
+                            ]
+
+                        Expect.equal resources.[0].Name "storage/default/queue1" "queue name for 'queue1' is wrong"
+                        Expect.equal resources.[1].Name "storage/default/queue2" "queue name for 'queue2' is wrong"
+                        
+                        let queueMetadata = seq [
```
isaacabraham, line 255: Use `seq { }` for sequence expressions.

samples/scripts/storage.fsx
```diff
@@ -7,6 +7,12 @@ let myStorage = storageAccount {
     name "myfarmerstorage"
     sku Storage.Sku.Standard_LRS
     add_queues [ "queue1"; "queue2" ]
+    add_queues_with_same_metadata 
+        ["queue3"; "queue4"]
+        (Map [
```
isaacabraham, line 12: Do we need the `Map` function call? I would have thought that this can happen in the implementation of the keyword.

isaacabraham (changes requested):

samples/scripts/storage.fsx
cmarshall10450: I didn't really like having the Map there myself but I wasn't sure what the DSL should look like instead. Do you have any recommendations?

Is it possible to just use:
{
    "key" = "value"
}

I didn't think that an anonymous record would work because all keys are chosen by the user, so no specific structure.

ninjarobot: We could pass two lists here - one for the queues and another that is a list of tuples to turn into a map.

But it starts to be confusing to consume when you are passing a couple of lists.

```fsharp
add_queues_with_same_metadata
    [
        "queue3"
        "queue4"
    ]
    [
        "environment", "dev"
        "source", "image"
    ]
```

I feel a builder would be cleaner and work with an overload, letting them define a binding if they want to set the same metadata on multiple queues. Yes, this is longer, but the DSL describes it better.

```fsharp
let devMetadata = [
    "environment", "dev"
    "source", "image"
]

add_queues [
    storageQueue {
        name "queue3"
        metadata devMetadata
    }
    storageQueue {
        name "queue4"
        metadata devMetadata
    }
]
```

martinbryant:
@cmarshall10450 Hi I know this request has been open for a while and would you possibly have time to re-visit this?

cmarshall10450:
Hi @martinbryant

I'll try to take a look at this over the weekend. Is there a consensus on how the DSL should look. I like @ninjarobot's builder proposal. What do you think?

martinbryant:
Hi @cmarshall10450 and thanks for replying.

I would go with @ninjarobot builder suggestion as I think the goal here is to make the DSL clearer.

cmarshall10450:
@martinbryant, I agree. This project is my only experience of F#, so I'll need to get back into the mindset of F# development first.

cmarshall10450:
Hopefully the changes that I have made are in line with what was suggested. I'm not familiar with F# so there will definitely be things to be improved on. I had some issues with getting the tests to pass - expected is same as actual but fails.

martinbryant:
@cmarshall10450 thank you for your changes - hopefully you have seen that the builds fail due to fantomas formatting required - @theprash @isaacabraham could you please have a look at reviewing the changes please

cmarshall10450:
Thanks @martinbryant. I've fixed the formatting issues.

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -648,4 +685,27 @@ type EndpointBuilder with
         let state = this.Origin(state, storage.Endpoint)
         this.DependsOn(state, storage.ResourceId)
 
+type StorageQueueBuilder() =
```
ninjarobot, line 688: Thanks for adding this builder. This makes the usage clearer.

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -298,16 +305,46 @@ type StorageAccountBuilder() =
...
     [<CustomOperation "add_queue">]
     member _.AddQueue(state: StorageAccountConfig, name: string) =
         { state with
-            Queues = state.Queues.Add(StorageResourceName.Create(name).OkValue)
+            Queues =
+                state.Queues
+                @ [
+                    {
+                        Name = StorageResourceName.Create(name).OkValue
+                        Metadata = None
```
ninjarobot, line 322: Thanks for keeping this backwards-compatible so no code breakage on the builders.

ninjarobot (approved):
Thanks for adding the builder for the storage queues - this makes it much clearer when adding metadata.

cmarshall10450:
@ninjarobot All looks good to me. Thanks for fixing the tests.

PR #991, opened 2022-11-24
Possibility of modifying NSG SecurityRule priorities, fix for #990

### LINKED ISSUE #990: Network Security Rules fail to deploy due to too much priority
NSG has security rules and each rule increases the priority by 100.
The max value before ARM fails is 4100 so basically due to Farmer, there is artificial limitation of 41 rules.

You run out of line-numbers if you increment by 100, but if you increment by one, then you cannot easily modify your existing construct. Like in a very old BASIC language!

There are multiple possible solutions:
- Increase less, e.g. per 5 only!
- Custom priority field in securityRule
- Or a priority-incrementor setting in NSG

Thorium:
This PR closes #990

* Changed to NSG Builder only, a new keyword `priority_incr` can be used to replace previously hard-coded value.
* For max backward combability, the default value is the old hard-coded builder value, 100.

* [x] **Tested my code** end-to-end against a live Azure subscription. See below...

**I did test that the values are correct, but this was a builder-side hard-coded value parametrization only, so there is nothing to be added to the data-models or ARM templates, so no changes to deploy to Azure.**

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## 1.7.12
+* NSG: Added option to modify generated SecurityRule priority step increment
 
 ## 1.7.11
 * Dedicated Hosts: Support for Host Groups and Hosts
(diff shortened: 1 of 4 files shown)


Thorium:
This is a very tiny changeset.
The CI failure is definitely not my fault, it fails on dotnet configs and I've not changed anything there.

ninjarobot (approved):
Thank you for the fix and for making it in a backwards compatible way.

PR #1000, opened 2022-12-30
Dedicated support for Startup Command.

### LINKED ISSUE #999: Can't set App Command Line on Linux web app deployments
If you're just deploying a "standard" web app onto Linux, it's not clear how to set the startup command.

isaacabraham:
This PR closes #999 

* Adds a `startup_command` keyword to the Web App builder.

```fsharp
webApp {
    startup_command "foo.exe"
}
```

I'll test it end-to-end shortly. I'm sure I've done Linux deployments in the past so I'm surprised I can't find a clear way of doing this - @ninjarobot do you know any way to already do this in Farmer?

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
+* Web App: Ability to set Startup Command directly (useful for Linux deployments).
 
 ## 1.7.13
 * KeyVault: Added support for disabling public network access.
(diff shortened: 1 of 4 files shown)


ninjarobot (approved):
LGTM.

I haven't done this exactly, but [here's a script](https://gist.github.com/ninjarobot/e4554b7962c13afbfd0fb0a9dd568a1b) I used to test the AppContainer environments early on where it will deploy a container registry and build an image to use for and end to end test. Is that the sort of thing you're looking for @isaacabraham ?

isaacabraham:
Yes, similar to that. The WebApp builder already lets you supply a Docker container and Azure Container registry path, plus the startup command for docker - what I think wasn't available previously was the ability to set the startup command if you just uploaded your app using Zip Deploy (HTTP upload) - in that case, App Service creates a docker container for you with your app, but doesn't know what to use for the startup command so it just sits there doing nothing.

isaacabraham:
@ninjarobot Ok I've remembered how you do it on Linux Web Apps again. If you supply a `web.config` file that contains the startup path, the app service uses that to do it. So this PR isn't really necessary.

isaacabraham:
I don't think it's really needed any more although it doesn't hurt to merge in.

PR #1001, opened 2022-12-31
Update the list of RBAC roles

isaacabraham:

* Update the list of all RBAC roles based on the latest values from the Azure CLI.

To do this, I extracted all roles:

```cmd
az role definition list --custom-role-only false --output json --query '[].{roleName:roleName, description:description, name:name}' > roles.json
```

Then the following script generates the F# that I can copy into the `IdentityExtensions.fs` file:

```fsharp
#r "nuget:FSharp.Data"

open FSharp.Data

type Roles = JsonProvider<"roles.json">

let samples = Roles.GetSamples()

let output = [
    let samples =
        samples
        |> Seq.sortBy (fun s -> s.RoleName)
        |> Seq.map (fun s ->
            {|
                Name = s.Name
                RoleName =
                    s.RoleName.Replace(" ", "").Replace("(Preview)", "").Replace("Deprecated", "")
                    |> Seq.filter System.Char.IsLetterOrDigit
                    |> Seq.toArray
                    |> System.String
                Description = s.Description
                IsPreview = s.RoleName.Contains "Preview"
                IsDeprecated = s.RoleName.Contains "Deprecated"
            |})

    for sample in samples do
    [
        $"""    /// {sample.Description}{if sample.IsPreview then " (This is a preview role)" else ""}"""
        if sample.IsDeprecated then $"""    [<Obsolete "This role has been deprecated.">]"""
        $"""    let {sample.RoleName} = makeRoleId "{sample.RoleName}" "{sample.Name}" """
    ]
]
System.IO.File.WriteAllLines("output.fsx", output)
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
+* Identity: Update the list of all RBAC roles to latest.
 
 ## 1.7.13
 * KeyVault: Added support for disabling public network access.
--- a/src/Farmer/IdentityExtensions.fs
+++ b/src/Farmer/IdentityExtensions.fs
@@ -48,30 +48,13 @@ module Roles =
                 Id = Guid.Parse roleId
             |}
 
-    /// acr push
-    let AcrPush = makeRoleId "AcrPush" "8311e382-0749-4cb8-b61a-304f252e45ec"
+    /// Can customize the developer portal, edit its content, and publish it.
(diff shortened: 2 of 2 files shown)


isaacabraham:
It does appear that a few roles that were previously there are no longer (this was just on a spot check e.g. a SignalR one). Is that to be expected @ninjarobot - or am I perhaps using that AZ CLI command incorrectly and it's returning a filtered list of roles?

PR #1011, opened 2023-02-14
make ARM distinction clearer

et1975:
This PR only touches the front page of the docs to make the distinction between ARM the templating JSON and ARM the orchestrator.


isaacabraham:
Thank you!

PR #1039, opened 2023-05-30
Farmer support for network interface

Huaxinli123:
This PR closes #

* Adding support for azure network interface
* Support both dynamic and static ip allocation

```fsharp
#r "nuget:Farmer"
open Farmer
open Farmer.Builders
open Farmer.Builders.NetworkInterface

arm {
    location Location.EastUS
    add_resources
        [
            vnet {
                name "test-vnet"
                add_address_spaces [ "10.0.0.0/16" ]
            }
            networkInterface {
                name "my-network-interface"
                subnet_prefix "10.0.100.0/24"
                link_to_vnet (virtualNetworks.resourceId "test-vnet")
                add_static_ip "10.0.100.10"
                accelerated_networking_flag false
                ip_forwarding_flag false
            }
        ]
}
```

### DIFF
--- a/src/Farmer/Builders/Builders.NetworkInterface.fs
+++ b/src/Farmer/Builders/Builders.NetworkInterface.fs
@@ -1,0 +1,129 @@
+module Farmer.Builders.NetworkInterface
+
+open Farmer
+open Farmer.Arm
+open Farmer
+open Farmer.Builders
+open Farmer.Network
+open Farmer.Arm.Network
+
+type NetworkInterfaceConfig =
+    {
+        Name: ResourceName
+        AcceleratedNetworkingflag: FeatureFlag option
+        IpForwarding: FeatureFlag option
+        IsPrimary: bool option
+        VirtualNetwork: LinkedResource option
+        SubnetPrefix: IPAddressCidr
+        PrivateIpAddress: string
+        Tags: Map<string, string>
+    }
+
+    interface IBuilder with
+        member this.ResourceId = networkInterfaces.resourceId this.Name
+
+        member this.BuildResources location =
+            [
+                //vnet
+                let vnetId =
+                    this.VirtualNetwork
+                    |> Option.defaultWith (fun _ -> raiseFarmer "Must set 'vnet' for network interface")
+
+                //subnet
+                {
+                    Subnet.Name = ResourceName "networkInterfaceSubnet"
+                    Prefix = IPAddressCidr.format this.SubnetPrefix
+                    VirtualNetwork = Some(vnetId)
+                    NetworkSecurityGroup = None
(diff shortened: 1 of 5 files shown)


src/Farmer/Builders/Builders.NetworkInterface.fs
```diff
@@ -0,0 +1,135 @@
...
+
+type NetworkInterfaceConfig =
+    {
+        Name: ResourceName
+        AcceleratedNetworkingflag: FeatureFlag option
+        IpForwarding: FeatureFlag option
+        IsPrimary: bool option
+        VirtualNetwork: LinkedResource option
+        SubnetPrefix: IPAddressCidr
+        PrivateIpAddress: string
```
ninjarobot, line 18: To avoid a magic string "", this could be `PrivateIpAddress: AllocationMethod` and the default value in the `_.Yield` method could return `AllocationMethod.DynamicPrivateIp`.

src/Farmer/Builders/Builders.NetworkInterface.fs
```diff
@@ -0,0 +1,135 @@
...
+                    AssociatedServiceEndpointPolicies = []
+                    PrivateEndpointNetworkPolicies = None
+                    PrivateLinkServiceNetworkPolicies = None
+                }
+
+                //ipConfig
+                let subnetIpConfigs =
+                    [
+                        {
+                            SubnetName = ResourceName "networkInterfaceSubnet"
```
ninjarobot, line 50: It might be more useful to let them specify the subnet name to attach it rather than hardcode this value.

src/Farmer/Builders/Builders.NetworkInterface.fs
```diff
@@ -0,0 +1,135 @@
...
+
+        member this.BuildResources location =
+            [
+                //vnet
+                let vnetId =
+                    this.VirtualNetwork
+                    |> Option.defaultWith (fun _ -> raiseFarmer "Must set 'vnet' for network interface")
+
+                //subnet
+                {
```
ninjarobot, line 33: Is there any need to generate the subnet or would they normally have created the subnet already and link this NIC to it? Then they could use things like a security group or service delegations on the subnet.

Also since they will probably create many NICs for the same subnet, this would end up generating a subnet resource for each one. Since they are all named the same, the duplicates won't end up in the final template, but probably don't need to be generating them at all.

ninjarobot (commented):
@Huaxinli123 thanks again for this PR. The code, tests, and docs look good, just have a few questions about the functionality.

ninjarobot (approved):
Thanks for addressing those. Once the merge conflicts are resolved, this is ready to merge.

PR #1040, opened 2023-06-02
Fixed scope of RoleAssignment for storage accounts

### LINKED ISSUE #1038: StorageBuilder grants incorrect scope to service principal
I believe I found a bug in using the storage builder's grant_access function. I found this by trying to grant two different system identities access as a storage blob data contributor to two different storage accounts. I ended up with an error of `RoleAssignmentExists`.

I traced the issue down to the following code (line 176 to be specific):

https://github.com/CompositionalIT/farmer/blob/287a0ff31d9f696215c61b90b7eaceedabe0c9ba/src/Farmer/Builders/Builders.Storage.fs#L171-L183

The problem is that the storage builder is granting the role at the resource group level, not the storage account level. So when the ARM template is generated, it leaves out the `scope` field from the JSON. I believe the correct code is: `Scope = SpecificResource this.ResourceId` but have not tested it yet.

BrianVallelunga:
This PR closes #1038

* In builder: role assignment is to the storage account, not the resource group
* Tests updated

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,6 +2,7 @@ Release Notes
 =============
 ## 1.7.23
 * Route server: Adds support for route server creation.
+* Storage Accounts: Fixes scope of role assignments.
 
 ## 1.7.22
 * AVS: Scripting subresource types.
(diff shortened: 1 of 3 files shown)


BrianVallelunga:
I'm not sure why fantomas added the extra line breaks, but the code works.

ninjarobot (approved):
Thanks for fixing the scope on this.

BrianVallelunga:
Great!

PR #1046, opened 2023-06-26
When making app-settings into key-vault secret names sanitize the name.

### LINKED ISSUE #1035: webApp secret_setting for ENV_PARAM with underscores
I am try to add a environment parameter to a ```webApp``` with a app-given name that I cannot change (e.g. ```MY_USER_NAME```). 

KeyVault doesn't allow me to create secrets with names with underscores, so I am wondering how I can set the environment variable to MY_USER_NAME with a value from the KeyVault. 

Is this possible? If so, how?

mike-wade:
This PR closes #1035 

* When creating keyvault secrets for web app appsettings replace characters unsupported in keyvault secret names with `-`.

* Not sure if documentation needs updating? I think this is a bit of a bugfix?
* Will wait for PR to be accepted in principal before doing the remaining tasks.

### DIFF
--- a/src/Farmer/Builders/Builders.KeyVault.fs
+++ b/src/Farmer/Builders/Builders.KeyVault.fs
@@ -70,23 +70,39 @@ type SecretConfig =
             Tags = Map.empty
         }
 
+    static member private allowedKeyNameCharRules = [ Char.IsLetterOrDigit; (=) '-' ]
+
+    static member internal sanitizeKeyName settingKey =
+        match SecretConfig.isValid settingKey with
+        | true -> settingKey
+        | false ->
+            settingKey
+            |> Seq.map (fun c ->
+                match (SecretConfig.allowedKeyNameCharRules |> Seq.exists (fun r -> r c)) with
+                | true -> c
+                | false -> '-')
+            |> String.Concat
+
     static member internal isValid key =
         let charRulesPassed =
-            let charRules = [ Char.IsLetterOrDigit; (=) '-' ]
-            key |> Seq.forall (fun c -> charRules |> Seq.exists (fun r -> r c))
+            key
+            |> Seq.forall (fun c -> SecretConfig.allowedKeyNameCharRules |> Seq.exists (fun r -> r c))
 
         let stringRulesPassed =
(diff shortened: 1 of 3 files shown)


src/Farmer/Builders/Builders.KeyVault.fs
```diff
@@ -70,23 +70,39 @@ type SecretConfig =
...
+            |> Seq.forall (fun c -> SecretConfig.allowedKeyNameCharRules |> Seq.exists (fun r -> r c))
 
         let stringRulesPassed =
             [ (fun l -> String.length l <= 127); String.IsNullOrWhiteSpace >> not ]
             |> Seq.forall (fun r -> r key)
 
-        if not (charRulesPassed && stringRulesPassed) then
+        (charRulesPassed && stringRulesPassed)
+
+    static member internal validate key =
```
ninjarobot, line 97: Why the new method here?

src/Farmer/Builders/Builders.KeyVault.fs
```diff
@@ -70,23 +70,39 @@ type SecretConfig =
             Tags = Map.empty
         }
 
+    static member private allowedKeyNameCharRules = [ Char.IsLetterOrDigit; (=) '-' ]
+
+    static member internal sanitizeKeyName settingKey =
+        match SecretConfig.isValid settingKey with
```
ninjarobot, line 76: Please use if/else instead of matching on true and false.

ninjarobot (commented):
The direction looks good, thanks @mike-wade. Please see comments.

src/Farmer/Builders/Builders.KeyVault.fs
ninjarobot: Yes, I missed the notification before, sorry about that. 

The new static method makes sense, thanks for clarifying.

ninjarobot (approved):
Please add a release notes entry and this is ready to go.

mike-wade:
Thanks @ninjarobot, this worked in my test to a real Azure sub. 

Its worth pointing out that the secret parameter name gets updated with a key-vault friendly parameter name so `csection:secret` needs a param with name of `csection-secret` at deploy time. Is this fine?

PR #1047, opened 2023-06-27
SQL Azure - add support for AD admin

### LINKED ISSUE #1036: Azure SQL Server - Active Directory Admin
Hello and Thanks for the excellent product.

I am setting my Azure SQL server instance with a Active Directory Group as admin and also setting it to AD only auth.

I currently do this as az cli commands. There is a pre and post set of commands to make farmer deploy repeatable (turn AD only off before applying template and subsequently turn on AD only).

I thought it will be useful to add this capability into Farmer.

The specific change would be to enable 
1. creating a Azure SQL instance with Activity Directory admin
2. turn on AD only auth

I tried this with an Arm expression first and it is supported.

I have made a few changes to see if I could figure out the code changes required and everything still seems to hang together. It took me a few hours to make this change as I don't writing F# everyday.

So now I am here to check if this can be added to Farmer. Shall I submit a PR for this change?

gursharan001:
This PR closes #1036 

* introduce new type `ActiveDirectoryAdminSettings`
* change to apply these settings to handle
    * sql admin with no AD admin (existing capability)
    * sql admin with AD admin (new)
    * AD admin only (new)

```fsharp
let activeDirectoryAdmin: ActiveDirectoryAdminSettings =
  {
    Login = "admingroup"
    Sid = "F9D49C34-01BA-4897-B7E2-3694BF3DE2CF"
    PrincipalType = ActiveDirectoryPrincipalType.Group
    AdOnlyAuth = false
  }

sqlServer {
     name "adtestserver"
     active_directory_admin (Some(activeDirectoryAdmin))
     admin_username "sqladminuser"
}
```

## Issues
I have found that ARM templates need modification to switch SQL server configuration among following -
1. sql admin with no AD admin
2. sql admin with AD admin
3. AD admin only

Please refer this blog post - https://www.codez.one/azure-sql-with-managed-identities-part-2/

One possible way forward is to make `AdOnlyAuth` member optional and let the user write correct farmer template by detecting existence of SQL server out of band (using AZ cli or powershell etc)

### DIFF
--- a/src/Farmer/Builders/Builders.Sql.fs
+++ b/src/Farmer/Builders/Builders.Sql.fs
@@ -22,6 +22,7 @@ type SqlAzureConfig =
         Name: SqlAccountName
         AdministratorCredentials: {| UserName: string
                                      Password: SecureParameter |}
+        ActiveDirectoryAdmin: ActiveDirectoryAdminSettings option
         MinTlsVersion: TlsVersion option
         FirewallRules: {| Name: ResourceName
                           Start: IPAddress
@@ -74,10 +75,14 @@ type SqlAzureConfig =
                     ServerName = this.Name
                     Location = location
                     Credentials =
-                        {|
-                            Username = this.AdministratorCredentials.UserName
-                            Password = this.AdministratorCredentials.Password
-                        |}
+                        match this.ActiveDirectoryAdmin  with
+                        | Some(x) when x.AdOnlyAuth -> Unchecked.defaultof<_>
+                        | _ ->
+                            {|
+                                Username = this.AdministratorCredentials.UserName
+                                Password = this.AdministratorCredentials.Password
+                            |}
+                    ActiveDirectoryAdmin = this.ActiveDirectoryAdmin
                     MinTlsVersion = this.MinTlsVersion
                     Tags = this.Tags
                 }
@@ -144,6 +149,7 @@ type SqlAzureConfig =
                                     Username = this.AdministratorCredentials.UserName
                                     Password = this.AdministratorCredentials.Password
                                 |}
+                            ActiveDirectoryAdmin = this.ActiveDirectoryAdmin
                             MinTlsVersion = this.MinTlsVersion
                             Tags = this.Tags
                         }
@@ -277,6 +283,7 @@ type SqlServerBuilder() =
                     UserName = ""
                     Password = SecureParameter ""
                 |}
+            ActiveDirectoryAdmin = None
             ElasticPoolSettings =
                 {|
                     Name = None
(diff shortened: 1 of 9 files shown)


src/Farmer/Builders/Builders.Sql.fs
```diff
@@ -74,10 +75,14 @@ type SqlAzureConfig =
                     ServerName = this.Name
                     Location = location
                     Credentials =
-                        {|
-                            Username = this.AdministratorCredentials.UserName
-                            Password = this.AdministratorCredentials.Password
-                        |}
+                        match this.ActiveDirectoryAdmin with
+                        | Some (x) when x.AdOnlyAuth -> Unchecked.defaultof<_>
```
ninjarobot, line 79: This particular check is in here many times so it would clarify the code to make this an active pattern. For example,

```fsharp
let (|ActiveDirAuth|AdOnlyAuth|StandardAuth|) activeDirAdmin =
    match activeDirAdmin with 
    | Some x when x.AdOnlyAuth -> AdOnlyAuth
    | Some x -> ActiveDirAuth
    | None -> StandardAuth
```
and then the code expresses the intent without having to read through the options and boolean logic:
```fsharp
match this.ActiveDirectoryAdmin with
| ActiveDirAuth -> // whatever happens for AD auth
| AdOnlyAuth -> // whatever happens when it's AD only
| StandardAuth -> // whatever happens when not using AD auth (not sure if it's called standard, this was just an example)
```

src/Tests/Tests.fsproj
```diff
@@ -100,8 +100,8 @@
       <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
       <PrivateAssets>all</PrivateAssets>
     </PackageReference>
-    <PackageReference Include="YoloDev.Expecto.TestSdk" Version="0.13.3" />
-    <PackageReference Update="FSharp.Core" Version="5.0.0" />
+    <PackageReference Include="YoloDev.Expecto.TestSdk" Version="0.14.0" />
+    <PackageReference Update="FSharp.Core" Version="7.0.300" />
```
ninjarobot, line 104: I agree these changes make sense, but are they needed for this PR?

ninjarobot (changes requested):
Thanks again for your continued work on this. I made some recommendations to try to reduce the code complexity a bit. The compiler can deal with it, but it's getting a bit difficult to follow some of the logic for building the various auth settings records.

src/Tests/Tests.fsproj
gursharan001: I wanted to upgrade `Expecto` so that I could write a `testTheory`. The other packages got upgraded as dependencies of `Expecto`

gursharan001:
@ninjarobot I have made further changes to hopefully simplify the code. Please do let me know if it can be made better. Thanks again.

ninjarobot (approved):
Looks great, thanks for simplifying the code a bit!

martin-bmc:
Maybe the wrong place to add this but I just tested this feature and found an issue when changing from an existing Azure AD admin to a new one. It seems like this is not supported in ARM/Bicep:
https://github.com/Azure/bicep/issues/4988

martin-bmc:
As far as I understand the comments on the issue, the Microsoft SQL team allows the value to be set but not changed via the ARM template that is generated in this PR:

> The initial (deployment) template has to have AAD admin set as a property of "Microsoft.Sql/servers" resource. The "update" template would have to have "Microsoft.Sql/servers/administrators" resource.

https://github.com/Azure/bicep/issues/4988#issuecomment-1291611393

It might be to complex for Farmer to handle this case but perhaps it should be mentioned in the documentation?

PR #1059, opened 2023-08-18
AzureWebJobsDashboard only added for V1 function

### LINKED ISSUE #1057: Function builder creates obsolete Functions 1.x "AzureWebJobsDashboard" app setting
It looks like Farmer takes the same storage account specified for `AzureWebJobsStorage` and automatically also uses it for the `AzureWebJobsDashboard` app setting:

https://github.com/CompositionalIT/farmer/blob/7d3d4e82f051aec6ac14c8bd41edd23e794913c7/src/Farmer/Builders/Builders.Functions.fs#L215-L216

When I deploy a function app from a Farmer app that generates this setting, I get the following logs in Application Insights when functions are executed:

> The Dashboard setting is no longer supported. See https://aka.ms/functions-dashboard for details.

and the guidance at that URL says:

> # AzureWebJobsDashboard
> Optional storage account connection string for storing logs and displaying them in the Monitor tab in the portal. This setting is only valid for apps that target version 1.x of the Azure Functions runtime.

as well as:

> For better performance and experience, runtime version 2.x and later versions use `APPINSIGHTS_INSTRUMENTATIONKEY` and App Insights for monitoring instead of `AzureWebJobsDashboard`.

It seems it would be better to either:

* Omit the `AzureWebJobsDashboard` setting completely (considering the latest Functions runtime version is 4.x)
* Or, if compatibility with 1.x is still important, conditionally generate `AzureWebJobsDashboard` only when the chosen runtime is Functions 1.x

I'd be happy to submit a PR for either solution.

martinbryant:
This PR closes #1057 

* Only add the AzureWebJobsDashboard setting to a V1 function app

```fsharp
```

### DIFF
--- a/src/Tests/Functions.fs
+++ b/src/Tests/Functions.fs
@@ -82,6 +82,7 @@ let tests =
                     functions {
                         name "test"
                         link_to_unmanaged_storage_account externalStorageAccount
+                        use_extension_version V1
                     }
 
                 let f = functionsBuilder :> IBuilder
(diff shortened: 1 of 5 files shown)


src/Tests/Functions.fs
```diff
@@ -416,16 +417,39 @@ let tests =
...
                         "WEBSITE_CONTENTAZUREFILECONNECTIONSTRING"
                         "WEBSITE_CONTENTSHARE"
                     ]
-                    |> List.map (settings.ContainsKey)
+                    |> List.map settings.ContainsKey
 
                 Expect.allEqual expectation true "Slot should have all literal settings"
             }
 
+            test "Functions App generates AzureWebJobsDashboard setting on version 1" {
```
isaacabraham, line 429: Could we have a test to prove it's *not* set on v2?

ninjarobot: Thanks for adding the test @martinbryant. 

@isaacabraham any other concerns with this one?

src/Tests/test-data/lots-of-resources.json
```diff
@@ -283,7 +283,7 @@
             },
             {
               "name": "FUNCTIONS_EXTENSION_VERSION",
-              "value": "~4"
```
ninjarobot, line 286: It doesn't have to be in this PR, but we probably should add a test that resources are generated as expected for a functionapp with the most recent version of the runtime.

martinbryant: @ninjarobot as there was a test data further down the file for the latest version for a function app and did not add another one - is that ok?

ninjarobot: Yes, that works, I wasn't sure if we had one, but I just wanted to make sure we cover the v4 scenario.

PR #1066, opened 2023-09-21
Fix a bug where extensions object is added outside the properties block of the container group

### LINKED ISSUE #1065: Bug - Recent feature for adding extensions object adds it outside the properties block
It was found that recent feature for adding extensions object resided outside of properties block.

ahmed-ebaid:
Fix a bug where extensions block is added outside the properties block of the container group
This PR closes #1065

* Add extensions object in side properties block of a container group

```fsharp
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,8 @@
 Release Notes
 =============
+## 1.7.29
+* Bug fix - Add Container Group extensions object under properties block of Container Group
+
 ## 1.7.28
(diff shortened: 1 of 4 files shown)


ninjarobot (approved):
Thanks for the fix @ahmed-ebaid. It has me thinking we need a structured object around the outer part of the JsonModel so `IArmResource` implementations only modify the `properties`. Not sure how to get that in place yet, but that would prevent this issue. I'm open to ideas.

PR #1095, opened 2024-02-02
Create b2c tenant

amine-mejaouel:
This PR closes #

* Allow to create B2c Tenant
* Please note that this does not work for B2C tenant update as it requires another arm template.

```fsharp
open Farmer
open Farmer.Builders
open Farmer.Deploy

let initialDomainName = "myb2c"

let myb2c =
    b2cTenant {
        initial_domain_name initialDomainName
        display_name "My B2C"
        sku B2cTenant.Sku.PremiumP1
        country_code "FR"
        data_residency B2cDataResidency.Europe
    }
    
let b2cDoesNotExist (initialDomainName: string) =
    let output =
        Az.AzHelpers.executeAz $"resource list --name '{initialDomainName}.onmicrosoft.com'"
        |> snd
    not (output.Contains initialDomainName)
    
let deployment =
    arm {
        location Location.FranceCentral
        add_resources
            [
                // This allows to avoid having an error when the B2C tenant already exists
                if b2cDoesNotExist initialDomainName then
                    myb2c
            ]
    }

```

### DIFF
--- a/src/Farmer/Builders/Builders.B2cTenant.fs
+++ b/src/Farmer/Builders/Builders.B2cTenant.fs
@@ -1,0 +1,98 @@
+[<AutoOpen>]
+module Farmer.Builders.B2cTenant
+
+open Farmer
+open Farmer.Arm
+open Farmer.Validation
+
+type B2cDomainName with
+
+    static member FromInitialDomainName initialDomainName =
+        [ containsOnlyM [ lettersOrNumbers ] ]
+        |> validate "B2c initial domain name" initialDomainName
+        |> Result.map (fun x -> B2cDomainName $"{x}.onmicrosoft.com")
+
+/// Check official documentation for more details: https://learn.microsoft.com/en-us/azure/active-directory-b2c/data-residency#data-residency
+type B2cDataResidency =
+    | UnitedStates
+    | Europe
+    | AsiaPacific
+    | Japan
+    | Australia
+
+    member this.Location =
+        match this with
+        | UnitedStates -> Location "United States"
+        | Europe -> Location "Europe"
+        | AsiaPacific -> Location "Asia Pacific"
+        | Japan -> Location "Japan"
+        | Australia -> Location "Australia"
+
+type B2cTenantConfig =
+    {
+        Name: B2cDomainName
+        DisplayName: string
+        DataResidency: Location
+        CountryCode: string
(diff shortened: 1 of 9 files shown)


src/Farmer/Builders/Builders.B2cTenant.fs
```diff
@@ -0,0 +1,98 @@
...
+
+type B2cDomainName with
+
+    static member FromInitialDomainName initialDomainName =
+        [ containsOnlyM [ lettersOrNumbers ] ]
+        |> validate "B2c initial domain name" initialDomainName
+        |> Result.map (fun x -> B2cDomainName $"{x}.onmicrosoft.com")
+
+/// Check official documentation for more details: https://learn.microsoft.com/en-us/azure/active-directory-b2c/data-residency#data-residency
+type B2cDataResidency =
```
ninjarobot, line 16: Can this be moved to `Common.fs` in the `B2cTenant` module alongside the SKU? Then users can `open Farmer.B2cTenant` and have the types they need.

ninjarobot (changes requested):
Thanks for the contribution @amine-mejaouel, this looks great. I have a minor recommendation to improve usage by putting both `Sku` and `B2cDataResidency` in the same module in `Common` to enable people to `open` it and get the related types.

src/Farmer/Builders/Builders.B2cTenant.fs
amine-mejaouel: I tried moving it, but I  saw that the user still needs to open the `Farmer.Builders` namespace in order to have access to the `b2cTenant`  builder.
Do you think that we still need to move it for other reasons that I may be ignoring ?

ninjarobot: It's not specific to the builder, so we generally try to keep those types in Common where they can be used in both builders and IArmResource. [Example of a very similar one](https://github.com/CompositionalIT/farmer/blob/e3e387e81834ce0c3a30427ac02f2679db90c89b/src/Farmer/Common.fs#L3700).

amine-mejaouel: Okay, thanks for the clarification, I updated the code accordingly.

ninjarobot (approved):
Thanks for moving the types to Common.

PR #1114, opened 2024-06-14
Support for App Service Linux automatic instrumentation.

### LINKED ISSUE #1113: App Service does not support Linux App Insights automatic instrumentation

isaacabraham:
This PR closes #1113

* Support for App Service automatic instrumentation in Linux.

```fsharp
let myWebApp = webApp {
    name "mywebapp"
    operating_system Linux
}
```

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,7 @@
 Release Notes
 =============
+## vNext
+* App Service: Automatic support for App Insights on Linux.
 
 ## 1.8.11
 * Service Bus: Added support for setting max message size.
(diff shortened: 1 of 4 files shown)


isaacabraham:
@ninjarobot I'm concerned about this being a breaking change. We can mitigate that with a warning (and maybe bump to 1.9) but should I add an extra setting on app service to disable automatic instrumentation? I can't imagine why that wouldn't be desired, but perhaps?

isaacabraham:
True. I think the worst case would be that the environment suddenly gets some extra settings - I don't see that as being dangerous though?

PR #1118, opened 2024-06-21
[Load Balancer] Using Subnets instead of VirtualNetworks to link per IP configured in the BackendAddressPool

### LINKED ISSUE #1117: No builder support for adding subnet per IP of LoadBalancer backendAddressPool
Currently there is no support to add the subnet of the IP address getting added in the backend address pool of the load balancer. 

Instead, all the IP addresses are getting added as a list and getting linked to the vnet in the arm deployment - which ideally should be configured to also support getting linked to subnets as well.

memeghaj:
This PR closes #1117 

* Using subnets instead of vnets in the backendAddressPool config of the LoadBalancer.
* Changes to existing builder structs and constructors for backendAddressPool.
* Changes to functionalities to add subnets and link to every IP configured in the backendAddressPool.

```fsharp
backendAddressPool {
      name "lb-backend"
      vnet "my-vnet"
      subnet "my-subnet"
      add_ip_addresses [
          "10.0.1.4"
          "10.0.1.5"
      ]
 }
```


src/Farmer/Arm/LoadBalancer.fs
```diff
@@ -187,10 +188,10 @@ type BackendAddressPool =
                                     properties =
                                         {|
                                             ipAddress = string addr.IpAddress
-                                            virtualNetwork =
```
ninjarobot, line 190: If someone passes a vnet instead of a subnet, this will just ignore it, although this does still work. Please make sure not to break compatibility with existing functionality and continue to emit these properties.

memeghaj:
The resulting template JSON was tested on a Azure Sub and checked the deployment which was successful.

src/Farmer/Arm/LoadBalancer.fs
memeghaj: The use case that we want to address through this change, is actually failing if both `subnet` and `vnet` are present on the deployment per IP.
It only serves our purpose if only subnet is mentioned along with IP.

^^ Tested this scenario already today.

ninjarobot: I understand _your_ use case is to specify the subnet, however everyone else who is already using the library to create a backend address pool is using the vnet. As it is implemented right now, this break existing use cases as they will just silently stop emitting the virtual network in the backend address pool.

You can make this a non-breaking change by specifying the subnet **name**, and if set, it will emit a full subnet Id and no vNet Id, otherwise it will leave it as just the vNet Id.

memeghaj: Addressed the same.

ninjarobot (approved):
Looks great @memeghaj - thanks for addressing the backwards compatibility.

PR #1121, opened 2024-07-02
Fixes PgSql ResourceId. Also small _. refactor.

### LINKED ISSUE #1120: PostgreSQL dependency generates incorrect path
Given this code:

```fsharp
let pgServer =
    postgreSQL {
        name "pgserver"
        add_database "foo"
        admin_username "theadmin"
    }

let app =
    webApp {
        name "mywebapp"
        depends_on pgServer
    }
```

The following ARM is generated:

```json
"apiVersion": "2021-03-01",
"dependsOn": [
    "[resourceId('Microsoft.DBforPostgreSQL/servers/databases', 'pgserver')]",
],
"kind": "app",
"location": "westeurope",
"name": "mywebapp",
```

The offending line

```json
"[resourceId('Microsoft.DBforPostgreSQL/servers/databases', 'pgserver')]",
```

fails: `the type 'Microsoft.DBforPostgreSQL/servers/databases' requires '2' resource name argument(s)` which makes sense - you need to either supply the DB name as well as the server, OR change the resource Id to just the server (remove the `/databases` bit.

isaacabraham:
This PR closes #1120 

* Fixes ResourceId in PgSql to be the server, NOT the database resource type.
* Small refactorings to go from `fun x -> x.Blah` to just `_.Blah`

* No documentation to update


src/Farmer/Builders/Builders.PostgreSQL.fs
```diff
@@ -46,7 +45,7 @@ type PostgreSQLConfig = {
 } with
 
     interface IBuilder with
```
isaacabraham, line 47: This is the key actual change for this PR.

src/Tests/PostgreSQL.fs
```diff
@@ -464,4 +463,42 @@ let tests =
             Expect.equal PostgreSQLFamily.Gen5.AsArmValue "Gen5" "Wrong value for Gen5 family"
             Expect.equal (PostgreSQLFamily.Gen5.ToString()) "Gen5" "Wrong value for Gen5 family"
         }
+
```
isaacabraham, line 466: New unit test

ninjarobot (approved):
Thanks for fixing that. I started to do some cleanup on pgsql and got totally derailed when I read [this](https://learn.microsoft.com/azure/postgresql/single-server/whats-happening-to-postgresql-single-server), so I haven't come back to it. Much of the existing builder can probably be mapped to the new flexible server, but I'm not sure it's worth the trouble.

PR #1126, opened 2024-07-11
Add support for Flexible Servers, databases and firewalls.

### LINKED ISSUE #1123: Migrate to PostgreSQL Flexible Database
Single Instance DBs are being phased out and should be replaced with Flexible Databases. This could be a breaking change to Farmer?

isaacabraham:
This PR closes #1123 

* Adds support for Flexible Servers, databases and firewalls.

```fsharp
open Farmer
open Farmer.Builders
open Farmer.PostgreSQL

let myPostgres = postgreSQL {
    name "aserverformultitudes42"
    admin_username "adminallthethings"
    storage_size 64<Gb>
    add_database "my_db"
    enable_azure_firewall
    storage_autogrow true

    // overloaded or model-specific keywords
    tier FlexibleTier.Burstable_B1ms
    server_version FlexibleVersion.V_16
    storage_performance_tier Vm.DiskPerformanceTier.P10
}
```

The main potential concern for this PR is that it adds support for the new Flexible server but allows the old Single Server model to still work. I propose we mark the old one as obsolete and at some point in the future completely decomission it. This will also simplify the codebase.

### DIFF
--- a/src/Farmer/Builders/Builders.PostgreSQL.fs
+++ b/src/Farmer/Builders/Builders.PostgreSQL.fs
@@ -10,11 +10,16 @@ open Farmer.PostgreSQL
 open Servers
 
 type ServerKind =
-    | SingleInstance of
+    | SingleServer of
         {|
             Version: Version
             Sku: Sku
             Capacity: int<VCores>
+            VirtualNetworkRules:
+                {|
+                    Name: ResourceName
+                    VirtualNetworkSubnetId: ResourceId
+                |} list
         |}
     | Flexible of FlexibleVersion * FlexibleTier
 
@@ -43,11 +48,6 @@ type PostgreSQLConfig = {
             Start: IPAddress
             End: IPAddress
         |} list
-    VirtualNetworkRules:
-        {|
-            Name: ResourceName
-            VirtualNetworkSubnetId: ResourceId
-        |} list
     Tags: Map<string, string>
 } with
 
@@ -57,11 +57,11 @@ type PostgreSQLConfig = {
         member this.BuildResources location = [
             let serverType, firewallRulesType, databasesType =
                 match this.Kind with
-                | SingleInstance _ -> servers, firewallRules, databases
+                | SingleServer _ -> servers, firewallRules, databases
                 | Flexible _ -> flexibleServers, flexibleFirewallRules, flexibleDatabases
 
             match this.Kind with
-            | SingleInstance config ->
+            | SingleServer config ->
                 {
                     Name = this.Name
                     Location = location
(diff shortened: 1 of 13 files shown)


isaacabraham:
The problem with the `Obsolete` attribute is that it makes warnings show up in the Farmer codebase itself!

isaacabraham:
I fixed the warnings last night, either by selectively choosing which elements to warn, and in the tests using the `#nowarn` directive directly within the Postgres tests.

src/Farmer/Builders/Builders.PostgreSQL.fs
```diff
@@ -405,18 +506,25 @@ type PostgreSQLBuilder() =
         }
 
     /// Adds a custom vnet rule given a name and a virtualNetworkSubnetId.
+    [<Obsolete(SingleServerObsoleteMessage)>]
     [<CustomOperation "add_vnet_rule">]
```
ninjarobot, line 510: This might be a place to support linking to a delegated subnet to make the function work for flexible servers. I agree this is a little different than firewall rules and would be cleaner with a new builder, but at least it could provide a way to make the Postgres server accessible to the vnet with this builder.


https://learn.microsoft.com/azure/templates/microsoft.dbforpostgresql/flexibleservers#network

ninjarobot (approved):
Thanks for adding this support for flexible servers while keeping g compatibility. I agree with your point that a new builder would simplify the codebase and especially looking at the vnet integration, it would make it clearer.

src/Farmer/Builders/Builders.PostgreSQL.fs
isaacabraham: Ah so this is now just a part of the Server itself rather than a separate virtualnetwork resource. I'll check this and update if possible. Worse case we merge like this and then add VNet support at a later date?

isaacabraham: @ninjarobot do you think that in the Flexible Server world, `delegatedSubnetResourceId` maps to what was `virtualNetworkSubnetId`?

ninjarobot: We can merge as is, I'm just trying to think how to support vnet connections in this case.

The more I think about it, adding a vnet rule is probably too different from the delegated subnet case. Adding a vnet rule allows resources that are in that vnet to be able to reach the Postgres server's public IP address whereas a delegated subnet actually attaches the server in that subnet directly and it gets a private IP address in the subnet. 

So I think it would be good to merge as is and come back to the delegated subnet, potentially with a new builder. Thoughts on that?

isaacabraham: Yes, we probablly need to think about that more. New builder - yes, we probably would then remove / replace the current one completely (or at least delete the obsolete methods + types).

isaacabraham:
> What do you think - a new minor version?
> 
>     * 1.8.14
> 
>     * 1.9.0

I think 1.9.0?

PR #1134, opened 2024-08-22
Attaching a containerGroup to a vnet no longer requires a profile

et1975:
This PR updates the condition for deploying ACI with a custom DNS config to account for new profile-less subnet option.

* Allow DNS if either Network Profile or a VNet are specified.

Seems straight-forward?

in  a `containerGroup` CE, specifying this combination of options:

```fsharp
        subnet subnet
        link_to_vnet vnetName
        dns_search_domains [domain]
```
should not throw "DNS configuration can only be set when attached to a virtual network."

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -3,6 +3,7 @@ Release Notes
 
 ## 1.9.2
 * Container Apps: Fix to container registry credential to not emit a secret for a managed identity.
+* ContainerGroup followup to #ff78f202dc - expand DNS config validation for profile-less vnet.
 
 ## 1.9.1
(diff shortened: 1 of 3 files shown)


et1975:
With profile properties now available on the ACI itself one can deploy a VNET-attached container by specifying
`subnet` and `link_to_vnet` instead of referencing a profile.
The existing validation would allow you to deploy this configuration, but attempting to specify DNS search in this perfectly valid scenario would cause Farmer to throw in misguided assumption that we are not attaching to a VNET because the profile is not specified.
It probably needs a deeper refactoring, but this change would unblock specifying DNS search if either attachment mechanism is used.

ninjarobot (approved):
Thanks for adding the tests!

PR #1135, opened 2024-08-27
Fix container-apps.md formatting

mattgallagher92:
Adds a line break before a table to (hopefully) fix the docs site (untested).

docs page showing plaintext markdown table syntax rather than a rendered table


ninjarobot (approved):

mattgallagher92:
Thanks @ninjarobot, seems to have worked:

docs page showing a rendered table

PR #1137, opened 2024-08-28
Container Apps MSI fix

### LINKED ISSUE #1136: Container apps using user assigned identity (MSI) for ACR authentication - template broken
### Issue No 1:
Generated template for containerapp looks as below:

```
      "properties": {
        "configuration": {
          "registries": [
            {
              "identity": "[resourceId('<GUID>', 'TestMSIrg', 'Microsoft.ManagedIdentity/userAssignedIdentities', 'TestMSI')]",
              "server": "testacr.azurecr.io",
              "username": ""
            }
          ],
          "secrets": [
            {
              "name": "testacr.azurecr.io",
              "value": "[resourceId('<GUID>', 'TestMSIrg', 'Microsoft.ManagedIdentity/userAssignedIdentities', 'TestMSI')]"
            }
          ]
        },
```
The deployment error is:
_Status Message: Secret name has an invalid value 'testacr.azurecr.io'. A value must consist of lower case alphanumeric characters, '-', and must start and end with an alphanumeric character. The length must not be more than 253 characters. (Code:ContainerAppInvalidSecretName)_
See
https://learn.microsoft.com/en-us/azure/container-apps/azure-resource-manager-api-spec?tabs=arm-template
for template reference

**Fix:** when using MSI, "secrets" should not be populated

### Issue No 2
To be able to use MSI for registry authentication, the containerapp has to have a user assigned identity, this is done by add_identity.
add_identity will add the MSI to the dependency list.
If the MSI is not defined in the template, the deployment validation fails:
_New-AzResourceGroupDeployment:  - Error: Code=InvalidTemplate; Message=Deployment template validation failed: 'The resource 'Microsoft.ManagedIdentity/userAssignedIdentities/TestMSI' is not defined in the template. Please see https://aka.ms/arm-syntax for usage details.'._

**Fix:** remove the dependency on the MSI being defined in the template

PR with fix will follow shortly

mafteiu:
This PR closes #1136

* modify App.fs to properly generate ARM template for container registry authentication with user assigned identity:
    * generated "secrets" is an empty array

Not necessary to update documentation - no new feature
Unit tests pass with no change
No new feature
```fsharp
```

### DIFF
--- a/src/Farmer/Arm/App.fs
+++ b/src/Farmer/Arm/App.fs
@@ -160,7 +160,6 @@ type ContainerApp = {
                 | KeyValue(name, Volume.AzureFileShare(_)) ->
                     storages.resourceId (this.Environment.Name, ResourceName name) |> Some
                 | _ -> None)
-        yield! this.Identity.Dependencies
     ]
 
     member private this.ResourceId = containerApps.resourceId this.Name
@@ -215,13 +214,7 @@ type ContainerApp = {
                                                 )
                                                 .Eval()
                                       |}
-                                    | ImageRegistryAuthentication.ManagedIdentityCredential cred -> {|
-                                        name = cred.Server
-                                        value =
-                                            match cred.Identity.Dependencies with
-                                            | [] -> String.Empty
-                                            | primaryDependency :: _ -> primaryDependency.ArmExpression.Eval()
-                                      |}
+                                    | ImageRegistryAuthentication.ManagedIdentityCredential cred -> ()
                                 for setting in this.Secrets do
                                     {|
                                         name = setting.Key.Value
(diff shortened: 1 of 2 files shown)


fahedmarakbi (approved):

src/Farmer/Arm/App.fs
```diff
@@ -160,7 +160,6 @@ type ContainerApp = {
                 | KeyValue(name, Volume.AzureFileShare(_)) ->
                     storages.resourceId (this.Environment.Name, ResourceName name) |> Some
                 | _ -> None)
-        yield! this.Identity.Dependencies
```
ninjarobot, line 163: This line should stay here so that MSI's defined in the same template will be deployed before this ContainerApp. If you're assigning an existing user assigned identity, use `link_to_identity` instead of `add_identity` and then the dependency won't be emitted here.

ninjarobot (changes requested):
The fix for the secret looks good, but `this.Identity.Dependencies` seems it should stay so that an MSI created in the same deployment will be created before this resource. Using an existing MSI can be done with `link_to_identity`.

ninjarobot (approved):

src/Farmer/Arm/App.fs
ninjarobot: Thanks for leaving that in.

PR #1144, opened 2024-10-24
Add support for Immutable Storage on container

xperiandri:
Implemented:
* `ImmutableStorageWithVersioning`
* `RequireInfrastructureEncryption`This PR closes #

Which test would you recommend to modify or implement?

```fsharp
let storage = storageAccount {
    name "wormtest222"
    sku Farmer.Storage.Sku.Standard_GRS
    disable_blob_public_access
    add_private_container
        "container"
        (blobContainerImmutabilityPolicies {
            allow_protected_append_writes AllAppendAllowed
            immutability_period_since_creation (365<Days> * 5)
        })
}

```

### DIFF
--- a/src/Farmer/Arm/Storage.fs
+++ b/src/Farmer/Arm/Storage.fs
@@ -5,7 +5,7 @@ open Farmer
 open Farmer.Storage
 
 let storageAccounts =
-    ResourceType("Microsoft.Storage/storageAccounts", "2022-05-01")
+    ResourceType("Microsoft.Storage/storageAccounts", "2023-05-01")
 
 let blobServices =
     ResourceType("Microsoft.Storage/storageAccounts/blobServices", "2019-06-01")
@@ -38,6 +38,15 @@ let roleAssignments =
     ResourceType("Microsoft.Storage/storageAccounts/providers/roleAssignments", "2018-09-01-preview")
 
 type Metadata = Map<string, string>
+
+type ImmutabilityPolicyState =
+    | Unlocked
+    | Locked
+
+    member this.ArmValue =
+        match this with
+        | Unlocked -> "Unlocked"
+        | Locked -> "Locked"
 
 [<RequireQualifiedAccess>]
 type NetworkRuleSetBypass =
@@ -98,21 +107,32 @@ type StorageAccount = {
     Location: Location
     Sku: Sku
     Dependencies: ResourceId list
-    EnableHierarchicalNamespace: bool option
-    NetworkAcls: NetworkRuleSet option
     StaticWebsite:
         {|
             IndexPage: string
             ErrorPage: string option
             ContentPath: string
         |} option
-    MinTlsVersion: TlsVersion option
-    SupportsHttpsTrafficOnly: FeatureFlag option
-    DnsZoneType: string option
+    EnableHierarchicalNamespace: bool option
+    DefaultToOAuthAuthentication: FeatureFlag option
     DisablePublicNetworkAccess: FeatureFlag option
     DisableBlobPublicAccess: FeatureFlag option
     DisableSharedKeyAccess: FeatureFlag option
-    DefaultToOAuthAuthentication: FeatureFlag option
+    DnsZoneType: string option
+    ImmutableStorageWithVersioning:
+        {|
+            Enable: bool option
+            ImmutabilityPolicy: {|
+                AllowProtectedAppendWrites: bool option
+                ImmutabilityPeriodSinceCreationInDays: int option
+                State: ImmutabilityPolicyState option
+            |} option
+        |} option
+    MinTlsVersion: TlsVersion option
+    NetworkAcls: NetworkRuleSet option
+    /// <remarks>Azure default is false</remarks>
+    RequireInfrastructureEncryption: bool option
+    SupportsHttpsTrafficOnly: FeatureFlag option
     Tags: Map<string, string>
 } with
 
@@ -149,15 +169,13 @@ type StorageAccount = {
                     | Blobs _ -> "BlobStorage"
                     | Files _ -> "FileStorage"
                     | BlockBlobs _ -> "BlockBlobStorage"
+                extendedLocation = "" // TODO:
+                identity = "" // TODO: user assigned identityt
                 properties = {|
-                    isHnsEnabled = this.EnableHierarchicalNamespace |> Option.toNullable
                     accessTier =
                         match this.Sku with
                         | Blobs(_, Some tier)
-                        | GeneralPurpose(V2(_, Some tier)) ->
-                            match tier with
-                            | Hot -> "Hot"
-                            | Cool -> "Cool"
+                        | GeneralPurpose(V2(_, Some tier)) -> tier.ArmValue
                         | _ -> null
                     networkAcls =
                         this.NetworkAcls
@@ -182,41 +200,30 @@ type StorageAccount = {
                             defaultAction = networkRuleSet.DefaultAction.ArmValue
                         |})
                         |> Option.defaultValue Unchecked.defaultof<_>
-                    minimumTlsVersion =
-                        match this.MinTlsVersion with
-                        | Some Tls10 -> "TLS1_0"
-                        | Some Tls11 -> "TLS1_1"
-                        | Some Tls12 -> "TLS1_2"
-                        | None -> null
-                    supportsHttpsTrafficOnly =
-                        match this.SupportsHttpsTrafficOnly with
-                        | Some FeatureFlag.Disabled -> "false"
-                        | Some FeatureFlag.Enabled -> "true"
-                        | None -> null
-                    dnsEndpointType =
-                        match this.DnsZoneType with
-                        | Some s -> s
-                        | None -> null
-                    publicNetworkAccess =
-                        match this.DisablePublicNetworkAccess with
-                        | Some FeatureFlag.Disabled -> "Enabled"
-                        | Some FeatureFlag.Enabled -> "Disabled"
-                        | None -> null
-                    allowBlobPublicAccess =
-                        match this.DisableBlobPublicAccess with
-                        | Some FeatureFlag.Disabled -> "true"
-                        | Some FeatureFlag.Enabled -> "false"
-                        | None -> null
-                    allowSharedKeyAccess =
-                        match this.DisableSharedKeyAccess with
-                        | Some FeatureFlag.Disabled -> "true"
-                        | Some FeatureFlag.Enabled -> "false"
-                        | None -> null
-                    defaultToOAuthAuthentication =
-                        match this.DefaultToOAuthAuthentication with
-                        | Some FeatureFlag.Disabled -> "false"
-                        | Some FeatureFlag.Enabled -> "true"
-                        | None -> null
+                    allowBlobPublicAccess = this.DisableBlobPublicAccess.BooleanValue
+                    allowSharedKeyAccess = this.DisableSharedKeyAccess.BooleanValue
+                    defaultToOAuthAuthentication = this.DefaultToOAuthAuthentication.BooleanValue
+                    dnsEndpointType = this.DnsZoneType |> Option.toObj
+                    encryption = {|
+                        requireInfrastructureEncryption = this.RequireInfrastructureEncryption |> Option.toNullable
+                    |}
+                    immutableStorageWithVersioning =
+                        this.ImmutableStorageWithVersioning
+                        |> Option.map (fun immutableStorage -> {|
+                            enable = immutableStorage.Enable |> Option.toNullable
+                            policy =
+                                immutableStorage.ImmutabilityPolicy
+                                |> Option.map (fun immutableStorage ->
+                                    {|
+                                        allowProtectedAppendWrites = immutableStorage.AllowProtectedAppendWrites |> Option.toNullable
+                                        immutabilityPeriodSinceCreationInDays = immutableStorage.ImmutabilityPeriodSinceCreationInDays |> Option.toNullable
+                                        state = immutableStorage.State |> Option.map _.ArmValue |> Option.toObj
+                                    |})
+                        |})
+                    isHnsEnabled = this.EnableHierarchicalNamespace |> Option.toNullable
+                    minimumTlsVersion = this.MinTlsVersion.ArmValue ()
+                    publicNetworkAccess = this.DisablePublicNetworkAccess.ArmValue ()
+                    supportsHttpsTrafficOnly = this.SupportsHttpsTrafficOnly.BooleanValue
                 |}
         |}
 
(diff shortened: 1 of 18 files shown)


xperiandri:
@ninjarobot do you have any remarks?

src/Farmer/Arm/Storage.fs
```diff
@@ -182,41 +206,33 @@ type StorageAccount = {
...
+                                        immutableStorage.AllowProtectedAppendWrites |> Option.toNullable
+                                    immutabilityPeriodSinceCreationInDays =
+                                        immutableStorage.ImmutabilityPeriodSinceCreation |> Option.toNullable
+                                    state = immutableStorage.State |> Option.map _.ArmValue |> Option.toObj
+                                |})
+                        |})
+                    isHnsEnabled = this.EnableHierarchicalNamespace |> Option.toNullable
+                    minimumTlsVersion = this.MinTlsVersion.ArmValue()
+                    publicNetworkAccess = this.DisablePublicNetworkAccess.ArmInvertedValue()
+                    supportsHttpsTrafficOnly = this.SupportsHttpsTrafficOnly.AsBoolean()
```
xperiandri, line 235: I ordered them by name as otherwise it is hard to compare to the specification

isaacabraham:
You want to write at least one test confirmation the JSON that is emitted in the generate ARM template is correct i.e. checking fields etc.

isaacabraham:
From an API design, what happens if you don't want to set any policies - do you need to provide a policy regardless (bear in mind computation expressions don't allow optional arguments AFAIK, and overloads must all have the same number of arguments)? I think perhaps creating a dedicated `container { }` CE which can have members on it including e.g.:

* `name`
* `accessibility` (public / private)
* `immutability_policy`

would be a better way to go.

isaacabraham:
Also can you make an issue first which clearly outlines what this PR is designed to solve - normally we create issues, get them approved, and then create a PR to resolve them (in that order :-))

src/Farmer/Types.fs
```diff
@@ -416,11 +417,40 @@ type FeatureFlag =
...
         | Enabled -> "Enabled"
         | Disabled -> "Disabled"
 
+    member this.ArmInvertedValue =
+        match this with
+        | Enabled -> "Disabled"
+        | Disabled -> "Enabled"
+
+[<AbstractClass; Sealed; Extension>]
+type FeatureFlagExtensions =
```
ninjarobot, line 436: Nice use of extensions to reduce the boilerplate with the `FeatureFlag option` type!

ninjarobot (approved):
Looks great, thanks for adding this support for newer storage account functionality. I made a few minor suggestions you can take or leave as applicable.

isaacabraham (changes requested):
Before merging, we need to modify the API - the use of a CE in this regard isn't going to work, from what I can see, from a usability perspective.

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -125,23 +135,35 @@ type StorageAccountConfig = {
...
-                SupportsHttpsTrafficOnly = this.SupportsHttpsTrafficOnly
-                DnsZoneType = this.DnsZoneType
+                DefaultToOAuthAuthentication = this.DefaultToOAuthAuthentication
                 DisablePublicNetworkAccess = this.DisablePublicNetworkAccess
                 DisableBlobPublicAccess = this.DisableBlobPublicAccess
                 DisableSharedKeyAccess = this.DisableSharedKeyAccess
-                DefaultToOAuthAuthentication = this.DefaultToOAuthAuthentication
+                DnsZoneType = this.DnsZoneType
+                EnableHierarchicalNamespace = this.EnableDataLake
+                ImmutableStorageWithVersioning = None // TODO: Implement
```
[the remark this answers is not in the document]
xperiandri: @ninjarobot I mean here that there is not such operation in the builder. That is what needs to be implemented

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -255,19 +278,25 @@ type StorageAccountBuilder() =
...
+        (state, access, names: string seq, ?immutabilityPolicies: BlobContainerImmutabilityPoliciesConfig)
+        =
+        {
+            state with
+                Containers =
+                    let containers =
+                        names
+                        |> List.ofSeq
+                        |> List.map (fun name ->
+                            ((StorageResourceName.Create name).OkValue, access, immutabilityPolicies))
```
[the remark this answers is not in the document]
xperiandri: I completely agree. I've missed that accidsentially

xperiandri:
> From an API design, what happens if you don't want to set any policies - do you need to provide a policy regardless (bear in mind computation expressions don't allow optional arguments AFAIK, and overloads must all have the same number of arguments)?

@isaacabraham, actually it works fine as all the tests pass. I verified that it works both ways and allow that optional parameter to be specified and omitted

isaacabraham:
> @isaacabraham, actually it works fine as all the tests pass. I verified that it works both ways and allow that optional parameter to be specified and omitted

That's good :-) Nonetheless, I think moving to a dedicated `container { }` CE at this stage makes more sense.

This PR still needs tests for the new code that's been created as well as the other points mentioned (including an issue / issues that reflect the actual changes that are being made here - it's not so much about upgrading the versions but about immutable blobs etc.).

Happy to talk through with you the API choices and anything else that's needed.

xperiandri:
I thought about that but considered immutable policy as not so big change to introduce container CE.
However further container customization support may need a CE

xperiandri:
I've implemented tests. Let me know if I miss something

xperiandri:
> Before merging, we need to modify the API - the use of a CE in this regard isn't going to work, from what I can see, from a usability perspective.

@isaacabraham my idea was to add now and preserve in the future ability to specify immutability policy
And add `blobContainer` CE and operations to `storageAccount` CE accepting `BlobContainerConfig` along with supporting additional container settings.
I have no intent to add them right now as it widens the scope of this PR.
Do you think `blobContainer` should be coupled with these changes?

isaacabraham:
I think so, yes. If you look elsewhere in the Farmer API, you won't really much in the way of CEs being supplied as secondary arguments to a custom keyword, and I really would like to retain the pattern that we have.

You can repurpose the CE you've already built to actually be the container one - just add in the extra name parameter. You can then add that as an overload for the top-level CE `add_private_container` member.

What do you think?

Thanks for adding the tests.

xperiandri:
@isaacabraham I've implemented `storageBlobContainer` CE as you suggested.
Any other remarks?

xperiandri:
I’ve done all the requested changes

src/Farmer/Arm/Storage.fs
```diff
@@ -182,41 +206,33 @@ type StorageAccount = {
...
-                        | None -> null
+                    allowBlobPublicAccess = this.DisableBlobPublicAccess.AsInvertedBoolean()
+                    allowSharedKeyAccess = this.DisableSharedKeyAccess.AsInvertedBoolean()
+                    defaultToOAuthAuthentication = this.DefaultToOAuthAuthentication.AsBoolean()
+                    dnsEndpointType = this.DnsZoneType |> Option.toObj
+                    encryption =
+                        this.RequireInfrastructureEncryption
+                        |> Option.map (fun _ -> {|
+                            requireInfrastructureEncryption = this.RequireInfrastructureEncryption
+                        |})
```
mattgallagher92, line 217: Using `this.RequireInfrastructureEncryption` inside the `map` seems weird! The behaviour is essentially this:

```fs
                        match this.RequireInfrastructureEncryption with
                        | Some true -> Some {| requireInfrastructureEncryption = Some true |}
                        | Some false -> Some {| requireInfrastructureEncryption = Some false |}
                        | None -> None
```

I might actually prefer that way of expressing it (I think it's a bit clearer), but presumably the `Some` is erased when we go to JSON, so we could simplify to this:

```suggestion
                        this.RequireInfrastructureEncryption
                        |> Option.map (fun b -> {| requireInfrastructureEncryption = b |})
```

Let me know if I'm wrong!

src/Farmer/Builders/Builders.Storage.fs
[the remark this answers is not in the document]
mattgallagher92: Let's use GitHub issues rather than TODO comments to track possible enhancements. Would you mind creating one @xperiandri?

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -795,5 +917,72 @@ type StorageQueueBuilder() =
...
+        (state: StorageBlobContainerConfig, policies: BlobContainerImmutabilityPoliciesConfig)
+        =
+        {
+            state with
+                ImmutabilityPolicies = policies.ToOption
+        }
+
+    /// Sets the name of the storage account.
+    [<CustomOperation "metadata">]
+    member _.Name(state: StorageBlobContainerConfig, metadata: (string * string) list) =
```
mattgallagher92, line 960:
```suggestion
    member _.Metadata(state: StorageBlobContainerConfig, metadata: (string * string) list) =
```

src/Farmer/Arm/Storage.fs
```diff
@@ -149,15 +175,13 @@ type StorageAccount = {
                     | Blobs _ -> "BlobStorage"
                     | Files _ -> "FileStorage"
                     | BlockBlobs _ -> "BlockBlobStorage"
+                extendedLocation = None // TODO:
+                identity = None // TODO: user assigned identity
```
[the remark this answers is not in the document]
mattgallagher92: I see from other discussions that perhaps it's for later enhancements. Let's use GitHub issues in that case.

mattgallagher92 (commented):
Thanks for your contribution @xperiandri and sorry for the delay - Isaac has been indisposed recently

I'm happy with the changes in principle, though have a few small queries in review comments.

src/Farmer/Arm/Storage.fs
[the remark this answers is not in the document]
xperiandri: If you prefer I can remove that however I would leave that

src/Farmer/Arm/Storage.fs
```diff
@@ -350,26 +366,92 @@ module BlobServices =
...
+                containers.Create(this.ResourceName, dependsOn = [ storageAccounts.resourceId this.StorageAccount ]) with
+                    properties = {|
+                        publicAccess = this.Accessibility.ArmValue
+                        metadata = this.Metadata |> Option.defaultValue Unchecked.defaultof<_>
+                    |}
+            |}
+
+type AllowProtectedAppendWrites =
+    | NoAppendAllowed
+    /// When enabled, new blocks can be written to both 'Append and Bock Blobs' while maintaining immutability protection and compliance. Only new blocks can be added and any existing blocks cannot be modified or deleted.
```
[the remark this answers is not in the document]
xperiandri: As far as I remember I copied that from official docs but we can modify this a bit as you suggested

src/Tests/Storage.fs
```diff
@@ -178,6 +143,299 @@ let tests =
...
+                                name "container1"
+                                metadata [ "environment", "dev"; "project", "farmer" ]
+                            }
+                            storageBlobContainer {
+                                name "container2"
+                                metadata [ "environment", "test"; "project", "barnyard" ]
+                            }
+                        ]
+
+                        add_blob_containers [
```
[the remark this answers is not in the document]
mattgallagher92: @xperiandri should the containers be used in assertions? If not, what's the point of adding them here?

mattgallagher92:
@xperiandri I've looked at the parts where you had TODOs in more depth based on `ImmutableStorageWithVersioning`. I see that there are two cases:

- The `immutableStorageWithVersioning` field in the ARM template has a representation in the `StorageAccount` type that implements `IArmResource`, which affects the generated JSON. However there is no corresponding configuration available in the builder, and no test that the generated JSON is correct.
 - The `extendedLocation` and `identity` fields will always be set to null in the generated JSON.

I have a few questions:

- Why did you add them if they're not used? Are the fields required in the ARM template JSON?
- If not, why not just leave them out of the Farmer codebase altogether?
- Why have you made `immutableStorageWithVersioning` configurable via the F# type but not done the same for `extendedLocation` and `identity`?

xperiandri:
> Why did you add them if they're not used? Are the fields required in the ARM template JSON?

Not required, just to mark they exist but not implemented

> If not, why not just leave them out of the Farmer codebase altogether?

I can if you prefer

> Why have you made `immutableStorageWithVersioning` configurable via the F# type but not done the same for `extendedLocation` and `identity`

I only implemented the piece I need and the scope defined by the PR name

mattgallagher92:
> > Why have you made immutableStorageWithVersioning configurable via the F# type but not done the same for extendedLocation and identity
>
> I only implemented the piece I need and the scope defined by the PR name

So you're using it directly on the `Farmer.Arm.Storage.StorageAccount` type rather than via the computation expression?

src/Farmer/Builders/Builders.Storage.fs
```diff
@@ -795,5 +917,72 @@ type StorageQueueBuilder() =
...
+        state with
+            AllowProtectedAppendWrites = Some value
+    }
+
+    /// Sets the ImmutabilityPeriodSinceCreation property.
+    [<CustomOperation "immutability_period_since_creation">]
+    member _.ImmutabilityPeriodSinceCreation(state: BlobContainerImmutabilityPoliciesConfig, value) = {
+        state with
+            ImmutabilityPeriodSinceCreation = Some value
+    }
```
xperiandri, line 982: > So you're using it directly on the Farmer.Arm.Storage.StorageAccount type rather than via the computation expression?

@mattgallagher92 no, here is the builder

xperiandri: I still have not got what your question is about.
Isaak asked to create a container CE.
I need immutability policies, so I also created CE for them.
So now we have blob container and blob container immutability policies CEs, and corresponding properties for them within the `StorageAccount` record.

Could you describe again what inconsistency you have found?

mattgallagher92: Sure 🙂 Thanks for your patience and sorry that this is dragging on so long. I don't have a lot of time available to review this (and unfortunately I'm going to be unavailable over the next few weeks).

When I search for references to the `ImmutableStorageWithVersioning` property on `Farmer.Arm.Storage.StorageAccount`, I see that it's used in 4 places. 3 of those places set it to `None`, and the remaining place uses it via `Option.map`, which will therefore also always return `None`. As far as I can see, there doesn't seem to be any means of setting it to anything other than `None` via a builder, so it seems to be that the `Option.map` code is not being tested and is perhaps even redundant.

Given it doesn't seem to be possible to set it via builder, I wondered whether you were setting the field directly on the record somehow, or whether this is something that you don't need.

mattgallagher92: @xperiandri I see a commit labelled `Implemented storage account level immutability policy setting and builder`, but I don't actually see the implementation in the diff?

PR #1152, opened 2024-11-05
Use Common Protocol if multiple Sources in a NSG Rule have a common protocol

### LINKED ISSUE #1151: Security Rules defaults to "Any" protocol when multiple sources are provided.
Problem:
The securityRule builder will default to "Any" when more than one sources are added. For example, the below code will generate a security rule with Any protocol. Since the the protocol being used across rules is same, we should use the common protocol instead of Any since that is most likely the functionality that the user intended. 

```
{
    name "same-protocols"
    description "Security Rule with multiple sources of same protocol"
    services [ "http", 80 ]
    add_source_network NetworkSecurity.TCP "10.28.0.0/24"
    add_source_network NetworkSecurity.TCP "10.30.0.0/24"
    add_destination_network "10.28.0.0/24"
} 
```

tanmay-j:
This PR closes #1151

* Changed the Security Rule Config to use the common protocol across sources and default to Any (*) if there are multiple protocols are used.

* The documentation does not mention how the protocol is calculated while building security rules.
* The change was tested via unit tests.


isaacabraham:
Thank you very much for this PR! I think if you can update the docs to make this explicit (as it wasn't before - which caused unintended behaviour) and update the release notes if not done, that would be great.

The only concern / question I have is that this will (unless I'm mistaken) generate different ARM templates for existing users with no code changes. That's something we try our best to avoid - what's the worst case scenario here?

ninjarobot (approved):
Looks great, thanks for the complete bug fix including tests that show the issue and repair.

tanmay-j:
Hey, yes this would lead to a change in the protocol property in the security rule but only for users using multiple sources with the same protocol within the same rule. I reckon that would be a very small minority of users but I don't think we can fix the issue without changing this behavior.

Issue #731, opened 2021-08-11
Conditional values in Farmer builders, or composing builders?


Thorium:
I didn't find documentation about this:
What is the best practice to work with conditional values in Farmer custom builders?

For example, the VM IP ARM (PublicIpId) is an Option type:

```fsharp
 let deployment = arm {
        location Location.UKSouth
        // more resources here
        add_resource myVm
        output "vmIP" myVm.PublicIpId.Value.ArmExpression
    }
```
The code fails if PublicIpId is None.
When the deployments and resources are compex, I wouldn't want to create multiple copies of the same builders.

I'd like to do something like:
```fsharp
    let deployment = arm {
        location Location.UKSouth
        // more resources here
        add_resource myVm
        match myVm.PublicIpId with
        | Some ip -> output "vmIP" ip.ArmExpression
        | None -> ()
    }
// or
    let deployment = arm {
        location Location.UKSouth
        // more resources here
        add_resource myVm
        if myVm.PublicIpId.IsSome then
            output "vmIP" myVm.PublicIpId.Value.ArmExpression
    }
```

> This control construct may only be used if the computation expression builder defines a 'For' method.

Or could I use composition and still take the advantage of the builder?
Currently I've only find a solution to jump out of the builder and start to build manually the state, losing the convenience of builder with proper type-conversions in overloaded methods:
```fsharp
    let composition =
        match myVm.PublicIpId with
        | None -> deployment
        | Some ip -> { deployment with Outputs = deployment.Outputs.Add("vmIP", ip.ArmExpression.Eval()) }

```

Thorium:
```fsharp
    let deployment = arm {
        location Location.UKSouth
        add_resource myVm
        output "vmIP" (myVm.PublicIpId |> Option.map(fun ip -> ip.ArmExpression))
    }
```

Now, only thing more to add, what if there are a list of VMs...

davidglassborow:
I've got exactly the same issue, trying to work out how to configure different instances (staging/acc/prod) from the same set of builders

dburriss:
@isaacabraham I have run into this a few times and the workarounds are decidedly inelegant.

I attached a PR to this with a proposal.

I initially checked to see if I could get `if` working in the webapp CE but it has been a few years but I am not even sure it is possible with a non-sequence data structure? The benefit of this would be that it could be added to each CE and unblock this issue without requiring a change to each "flag" operator.

[...]

PR #1185, opened 2025-02-13
Example proposal for Issue #731

dburriss:
This is linked to the issue described in https://github.com/CompositionalIT/farmer/issues/731 and is just for demo purposes.

I have encountered this issue a few times with Farmer and do not know of an elegant solution.
Something like this can get you out in a pinch but it requires knowing what the builder is setting, which leaks the details. It's also just not very elegant.

```fsharp
let mutable myWebApp = webApp {
            name "myWebApp"
            service_plan_name "myServicePlan"
            setting "myKey" "aValue"
            sku WebApp.Sku.B1
            always_on
            app_insights_off
            worker_size Medium
            number_of_workers 3
            run_from_package
            system_identity
        }
        let disableAlwaysOn = false
        if disableAlwaysOn then 
            myWebApp <- { myWebApp with WebAppConfig.CommonWebConfig.AlwaysOn = true }
```
An elegant solution would be if we could implement `For` and `Zero` in a way that allowed for `if` in the CE but playing around with it for a few minutes did not work as expected. I have not worked with CEs for a while but not sure if this is even possible for a non-collection based CE.

A simple solution (albeit quite some work) is to take in a value but add an optional parameter.

```fsharp
let expected = false
let template = webApp {
  name "web"
  always_on expected
}
```

This matches what I usually do in builders in C# where boolean switching takes in a boolean with a default value.

### DIFF
--- a/src/Farmer/Builders/Builders.WebApp.fs
+++ b/src/Farmer/Builders/Builders.WebApp.fs
@@ -947,24 +947,6 @@ type WebAppBuilder() =
         VirtualApplications = Map []
         FunctionAppScaleLimit = None
     }
-    member __.Combine (currentValueFromYield: WebAppConfig, accumulatorFromDelay: WebAppConfig) =
-        System.Console.WriteLine($"currentValueFromYield: {currentValueFromYield.CommonWebConfig.AlwaysOn}")
-        System.Console.WriteLine($"accumulatorFromDelay: {accumulatorFromDelay.CommonWebConfig.AlwaysOn}")
-        { currentValueFromYield with
-           WebAppConfig.CommonWebConfig.AlwaysOn = false//accumulatorFromDelay.CommonWebConfig.AlwaysOn
-        }
-
-    member this.For(state: WebAppConfig , f: unit -> WebAppConfig) =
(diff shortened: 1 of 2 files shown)


mattgallagher92:
Thanks @dburriss! I like this approach 🙂 Particularly because it's a very simple change that is backwards compatible.

It will only work for simple on/off cases and not those where we need to pass in more complex data, but is definitely a good idea for them.

As you say, it's potentially quite a lot of work, but would be fairly easy to apply in increments - there's no need to change every Farmer CE in one PR. If you'd like to raise a PR(s) for the cases that you need (or more!), I'm sure that we'll be happy to merge them 🙂

dburriss:
@mattgallagher92 Ok nice. As I mentioned, I have a usage where I will likely need this flexibility on a couple of resources, so I will make the changes on the CEs for all boolean switches as needed (I will do all on a CE, not just the ones I need ;) )
This will mostly be compute resources e.g. WebApp, Containers, maybe VMs. It has been 4 years since I contributed to Farmer. The codebase is familiar but there are A LOT more resources and the files are A LOT bigger :)

PR #1202, opened 2025-05-22
Namespace for sbus scale rules

### LINKED ISSUE #1188: Add Dapr service bus scale rule custom operation
`containerApp` builders already have support for `add_custom_scale_rule`, which I've used a few times to add service bus scale rules. I think that perhaps a specific `add_service_bus_scale_rule` operation could be useful too.

I hope to be able to contribute this in the next few weeks, just opening this issue in case anyone has comments before I raise a PR.

Larocceau:
This PR closes #1188 

* Adds the field namespace to the ServiceBusScaleRule, which is used to associate the scale rule to the right sbus

* Instead of testing on a existing deployment, I compared a ARM template of a successful implementation of this feature using a custom scale rule, with one using this feature. There was no significant difference in the two templates.

```fsharp
containerApp {
      ....
      add_servicebus_scale_rule "sb-keda-scale" {
          QueueName = "wishrequests"
          MessageCount = 5
          SecretRef = "servicebusconnectionkey"
          Namespace = "servicebus"
      }
}
```


src/Tests/ContainerApps.fs
mattgallagher92, line 1: Should one of the tests check that the value is set correctly?

RELEASE_NOTES.md
mattgallagher92, line 1: Do the docs need updating too?

mattgallagher92 (commented):
Is this field optional? If so, this change will break any existing clients, so we should probably make the new field optional too. If it's mandatory then there can't be any existing clients so we don't have to worry about that, in which case this change is fine (but I've added some comments about some things to improve).

Larocceau:
Hi @ninjarobot! Sorry for the delay; things been a bit messy for us, and I've left CIT (as has most other dev staff). I did not get around this, but would be happy to wrap this up later.

PR #1208, opened 2025-08-01
Support Ability to Enable Azure Monitor on AKS and Populate Prometheus Rules

### LINKED ISSUE #1209: Support Azure Monitor Data Collection Rules and Promethus Rules Groups for AKS

michelleli01:
This PR closes #1209 

* Adds data collection endpoints and data collection rules 
* Adds prometheus rule group and prometheus rules 
* Adds flag to enable azure monitor in aks builder

```fsharp
```


src/Farmer/Arm/AlertsManagement.fs
```diff
@@ -0,0 +1,128 @@
...
+        Severity = None
+        Actions = None
+        ResolveConfiguration = None
+        For = None
+    }
+
+    static member ToArmJson(rule: PrometheusRule) = {|
+        actions =
+            rule.Actions
+            |> Option.map (fun actions -> actions |> List.map (fun action -> action.ToArmJson))
```
ninjarobot, line 61: This can be simplified to just the following:
```fsharp
|> Option.map (List.map _.ToArmJson)
```
There are lots of lambdas that could be removed with this, but I mainly would recommend this for the ones that can remove significant noise like this.

src/Farmer/Arm/AlertsManagement.fs
```diff
@@ -0,0 +1,128 @@
...
+            |> Option.defaultValue Unchecked.defaultof<_>
+        record = rule.Record |> Option.defaultValue Unchecked.defaultof<_>
+        expression = rule.Expression
+        labels =
+            rule.Labels
+            |> Option.map (Map.toList >> dict)
+            |> Option.defaultValue Unchecked.defaultof<_>
+        alert = rule.Alert |> Option.defaultValue Unchecked.defaultof<_>
+        enabled = rule.Enabled |> Option.map (fun e -> e.AsBoolean)
```
ninjarobot, line 70: You can remove the lambda here, like this:
```fsharp
enabled = rule.Enabled |> Option.map _.AsBoolean
```

src/Farmer/Arm/AlertsManagement.fs
```diff
@@ -0,0 +1,128 @@
...
+            |> Option.map (fun severity ->
+                match severity with
+                | AlertSeverity.Critical -> 0
+                | AlertSeverity.Error -> 1
+                | AlertSeverity.Warning -> 2
+                | AlertSeverity.Informational -> 3
+                | AlertSeverity.Verbose -> 4)
+        resolveConfiguration =
+            rule.ResolveConfiguration
+            |> Option.map (fun config -> config.ToArmJson)
```
ninjarobot, line 82: Also can simplify this to
```fsharp
|> Option.map _.ToArmJson
```

src/Farmer/Arm/AlertsManagement.fs
```diff
@@ -0,0 +1,128 @@
...
+                            |> Option.map (fun name -> name.Value)
+                            |> Option.defaultValue Unchecked.defaultof<_>
+                        description = this.Description
+                        enabled = this.Enabled |> Option.map (fun e -> e.AsBoolean)
+                        interval =
+                            this.Interval
+                            |> Option.map (fun interval ->
+                                match interval with
+                                | IsoDateTime x -> x)
+                        scopes = scopes |> Set.map (fun scope -> scope.Eval())
```
ninjarobot, line 125: This can simplify with
```fsharp
Set.map _.Eval()
```

src/Farmer/Arm/AlertsManagement.fs
```diff
@@ -0,0 +1,128 @@
...
+                            this.ClusterName
+                            |> Option.map (fun name -> name.Value)
+                            |> Option.defaultValue Unchecked.defaultof<_>
+                        description = this.Description
+                        enabled = this.Enabled |> Option.map (fun e -> e.AsBoolean)
+                        interval =
+                            this.Interval
+                            |> Option.map (fun interval ->
+                                match interval with
+                                | IsoDateTime x -> x)
```
ninjarobot, line 124: This will be a little clearer as
```fsharp
|> Option.map (function | IsoDateTime x -> x)
```

src/Farmer/Arm/AlertsManagement.fs
```diff
@@ -0,0 +1,128 @@
...
+                        scopes = scopes |> Set.map (fun scope -> scope.Eval())
+                        rules = this.Rules |> List.map (fun r -> r |> PrometheusRule.ToArmJson)
```
ninjarobot, line 126: This simplifies to
```fsharp
rules = this.Rules |> List.map PrometheusRule.ToArmJson
```

src/Farmer/Builders/Builders.AlertsManagement.fs
```diff
@@ -0,0 +1,199 @@
...
+module Farmer.Builders.AlertsManagement
+
+open Farmer
+open Farmer.Arm
+
+type PrometheusRuleConfig = {
+    Record: string option
+    Expression: string
+    Labels: Map<string, string> option
+    Enabled: FeatureFlag option
```
ninjarobot, line 11: It looks like these are going to default to disabled. Is that the intention or should this not be an option and it default to enabled so people don't have to remember to explicitly enable each rule?

src/Farmer/Builders/Builders.AlertsManagement.fs
```diff
@@ -0,0 +1,199 @@
...
+
+
+let prometheusRule = PrometheusRuleBuilder()
+
+type PrometheusRuleGroupConfig = {
+    Name: ResourceName
+    Description: string option
+    ClusterName: ResourceName option
+    Tags: Map<string, string>
+    Enabled: FeatureFlag option
```
ninjarobot, line 89: Same here, doesn't this default to deploying a disabled rule group if they didn't specifically enable it?

ninjarobot (commented):
Thanks for the extensive PR to add all this Azure Monitor support for AKS. I have a few suggestions to use property shorthand and reduce the noise from some of the lambdas. Please give those some consideration where they improve code clarity. 

One functionality adjustment I wanted to suggest is to make sure the rules and rule groups are enabled by default. It _looks_ like they would be disabled unless someone explicitly enables them which is probably not the intention.

src/Farmer/Arm/AlertsManagement.fs
michelleli01: Addressed all of these in latest commit to use lambdas!

src/Farmer/Builders/Builders.AlertsManagement.fs
michelleli01: Changed to explicitly set these to default in latest commit

ninjarobot (approved):
Looks great, thanks for simplifying some of the complex lambdas and for the explanation on the rules and collections being enabled by default.

PR #1215, opened 2025-09-29
Basic WebApp supports VNET integration

Ryan-Palmer:
Addresses #1216


ninjarobot (approved):
@Ryan-Palmer thanks for addressing this and providing a quick fix.

PR #1225, opened 2025-10-28
Add TLS1.3, remove unsecure TLS1.0 and TLS1.1 support.

Thorium:
A little maintenance PR...


Thorium:
Yes I noticed, but I did that: 
image

What version it should use?

PR #1232, opened 2025-11-04
Added some production-default recommendations

Thorium:
I did ask from Claude Sonnet what is in FsCDK (AWS) that Farmer (Azure) is missing, and it came up with this PR:

- Farmer lacks best-practice safe settings for production ready environments.
- So this build adds those

Farmer already has some opinionated approaches, but it doesn't really go into "production-environment-focus" (cost-efficiency, logging, hardened security, and so on).

Be suspicious about agents... Meanwhile this PR can be ready for merge, I'm not actually expecting that.
Rather the question is, would we want to take Farmer into this direction or no?

### DIFF
--- a/docs/content/api-overview/resources/app-insights.md
+++ b/docs/content/api-overview/resources/app-insights.md
@@ -19,6 +19,8 @@ The App Insights builder is used to create Application Insights accounts. Use th
 | name | Sets the name of the App Insights instance. |
 | disable_ip_masking | Disable IP masking. |
 | sampling_percentage | Define sampling percentage (0-100) |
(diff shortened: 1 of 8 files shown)


ninjarobot (approved):
These changes look good from the standpoint of functionality and compatibility. All the documentation is definitely appreciated as well.

Thorium:
This was rather question (with code attached) of do we want to take this direction, rather than a PR intended to be merged. We can close the PR but what you think about the idea?

Thorium:
Generally working with all agents, you add AGENTS.md as general instructions and patterns how they should do things. You can ask e.g. Claude to generate you a template and then modify it by your needs.

